PR #4, opened 2018-08-29
Story/vspc 2

Neehaarika:
User should be able to define a label when creating a new space link


jdamerow:
I'm confused, parts of this look like the code from VSPC-1. Only changes for VSPC-2 should be submitted here. Also, the label doesn't seem to be passed to the controller and stored in the database?

PR #5, opened 2018-09-05
Story/vspc 6

Neehaarika:
When adding a new space link, the rotation the user enters should be reflected in the icon


vspace/src/main/webapp/WEB-INF/views/staff/spaces/space.jsp
```diff
@@ -56,6 +58,8 @@ $( document ).ready(function() {
 	$("#cancelSpaceLinkBtn").click(function() {
 		storeX = null;
 		storeY = null;
+		$("#arrow").remove();
+		$("#arrow").removeAttr("id");
```
jdamerow: if #arrow is already removed, why do you remove the attribute as well?

Neehaarika: Will modify that.

PR #16, opened 2018-09-19
Story/vspc 2

Neehaarika:
Used SpaceLink, instead of Display.


jdamerow:
There needs to be a new line after the label input form. Right it looks like this (type is now broken up into two lines):
image

Also, the label should be shown on the image. For an alert, inside the alert box. For an arrow, at the end of the error (not the pointy side, but the other one).

PR #23, opened 2018-10-12
Story/vspc 15

Neehaarika:
The domain classes for modules need to be extended.

### DIFF
--- a/vspace/src/main/java/edu/asu/diging/vspace/core/model/IBranchingPoint.java
+++ b/vspace/src/main/java/edu/asu/diging/vspace/core/model/IBranchingPoint.java
@@ -1,0 +1,12 @@
+package edu.asu.diging.vspace.core.model;
+
+import java.util.List;
+
+import edu.asu.diging.vspace.core.model.impl.Choice;
+
+public interface IBranchingPoint {
+
+	List<Choice> getChoices();
+	
+	void setChoices(List<Choice> choices);
+}
--- a/vspace/src/main/java/edu/asu/diging/vspace/core/model/IChoice.java
+++ b/vspace/src/main/java/edu/asu/diging/vspace/core/model/IChoice.java
@@ -1,0 +1,8 @@
+package edu.asu.diging.vspace.core.model;
+
+public interface IChoice extends IVSpaceElement{
+	
+	ISequence getLink();
+	
+	void setLink(ISequence link);
+}
--- a/vspace/src/main/java/edu/asu/diging/vspace/core/model/ISlide.java
+++ b/vspace/src/main/java/edu/asu/diging/vspace/core/model/ISlide.java
@@ -1,0 +1,7 @@
+package edu.asu.diging.vspace.core.model;
+
+
+public interface ISlide {
+	
+	
+}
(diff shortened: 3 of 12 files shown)


vspace/src/main/java/edu/asu/diging/vspace/core/model/IBranchingPoint.java
```diff
@@ -0,0 +1,12 @@
...
+
+import java.util.List;
+
+import edu.asu.diging.vspace.core.model.impl.Choice;
+
+public interface IBranchingPoint {
+
+	List<Choice> getChoices();
+	
+	void setChoices(List<Choice> choices);
```
jdamerow: those methods need to return/work with the interfaces not the concrete implementations

vspace/src/main/java/edu/asu/diging/vspace/core/model/IChoice.java
```diff
@@ -0,0 +1,8 @@
+package edu.asu.diging.vspace.core.model;
+
+public interface IChoice extends IVSpaceElement{
+	
+	ISequence getLink();
```
jdamerow: Should be called 'getSequence' not getLink since it's returning a sequence

vspace/src/main/java/edu/asu/diging/vspace/core/model/ISlide.java
```diff
@@ -0,0 +1,7 @@
+package edu.asu.diging.vspace.core.model;
+
+
+public interface ISlide {
+	
+	
+}
```
jdamerow: doesn't seem right that this is empty

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/BranchingPoint.java
```diff
@@ -0,0 +1,37 @@
...
+	@Id
+	@GeneratedValue(generator = "id-generator")
+	@GenericGenerator(name = "id-generator", 	
+					parameters = @Parameter(name = "prefix", value = "BPNT"), 
+					strategy = "edu.asu.diging.vspace.core.data.IdGenerator"
+			)
+	private String id;
+	
+	@OneToMany(mappedBy = "branchingpoint", targetEntity=Choice.class)
+	private List<Choice> choices;
```
jdamerow: needs to use interface

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/Choice.java
```diff
@@ -0,0 +1,49 @@
...
+import org.hibernate.annotations.Parameter;
+
+import edu.asu.diging.vspace.core.model.IChoice;
+import edu.asu.diging.vspace.core.model.ISequence;
+
+public class Choice extends VSpaceElement implements IChoice{
+	
+	@Id
+	@GeneratedValue(generator = "id-generator")
+	@GenericGenerator(name = "id-generator",
```
jdamerow: your id-generators all have the same name, they need to be named differently

vspace/src/main/java/edu/asu/diging/vspace/core/model/ISlide.java
Neehaarika: all the properties of Slide - title, description, and image have been declared in other interfaces. Do you want me to remove this interface in that case?

jdamerow: your implementaiton of ISlide has getters and setters for an image though, doesn't it?

jdamerow: also the interface doesn't extends IVSpaceElement

PR #26, opened 2018-10-23
Story/vspc 14

Neehaarika:
Addressed comments and wrote test cases for getModule() method. Please review.

### DIFF
--- a/vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ModuleManager.java
+++ b/vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ModuleManager.java
@@ -30,9 +30,9 @@ public class ModuleManager implements IModuleManager {
 
     @Override
     public IModule getModule(String id) {
-        boolean isExists = moduleRepo.existsById(id);
-        if (isExists) {
-            return moduleRepo.findById(id).get();
+        Optional<Module> module = moduleRepo.findById(id);
+        if (module.isPresent()) {
+            return module.get();
         }
         return null;
     }
--- a/vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ModuleManager.java
+++ b/vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ModuleManager.java
@@ -1,10 +1,7 @@
 package edu.asu.diging.vspace.core.services.impl;
 
-import java.util.NoSuchElementException;
 import java.util.Optional;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
@@ -19,7 +16,7 @@ public class ModuleManager implements IModuleManager {
     @Autowired
     private ModuleRepository moduleRepo;
     
-    private final Logger logger = LoggerFactory.getLogger(getClass());
+    //private final Logger logger = LoggerFactory.getLogger(getClass());
 
     /*
      * (non-Javadoc)
@@ -37,12 +34,8 @@ public class ModuleManager implements IModuleManager {
     public IModule getModule(String id) {
         Optional<Module> module = moduleRepo.findById(id);
         if (module.isPresent()) {
-            try {
-                return module.get();
-            } catch(NoSuchElementException e) {
-                logger.error("No Such Element Exception", e);
-            }            
+            return module.get();           
         }
         return null;
-    }
+    }       
 }
--- a/vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ModuleManager.java
+++ b/vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ModuleManager.java
@@ -13,21 +13,25 @@ import edu.asu.diging.vspace.core.services.IModuleManager;
 @Service
 public class ModuleManager implements IModuleManager {
 
-	@Autowired
-	private ModuleRepository moduleRepo;
-	
-	/* (non-Javadoc)
-	 * @see edu.asu.diging.vspace.core.services.impl.IModuleManager#storeModule(edu.asu.diging.vspace.core.model.IModule, java.lang.String)
-	 */
-	@Override
-	public IModule storeModule(IModule module) {
-		return moduleRepo.save((Module)module);
-	}
-	
-	@Override
+    @Autowired
+    private ModuleRepository moduleRepo;
+
+    /*
+     * (non-Javadoc)
+     * 
+     * @see
+     * edu.asu.diging.vspace.core.services.impl.IModuleManager#storeModule(edu.asu.
+     * diging.vspace.core.model.IModule, java.lang.String)
+     */
+    @Override
+    public IModule storeModule(IModule module) {
+        return moduleRepo.save((Module) module);
+    }
+
+    @Override
     public IModule getModule(String id) {
         Optional<Module> module = moduleRepo.findById(id);
-        if (module != null && module.isPresent()) {
+        if (module.isPresent()) {
             return module.get();
         }
         return null;
--- a/vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ModuleManager.java
+++ b/vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ModuleManager.java
@@ -1,5 +1,7 @@
 package edu.asu.diging.vspace.core.services.impl;
 
+import java.util.Optional;
+
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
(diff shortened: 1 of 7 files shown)


vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ModuleManagerTest.java
```diff
@@ -0,0 +1,41 @@
...
+        MockitoAnnotations.initMocks(this);
+    }
+
+    @Test
+    public void test_getModule_success() {
+
+        Module newModule = new Module();
+        Optional<Module> findModule = Optional.of(newModule);
+        managerToTest.storeModule(newModule);
+        Mockito.when(moduleRepo.findById(newModule.getId())).thenReturn(findModule);
```
jdamerow: getId returns null here, it needs to be set first

jdamerow: also, let's first set up the test then test, meaning first create all the objects you need and tell mockito what to do, then call the method to test, then assert the results. right now you're mixing the order.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ModuleManagerTest.java
```diff
@@ -0,0 +1,41 @@
...
+        IModule moduleTest = managerToTest.getModule(findModule.get().getId());
+        Assert.assertEquals(moduleTest, newModule);
```
jdamerow: actually that's not really what we require, the objects don't have to be the same, but the properties of them have to be. let's test for the id and title.

Neehaarika:
Refactored the test cases for the getModule(String id) based on id as per the method implementation.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ModuleManager.java
```diff
@@ -11,14 +13,27 @@
...
+    @Override
+    public IModule storeModule(IModule module) {
+        return moduleRepo.save((Module) module);
+    }
+
+    @Override
+    public IModule getModule(String id) {
+        boolean isExists = moduleRepo.existsById(id);
+        if (isExists) {
+            return moduleRepo.findById(id).get();
```
jdamerow: nope, just do a 'isPresent' on the optional before calling get.

Neehaarika: moduleRepo.existById(id) is internally performing a isPresent() check in its method implementation. Also since I faced issue to test that condition in test cases, I changed it to existsById which can be mocked as it is a repo method.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ModuleManager.java
```diff
@@ -11,14 +16,33 @@
...
+
+    @Override
+    public IModule getModule(String id) {
+        Optional<Module> module = moduleRepo.findById(id);
+        if (module.isPresent()) {
+            try {
+                return module.get();
+            } catch(NoSuchElementException e) {
+                logger.error("No Such Element Exception", e);
+            }
```
jdamerow: you already checked that there is an object present, so no need to catch the exception

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ModuleManagerTest.java
```diff
@@ -0,0 +1,54 @@
...
+    public void init() {
+        MockitoAnnotations.initMocks(this);
+    }
+
+    @Test
+    public void test_getModule_idExists() {
+
+        Module newModule = new Module();
+        Optional<Module> mockModule = Optional.of(newModule);
+        managerToTest.storeModule(newModule);
```
jdamerow: why do you call 'storeModule' here?

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ModuleManagerTest.java
```diff
@@ -0,0 +1,54 @@
...
+    @Test
+    public void test_getModule_idNotExists() {
+
+        Module newModule = new Module();
+        Optional<Module> mockModule = Optional.of(newModule);
+        Mockito.when(mockModuleRepo.findById("mockString")).thenReturn(mockModule);
+
+        IModule iModuleActual = managerToTest.getModule("mockString");
+        Assert.assertEquals(mockModule.get().getId(), iModuleActual.getId());
+        Assert.assertEquals(mockModule.get().getName(), iModuleActual.getName());
```
jdamerow: you're just testing that the id exists again. but this test is called 'idNotExists'?

Neehaarika:
Please review.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ModuleManager.java
```diff
@@ -11,14 +13,29 @@
...
-	 * @see edu.asu.diging.vspace.core.services.impl.IModuleManager#storeModule(edu.asu.diging.vspace.core.model.IModule, java.lang.String)
-	 */
-	@Override
-	public IModule storeModule(IModule module) {
-		return moduleRepo.save((Module)module);
-	}
+    @Autowired
+    private ModuleRepository moduleRepo;
+    
+    //private final Logger logger = LoggerFactory.getLogger(getClass());
```
jdamerow: remove

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ModuleManagerTest.java
```diff
@@ -0,0 +1,47 @@
...
+
+        IModule iModuleActual = managerToTest.getModule(newModule.getId());
+        Assert.assertEquals(mockModule.get().getId(), iModuleActual.getId());
+        Assert.assertEquals(mockModule.get().getName(), iModuleActual.getName());
+    }
+    
+    @Test(expected = NoSuchElementException.class)
+    public void test_getModule_idNotExists() throws Exception {
+        Mockito.when(mockModuleRepo.findById(Mockito.anyString())).thenReturn(Optional.empty());
+        Mockito.when(mockModuleRepo.findById(Mockito.anyString()).get()).thenThrow(new NoSuchElementException());;
```
jdamerow: when is that thrown? the repository doesn't throw that I believe.

jdamerow: and your manager should just return null if there is no such module (which it does).

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ModuleManagerTest.java
```diff
@@ -0,0 +1,46 @@
...
+        IModule iModuleActual = managerToTest.getModule(newModule.getId());
+        Assert.assertEquals(mockModule.get().getId(), iModuleActual.getId());
+        Assert.assertEquals(mockModule.get().getName(), iModuleActual.getName());
+    }
+
+    @Test(expected = NullPointerException.class)
+    public void test_getModule_idNotExists() throws Exception {
+        Assert.assertEquals(mockModuleRepo.findById(Mockito.anyString()).get(),
+                managerToTest.getModule(Mockito.anyString()).getId());
+    }
```
jdamerow: this doesn't make sense. You say that a null pointer exception is expected, but the getModule method does not throw a null pointer. You need to test that the method returns null if the id doesn't exist.

Neehaarika:
Please Review.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ModuleManagerTest.java
```diff
@@ -0,0 +1,46 @@
...
+    @Before
+    public void init() {
+        MockitoAnnotations.initMocks(this);
+    }
+
+    @Test
+    public void test_getModule_idExists() {
+        Module newModule = new Module();
+        Optional<Module> mockModule = Optional.of(newModule);
+        Mockito.when(mockModuleRepo.findById(newModule.getId())).thenReturn(mockModule);
```
jdamerow: newModule.getId() would return null, wouldn't it? findById(null) doesn't make sense, it's not a valid call.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ModuleManagerTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+        IModule iModuleActual = managerToTest.getModule(newModule.getId());
+        Assert.assertEquals(mockModule.get().getId(), iModuleActual.getId());
+        Assert.assertEquals(mockModule.get().getName(), iModuleActual.getName());
+    }
+
+    @Test
+    public void test_getModule_idNotExists() throws Exception {
+        Mockito.when(mockModuleRepo.findById(Mockito.anyString())).thenReturn(Optional.empty());
+        Assert.assertEquals(null, managerToTest.getModule(Mockito.anyString()));
```
jdamerow: I believe there there is an assertNull method you can use instead

jdamerow:
test this please

PR #28, opened 2018-11-09
Story/vspc 17

Neehaarika:
Please review.


vspace/src/main/java/edu/asu/diging/vspace/core/model/display/impl/ExternalLinkDisplay.java
```diff
@@ -0,0 +1,118 @@
...
+
+import org.hibernate.annotations.GenericGenerator;
+import org.hibernate.annotations.Parameter;
+
+import edu.asu.diging.vspace.core.model.IExternalLink;
+import edu.asu.diging.vspace.core.model.display.IExternalLinkDisplay;
+import edu.asu.diging.vspace.core.model.impl.ExternalLink;
+import edu.asu.diging.vspace.core.model.impl.VSpaceElement;
+
+public class ExternalLinkDisplay extends VSpaceElement implements IExternalLinkDisplay {
```
jdamerow: Let's create a display superclass for all link displays, since they will all have positional data, type data, etc.

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/ExternalLink.java
```diff
@@ -0,0 +1,81 @@
...
+      strategy = "edu.asu.diging.vspace.core.data.IdGenerator")
+	private String id;
+	
+	@ManyToOne(targetEntity=Space.class)
+    @JoinColumn(name="space_id", nullable=false)
+	private ISpace space;
+	
+	@ManyToOne(targetEntity=ExternalLink.class)
+    @JoinColumn(name="external_link", nullable=false)
+	private String externalLink;
```
jdamerow: why is that a manytoone? It's just a string.

jdamerow: also indentation is messed up

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceManager.java
```diff
@@ -145,6 +170,23 @@ public ISpaceLinkDisplay createSpaceLink(String title, ISpace source, float posi
         return display;
     }
 
+    @Override
+    public IExternalLinkDisplay createExternalLink(String title, ISpace space, float positionX, float positionY,
+            String url) throws SpaceDoesNotExistException {
```
jdamerow: this exception is never thrown. also, looking at my code above, there would be a different exception thrown if the space does not exist. this also needs to be fixed.

Neehaarika: Ok. Here we don't have to check for target space so I can remove that exception. Should we check if the URL(externallink) is a valid one or not and throw an Exception?

vspace/src/main/java/edu/asu/diging/vspace/core/model/display/impl/Display.java
```diff
@@ -0,0 +1,42 @@
+package edu.asu.diging.vspace.core.model.display.impl;
+
+import edu.asu.diging.vspace.core.model.display.IDisplay;
+import edu.asu.diging.vspace.core.model.impl.VSpaceElement;
+
+public abstract class Display extends VSpaceElement implements IDisplay {
```
jdamerow: Let's call that LinkDisplay since it's meant for links.

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/ExternalLink.java
```diff
@@ -0,0 +1,81 @@
...
+
+import edu.asu.diging.vspace.core.model.IExternalLink;
+import edu.asu.diging.vspace.core.model.ISpace;
+
+@Entity
+public class ExternalLink extends VSpaceElement implements IExternalLink {
+	
+	@Id 
+	@GeneratedValue(generator = "id-generator")
+    @GenericGenerator(name = "extlink-id-generator",
```
jdamerow: indentation

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/ExternalLink.java
```diff
@@ -0,0 +1,81 @@
...
+import org.hibernate.annotations.Parameter;
+
+import edu.asu.diging.vspace.core.model.IExternalLink;
+import edu.asu.diging.vspace.core.model.ISpace;
+
+@Entity
+public class ExternalLink extends VSpaceElement implements IExternalLink {
+	
+	@Id 
+	@GeneratedValue(generator = "id-generator")
```
jdamerow: this should be the same as the generic generator you specify below

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceManager.java
```diff
@@ -123,14 +143,25 @@ public ISpace getFullyLoadedSpace(String id) {
...
         // we need this to fully load the space
+        ISpace target;
         source = spaceRepo.findById(source.getId()).get();
-        ISpace target = spaceRepo.findById(linkedSpaceId).get();
-        if (target == null) {
+        boolean isExists = spaceRepo.existsById(linkedSpaceId);
+        if(!isExists) {
             throw new SpaceDoesNotExistException();
+        } else {
+        target = spaceRepo.findById(linkedSpaceId).get();
```
jdamerow: let's get an optional from 'findById' and call isPresent on it. The way you do it, this could still end up in an exception in case someone delete the space between the two calls you make to the repository (existsById and findById). There is also no reason for two DB calls.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceManager.java
```diff
@@ -145,6 +176,23 @@ public ISpaceLinkDisplay createSpaceLink(String title, ISpace source, float posi
         return display;
     }
 
+    @Override
+    public IExternalLinkDisplay createExternalLink(String title, ISpace space, float positionX, float positionY,
+            String url) {
+        // we need this to fully load the space
+        space = spaceRepo.findById(space.getId()).get();
```
jdamerow: what if the space doesn't exist?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceManager.java
jdamerow: line 156 also needs to check if the space actually exist before getting it

Neehaarika: Maybe I can just use the getSpace(String id) method which does and checks all the conditions?

Neehaarika: We are doing a null check for space in AddSpaceLinkController and fetching the space before calling spaceManager. So isnt it assured that the space exists?

jdamerow: sure that works.

PR #46, opened 2019-04-05
[VSPC-34] made changes to indentation as well as moved all db logic from controller to service

ejmath12:


vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
```diff
@@ -62,4 +78,23 @@ public ImageData getImageDimensions(IVSImage image, int width, int height) {
...
+    
+    /**
+     * Method to return the required attributes of a requested page  
+     * 
+     * @param currPage current page requested from ui
+     * @return map of required attributes
+     */ 
+    @Override
+    public Map<String, Object> getImageListingAttr(int currPage) {
+        long totalImages = imageRepo.count();
```
jdamerow: what if current page provided is -1? you seem to pass that through unchecked.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
```diff
@@ -62,4 +78,23 @@ public ImageData getImageDimensions(IVSImage image, int width, int height) {
...
+    
+    
+    /**
+     * Method to return the required attributes of a requested page  
+     * 
+     * @param currPage current page requested from ui
+     * @return map of required attributes
+     */ 
+    @Override
+    public Map<String, Object> getImageListingAttr(int currPage) {
```
jdamerow: spell out the the variable name. there is really no real reason to abbreviate current

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
```diff
@@ -62,4 +78,23 @@ public ImageData getImageDimensions(IVSImage image, int width, int height) {
...
+        long totalImages = imageRepo.count();
+        boolean areImgsPerfPacked = totalImages % pageSize == 0 ? true : false;
```
jdamerow: merge with line 94

vspace/src/main/java/edu/asu/diging/vspace/core/model/SortByField.java
```diff
@@ -0,0 +1,16 @@
...
+    private final String value;
+
+    private SortByField(String value) {
+        this.value = value;
+    }
+
+    @Override
+    public String toString() {
+        return value;
+    }
```
jdamerow: I wouldn't do this. This might have unintended side effects if used in a jsp page. Just add a method to retrieve the value.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceManager.java
```diff
@@ -65,7 +65,10 @@
     @Override
     public CreationReturnValue storeSpace(ISpace space, byte[] image, String filename) {
         IVSImage bgImage = null;
-        List<SpaceDisplay> displays = spaceDisplayRepo.getBySpace(space);
+        List<SpaceDisplay> displays = null;
+        if (space.getId() != null) {
+        	displays = spaceDisplayRepo.getBySpace(space);
+        }
```
jdamerow: indentation still off

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
```diff
@@ -62,4 +78,23 @@ public ImageData getImageDimensions(IVSImage image, int width, int height) {
...
+        long totalPossiblePages = totalImages / pageSize;
+        final long totalPages = areImgsPerfPacked ? totalPossiblePages : totalPossiblePages + 1;
+        Pageable sortByRequestedField = PageRequest.of(currPage-1, pageSize, Sort.by(SortByField.CREATION_DATE.toString()));
+        Page<VSImage> requestedImages = imageRepo.findAll(sortByRequestedField);
+        Map<String, Object> attributesMap = new HashMap<String, Object>(){{put("images", requestedImages.getContent()); put("totalpages", totalPages); put("currentpage", currPage); }};
```
jdamerow: Just return a list of the requested images. the caller of the method knows which page they requested and you could just have a separate method to get the total number of images or pages. Returning a map makes it very error prone as changing the content of the map can be done without any compiler error, which makes it very error prone. hardcoding the keys makes it even worse.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
jdamerow: also, bad method name. Why image listing attr? which attributes?  we want a list of images according to the parameters we hand to the method.

vspace/src/main/webapp/WEB-INF/tiles/skeleton_bt4.jsp
```diff
@@ -64,6 +64,12 @@
                   Modules
                 </a>
               </li>
+              <li class="nav-item">
```
jdamerow: should only be shown for admins and staff

vspace/src/main/webapp/WEB-INF/tiles/skeleton_bt4.jsp
```diff
@@ -64,6 +64,12 @@
...
+                <a class="nav-link ${currentPage == "images" ? "active" : ""}" href="<c:url value="/staff/images/list/1" />">
```
jdamerow: currentPage does not seem to be defined in tiles def (but it should be)

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
```diff
@@ -62,4 +77,38 @@ public ImageData getImageDimensions(IVSImage image, int width, int height) {
         Float newHeight = (new Float(width) / new Float(imageWidth)) * new Float(imageHeight);
         return new ImageData(newHeight.intValue(), width);
     }
+    
+    /**
+     * Method to return the requested images  
+     * 
+     * @param currentPage current page requested from ui
+     * @return list of requested images
+     */
```
jdamerow: what are your assumptions for currentPage? either you need to be very clear about that in your javadoc or you need to check that the value is valid.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
```diff
@@ -62,4 +77,38 @@ public ImageData getImageDimensions(IVSImage image, int width, int height) {
...
+    
+    /**
+     * Method to return the total pages sufficient to display all images  
+     * 
+     * @return totalPages required to display all images in DB
+     */ 
+    @Override
+    public long getTotalPages() {
+        final long totalPages = (imageRepo.count() % pageSize == 0) ? imageRepo.count() / pageSize : (imageRepo.count() / pageSize) + 1;
+        return totalPages;
```
jdamerow: merge with line 101

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ListImagesController.java
```diff
@@ -0,0 +1,32 @@
...
+
+@Controller
+public class ListImagesController {
+
+    @Autowired
+    private IImageService imageService;
+
+    @RequestMapping("/staff/images/list/{page}")
+    public String listSpaces(@PathVariable String page, Model model) {
+        int currentPage;
```
jdamerow: define where needed (meaning swithc line 18 and 19)

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
```diff
@@ -62,4 +77,38 @@ public ImageData getImageDimensions(IVSImage image, int width, int height) {
...
+    
+    /**
+     * Method to return the total image count  
+     * 
+     * @return total count of images in DB
+     */ 
+    @Override
+    public long getTotalImageCount() {
+        return imageRepo.count();
+    }
```
jdamerow: the new methods need test cases

ejmath12:
I have made the changes on the java files and added test functions. Please check the commits and let me know if they need changes.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
jdamerow: Needs to say that the given page number **has** to be between 1 and number of pages. And what happens if it doesn't?

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ImageServiceTest.java
```diff
@@ -61,4 +84,33 @@ public void test_getImageDimensions_fullHeight() {
         Assert.assertEquals(333, data.getWidth());
         Assert.assertEquals(200, data.getHeight());
     }
+    
+    @Test
+    public void test_getTotalPages_success() {
+        when(imageRepo.count()).thenReturn(12L);
+        assertEquals(2, serviceToTest.getTotalPages());
+        verify(imageRepo, times(2)).count();
```
jdamerow: this isn't needed. we don't care how often count is called. If the implementation changes and count is assigned to a variable before doing the calculate, this test should still succeed as long as the result is correct.

jdamerow: what about the case, when there are no images stored yet? there should be a test case, since this is a "special" situation.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ImageServiceTest.java
```diff
@@ -61,4 +84,33 @@ public void test_getImageDimensions_fullHeight() {
...
+        image.setFilename(IMG_FILENAME);
+        image.setFileType(IMG_CONTENT_TYPE);
+        List<VSImage> images = new ArrayList<>();
+        images.add((VSImage)image);
+        Pageable sortByRequestedField = PageRequest.of(0, 10, Sort.by(SortByField.CREATION_DATE.getValue()));
+        when(imageRepo.findAll(sortByRequestedField)).thenReturn(new PageImpl<VSImage>(images));
+        List<VSImage> requestedImages = serviceToTest.getRequestedImages(1);
+        assertEquals(IMG_ID, requestedImages.get(0).getId());
+        verify(imageRepo).findAll(sortByRequestedField);
+    }
```
jdamerow: what if there is an invalid page number given?

vspace/pom.xml
```diff
@@ -183,7 +183,14 @@
             <version>4.12</version>
             <scope>test</scope>
         </dependency>
-
+        <!-- https://mvnrepository.com/artifact/org.springframework/spring-test -->
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-test</artifactId>
+            <scope>test</scope>
+        </dependency>
```
jdamerow: And do you really need this?

ejmath12: Couldnt find a way to pass custom value for page size without using this library.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
jdamerow: just talked abou this

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ListImagesController.java
```diff
@@ -0,0 +1,32 @@
...
+    @Autowired
+    private IImageService imageService;
+
+    @RequestMapping("/staff/images/list/{page}")
+    public String listSpaces(@PathVariable String page, Model model) {
+        long totalPages = imageService.getTotalPages();
+        int currentPage;
+        try {
+            currentPage = Integer.parseInt(page);
+            currentPage = (currentPage<1 || currentPage>totalPages)?1:currentPage;
```
jdamerow: if you check this in the service class, you don't need to do this here anymore

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ImageServiceTest.java
```diff
@@ -61,4 +83,52 @@ public void test_getImageDimensions_fullHeight() {
...
     }
+    
+    @Test
+    public void test_getTotalPages_success() {
+        when(imageRepo.count()).thenReturn(12L);
+        assertEquals(2, serviceToTest.getTotalPages());
+    }
+    
+    @Test
+    public void test_getTotalPages_when_zero_images() {
```
jdamerow: whenZeroImages (camelcase)

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ImageServiceTest.java
```diff
@@ -61,4 +83,52 @@ public void test_getImageDimensions_fullHeight() {
...
+        images.add((VSImage)image);
+        Pageable sortByRequestedField = PageRequest.of(0, 10, Sort.by(SortByField.CREATION_DATE.getValue()));
+        when(imageRepo.findAll(sortByRequestedField)).thenReturn(new PageImpl<VSImage>(images));
+        List<VSImage> requestedImages = serviceToTest.getRequestedImages(1, 0);
+        assertEquals(IMG_ID, requestedImages.get(0).getId());
+        verify(imageRepo).findAll(sortByRequestedField);
+    }
+  
+    @Test
+    public void test_getRequestedImages_invalid_page() {
```
jdamerow: camal case (see our style guidelines)

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ImageServiceTest.java
```diff
@@ -61,4 +83,52 @@ public void test_getImageDimensions_fullHeight() {
...
+    }
+  
+    @Test
+    public void test_getRequestedImages_invalid_page() { 
+        IVSImage image = new VSImage();
+        image.setId(IMG_ID);
+        image.setFilename(IMG_FILENAME);
+        image.setFileType(IMG_CONTENT_TYPE);
+        List<VSImage> images = new ArrayList<>();
+        images.add((VSImage)image);
```
jdamerow: since you need this code several times, put it in the setup method

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ImageServiceTest.java
jdamerow: also, there are two cases for this < 1 and > total pages

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ImageServiceTest.java
```diff
@@ -61,4 +83,52 @@ public void test_getImageDimensions_fullHeight() {
...
+        Pageable sortByRequestedField = PageRequest.of(0, 10, Sort.by(SortByField.CREATION_DATE.getValue()));
+        when(imageRepo.findAll(sortByRequestedField)).thenReturn(new PageImpl<VSImage>(images));
+        List<VSImage> requestedImages = serviceToTest.getRequestedImages(-2, 0);
+        assertEquals(IMG_ID, requestedImages.get(0).getId());
+        verify(imageRepo).findAll(sortByRequestedField);
+    }
+    
+    @Test
+    public void test_getTotalImageCount_success() {
```
jdamerow: what if there are no images

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ImageServiceTest.java
```diff
@@ -61,4 +90,76 @@ public void test_getImageDimensions_fullHeight() {
...
+        assertEquals(0, serviceToTest.getTotalPages());
+    }
+   
+    @Test
+    public void test_getImages_success() { 
+        Pageable sortByRequestedField = PageRequest.of(0, 10, Sort.by(SortByField.CREATION_DATE.getValue()));
+        when(imageRepo.count()).thenReturn(1L);
+        when(imageRepo.findAll(sortByRequestedField)).thenReturn(new PageImpl<VSImage>(images));
+        List<VSImage> requestedImages = serviceToTest.getImages(1);
+        assertEquals(IMG_ID, requestedImages.get(0).getId());
```
jdamerow: you also want to check that only one image is returned

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ImageServiceTest.java
```diff
@@ -61,4 +90,76 @@ public void test_getImageDimensions_fullHeight() {
...
+    }
+    
+    @Test
+    public void test_getImages_pageGreaterThanTotalPages() { 
+        Pageable sortByRequestedField = PageRequest.of(0, 10, Sort.by(SortByField.CREATION_DATE.getValue()));
+        when(imageRepo.count()).thenReturn(1L);
+        when(imageRepo.findAll(sortByRequestedField)).thenReturn(new PageImpl<VSImage>(images));
+        List<VSImage> requestedImages = serviceToTest.getImages(5);
+        assertEquals(IMG_ID, requestedImages.get(0).getId());
+        verify(imageRepo).findAll(sortByRequestedField);
```
jdamerow: to really test this you need more than one page. there could be a bug in which, the first page would requested when the given page number is greater than the total. You wouldn't catch this with this test.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ImageServiceTest.java
```diff
@@ -61,4 +90,76 @@ public void test_getImageDimensions_fullHeight() {
...
+    @Test
+    public void test_validatePageNumber_whenPageIsNegative() {
+        when(imageRepo.count()).thenReturn(1L);
+        assertEquals(1, serviceToTest.validatePageNumber(-1));
+    }
+    
+    @Test
+    public void test_validatePageNumber_pageGreaterThanTotalPages() {
+        when(imageRepo.count()).thenReturn(1L);
+        assertEquals(1, serviceToTest.validatePageNumber(5));
```
jdamerow: same as above. you need more than one page for this test to work properly.

jdamerow:
Jenkins, test this please.

jdamerow:
please resolve conflicts

PR #47, opened 2019-04-05
[VSPC-37] - custom access denied page instead of 403 forbidden default page.

ejmath12:


vspace/src/main/java/edu/asu/diging/vspace/web/HomeController.java
```diff
@@ -26,4 +26,9 @@ public String home(Model model) {
     public String login() {
         return "login";
     }
+    
+    @RequestMapping(value = "/403") 
+    public String accessDenied() {
+    	return "accessdenied";
```
jdamerow: use camel case

jdamerow:
ok to test

vspace/src/main/webapp/WEB-INF/views/accessdenied.jsp
```diff
@@ -0,0 +1,11 @@
+<%@ page pageEncoding="UTF-8" %>
+<%@ taglib uri="http://java.sun.com/jsp/jstl/core" prefix="c"%>
+<%@ taglib prefix="sec" uri="http://www.springframework.org/security/tags" %>
+
+
+<div style="padding-top: 60px;">
+
+<h1>You dont seem to have the authority to access the requested page!</h1>
+<h2><a href="<c:url value="/staff/dashboard"/>">Go to dashboard</a></h2>
+</div>
```
jdamerow: Thinking about this, make this page use the exhibition layout and don't provide a link to the dashboard. This page will always be shown, e.g. when a visitor of the exhibition tries to access a restricted page. If they are not logged in they should not see the staff area layout or a link to the dashboard.

jdamerow: Let me correct that, if a user has an account but doesn't not have staff or admin rights, they should not necessarily see the staff area layout.

jdamerow:
let's center that message
image

jdamerow:
let's give that some padding on the sides (use the bootstrap classes).
image

jdamerow:
Jenkins, test this please

PR #50, opened 2019-04-12
Story/vspc 39

ejmath12:


vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceManager.java
```diff
@@ -138,4 +142,24 @@ public ISpace getFullyLoadedSpace(String id) {
...
+        if(id!=null) {
+            if(!spaceRepo.findById(id).isPresent()){
+                return CustomResponses.NOT_FOUND;
+            } else {
+                spaceRepo.deleteById(id);
+                return CustomResponses.SUCCESS;
+            }
+        } else {
+            return CustomResponses.NULL_INPUT;
+        }
```
jdamerow: this method does not need to return anything. if someone tries to delete a space that's not there, that's fine. just don't do anything. if it was successfully deleted, great, don't return anything. if id is null, just throw an exception since it clearly is an invalid input.

vspace/src/main/java/edu/asu/diging/vspace/web/CustomResponses.java
```diff
@@ -0,0 +1,20 @@
+package edu.asu.diging.vspace.web;
+
+public enum CustomResponses {
```
jdamerow: unnecessary (see above), but if it would  be needed, this is a bad class (enum) name.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceController.java
```diff
@@ -5,28 +5,34 @@
...
+        ISpace space = spaceManager.getFullyLoadedSpace(id);
+        model.addAttribute("space", space);
+        model.addAttribute("spaceLinks", linkManager.getSpaceLinkDisplays(id));
+        model.addAttribute("externalLinks", linkManager.getExternalLinkDisplays(id));
+        model.addAttribute("spaces", spaceManager.getAllSpaces());
+        return "staff/space";
+    }
+
+    @RequestMapping(value = "/staff/space/{id}", method = RequestMethod.DELETE)
+    public @ResponseBody String deleteSpace(@PathVariable String id) {
```
jdamerow: put in it's own controller

vspace/src/main/webapp/WEB-INF/tiles/skeleton_bt4.jsp
```diff
@@ -93,7 +93,7 @@
         </main>
       </div>
     </div>
-    
+    <tiles:insertAttribute name="deleteModal" />
```
jdamerow: this would make a modal delete attribute available in all pages, which is unnecessary and doesn't make sense. It just blows up the list of attributes. put it where it is actually needed.

jdamerow: if you don't want to put the code inside the page itself, you can always include it directly in the jsp page via an include.

vspace/src/main/webapp/WEB-INF/views/deleteModal.jsp
```diff
@@ -0,0 +1,49 @@
...
+<script>
+	$(document).ready(function() {
+		$('#confirm-delete').on('click', '.btn-ok', function(e) {
+			var url = $(this).data('url');
+			var urlToLoadOnSuccess = $(this).data('urlToLoadOnSuccess');
+			$.ajax({
+				url : url,
+				type : 'DELETE',
+				success : function(response) {
+					console.log(response);
```
jdamerow: remove

vspace/src/main/webapp/WEB-INF/views/deleteModal.jsp
```diff
@@ -0,0 +1,49 @@
...
+						$('#closeButton').click();
+						$('#errorMessage').find("strong").text(response+", please reload the page and try again");
+						$('#errorMessage').show();							
+					}
+				}
+			});
+		});
+		$('#confirm-delete').on('show.bs.modal', function(e) {
+			var data = $(e.relatedTarget).data();
+			console.log(data);
```
jdamerow: remove

vspace/src/main/webapp/WEB-INF/views/deleteModal.jsp
```diff
@@ -0,0 +1,49 @@
...
+	});
+</script>
+<div class="modal fade" id="confirm-delete" tabindex="-1" role="dialog"
+	aria-labelledby="myModalLabel" aria-hidden="true">
+	<div class="modal-dialog">
+		<div class="modal-content">
+			<div class="modal-header">
+				<button type="button" class="close" data-dismiss="modal"
+					aria-hidden="true">×</button>
+				<h4 class="modal-title" id="myModalLabel"></h4>
```
jdamerow: let's change those default ids. if everybody leaves then when copying the code from bootstrap, we'll end up with duplicate ids.

jdamerow: Also, let's add a heading (e.g. "Confirm Space Deletion")

vspace/src/main/webapp/WEB-INF/views/deleteModal.jsp
```diff
@@ -0,0 +1,49 @@
...
+			</div>
+			<div class="modal-body">
+				<p>Are you sure you want to delete?</p>
```
jdamerow: should say which space the user is about to delete.

vspace/src/main/webapp/WEB-INF/views/staff/spaces/space.jsp
```diff
@@ -414,7 +414,12 @@ $( document ).ready(function() {
   <br>
   Modified on <span class="date">${space.modificationDate}</span> by ${space.modifiedBy}.     
 </div>
-
+<div id="errorMessage" class="alert alert-danger alert-dismissible fade show" role="alert" style="display: none;">
+  <strong></strong>
+  <button type="button" class="close" data-dismiss="alert" aria-label="Close">
+    <span aria-hidden="true">&times;</span>
+  </button>
+</div>
```
jdamerow: there is an error alert box in the skeleton file. Let's use that one instead (you probably have to adjust the if structure in the skeleton, so that the alert is always there, just not always shown).

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DeleteSpaceController.java
```diff
@@ -0,0 +1,28 @@
...
+public class DeleteSpaceController {
+    
+    @Autowired
+    private ISpaceManager spaceManager;
+
+    @RequestMapping(value = "/staff/space/{id}", method = RequestMethod.DELETE)
+    public ResponseEntity<?> deleteSpace(@PathVariable String id) {
+        try {
+            spaceManager.deleteSpaceById(id);
+        } catch(Exception exception) {
```
jdamerow: what exception is caught here? never catch Exception if you can catch a more specific one.

jdamerow: Especially if you don't log the exception. This could be any kind of exception and there would be no way for us to find out what actually went wrong.

vspace/src/main/webapp/WEB-INF/views/staff/spaces/space.jsp
jdamerow: do you still need this?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DeleteSpaceController.java
```diff
@@ -0,0 +1,29 @@
...
+public class DeleteSpaceController {
+    
+    @Autowired
+    private ISpaceManager spaceManager;
+
+    @RequestMapping(value = "/staff/space/{id}", method = RequestMethod.DELETE)
+    public ResponseEntity<?> deleteSpace(@PathVariable String id) {
+        try {
+            spaceManager.deleteSpaceById(id);
+        } catch(IllegalArgumentException | EmptyResultDataAccessException exception) {
```
jdamerow: we just talked about that

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceManager.java
```diff
@@ -141,4 +142,22 @@ public ISpace getFullyLoadedSpace(String id) {
...
+     *            if id is null throws exception, else delete corresponding
+     *            space
+     * @throws SpaceDoesNotExistException 
+     */
+    @Override
+    public void deleteSpaceById(String id) throws SpaceDoesNotExistException {
+        try {
+            spaceRepo.deleteById(id);
+        } catch (IllegalArgumentException | EmptyResultDataAccessException exception) {
+            throw new SpaceDoesNotExistException();
```
jdamerow: you need to pass exception, or the stack trace is lost.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SpaceManagerTest.java
```diff
@@ -106,4 +106,12 @@ public void test_storeSpace_success() throws FileStorageException {
         Mockito.verify(spaceRepo).save(space);
         Mockito.verify(spaceDisplayRepo).save((SpaceDisplay)spaceDisplay);
     }
+    
+
+    
+    @Test(expected = IllegalArgumentException.class)
```
jdamerow: this test should fail now

jdamerow:
Jenkins, test this please.

jdamerow:
I'll merge this so Bailey can start using it, but it needs some modifications. I'll create a new ticket for that though.

PR #55, opened 2019-04-18
Story/vspc 45

ejmath12:


vspace/src/main/java/edu/asu/diging/vspace/core/factory/impl/ImageFactory.java
```diff
@@ -23,4 +35,28 @@ public IVSImage createImage(String filename, String fileType) {
...
+     * 
+     * @param imageId - image unique identifier
+     * @param imageForm - ImageForm with updated values for image fields
+     * @return throws FileNotFoundException if no image exists with id, 
+     * throws FileStorageException if file renaming fails 
+     */ 
+    @Override
+    public void editImage(String imageId, ImageForm imageForm) throws FileNotFoundException, FileStorageException {
+        if (imageRepo.findById(imageId).isPresent()) {
+            VSImage image = imageRepo.findById(imageId).get();
```
jdamerow: I assume that findById is cached but I'm not sure. Have you checked if this hits the DB once or twice? If it's twice you should move that line up and use the image in your if statement.

vspace/src/main/java/edu/asu/diging/vspace/core/factory/impl/ImageFactory.java
```diff
@@ -23,4 +35,28 @@ public IVSImage createImage(String filename, String fileType) {
...
+            if (storage.renameImage(image, imageForm.getFileName())) {
```
jdamerow: filenames do not need to be changable at this moment, only name and description.

vspace/src/main/java/edu/asu/diging/vspace/core/factory/impl/ImageFactory.java
```diff
@@ -23,4 +35,28 @@ public IVSImage createImage(String filename, String fileType) {
...
+                image.setFilename(imageForm.getFileName());
+                image.setDescription(imageForm.getDescription());
+                imageRepo.save(image);
+            } else {
+                throw new FileStorageException("File renaming unsuccessful.");
+            }
+        } else {
+            throw new FileNotFoundException();
```
jdamerow: this is an IO exception typically used when files stored in the filesystem or so can't be found. It's confusing why this would be thrown when the entry couldn't be found in the DB>

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditImageController.java
```diff
@@ -0,0 +1,62 @@
...
+import edu.asu.diging.vspace.core.model.impl.VSImage;
+import edu.asu.diging.vspace.web.staff.forms.ImageForm;
+
+@Controller
+public class EditImageController {
+
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+    
+    @Autowired
+    private ImageRepository imageRepo;
```
jdamerow: your controller should not depend on the repository, just on the manager. Your manager is the connection to the repository to keep it loosely coupled.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditImageController.java
```diff
@@ -0,0 +1,62 @@
...
+        try {
+            imageFactory.editImage(imageId, imageForm);
+        } catch (FileNotFoundException | FileStorageException exception) {
+            logger.error("Edit Image Failed" + exception);
+            attributes.addAttribute("alertType", "danger");
+            attributes.addAttribute("showAlert", "true");
+            if (exception instanceof FileStorageException)
+                attributes.addAttribute("message", "Error occured while renaming file name. Please try again.");
+            else
+                attributes.addAttribute("message", "The image id provided is invalid. Please try again");
```
jdamerow: please follow our styleguide

vspace/src/main/java/edu/asu/diging/vspace/core/factory/impl/ImageFactory.java
```diff
@@ -23,4 +35,28 @@ public IVSImage createImage(String filename, String fileType) {
...
+    /**
+     * Method to rename image   
+     * 
+     * @param imageId - image unique identifier
+     * @param imageForm - ImageForm with updated values for image fields
+     * @return throws FileNotFoundException if no image exists with id, 
+     * throws FileStorageException if file renaming fails 
+     */ 
+    @Override
+    public void editImage(String imageId, ImageForm imageForm) throws FileNotFoundException, FileStorageException {
```
jdamerow: factories only create objects, they shouldn't do anything else. the methods you put here are methods for manager/service classes.

jdamerow:
Jenkins test this please.

PR #74, opened 2019-05-21
Story/vspc 20

tjquinn1:
added functionality to delete image and text blocks. Added alerts for when adding and deleting blocks fails. moved block UI creation to ajax success method so it is only created if everything works and does not need to be removed again in the fail case.


jdamerow:
please resolve conflicts

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/Sequence.java
```diff
@@ -22,18 +26,21 @@
...
         strategy = "edu.asu.diging.vspace.core.data.IdGenerator")
     private String id;
+    
+    @OneToOne(targetEntity = Module.class)
+    private IModule module;
 
-    @OneToMany(targetEntity = Slide.class)
-    private List<ISlide> slides;
-
+    @ManyToMany(targetEntity = Slide.class, cascade=CascadeType.ALL)
```
jdamerow: you do not want to delete or save all slides when the sequence is saved/deleted. So this should not be cascade type all.

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/Sequence.java
```diff
@@ -22,18 +26,21 @@
...
+    @OrderColumn(name="slide_order")
+    List<ISlide> slides = new ArrayList<>();
```
jdamerow: should be private

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ModuleManager.java
```diff
@@ -66,4 +71,12 @@ public IModule getModule(String id) {
         IModule module = getModule(moduleId);
         return new ArrayList<>(slideRepo.findByModule((Module) module));
     }
+
+
+    @Override
+    public List<ISequence> getModuleSequences(String moduleId) {
+        IModule module = getModule(moduleId);
```
jdamerow: what if module is null?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SequenceManager.java
```diff
@@ -0,0 +1,66 @@
...
+    private SlideRepository slideRepo;
+
+    @Override
+    public ISequence storeSequence(String moduleId, SequenceForm sequenceForm) {
+        String[] orderedSlideIds = sequenceForm.getOrderedSlideIds().split(",");
+        List<String> slideIds = Arrays.asList(orderedSlideIds);
+        Map<String, Integer> idOrderMap = new HashMap<>();
+        for (int i = 0; i < orderedSlideIds.length; ++i) {
+            idOrderMap.put(orderedSlideIds[i], i);
+        }
```
jdamerow: why do you need a map here? you could probably just use indexOf of the list in line 52.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -41,4 +45,9 @@ public ISlide getSlide(String slideId) {
         }
         return null;
     }
+
+    @Override
+    public List<ISlide> getSlidesInSequence(String sequenceId) {
```
jdamerow: why is this method necessary? you could just do sequence.getSlides ?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/forms/SequenceForm.java
```diff
@@ -0,0 +1,27 @@
+package edu.asu.diging.vspace.web.staff.forms;
+
+public class SequenceForm {
+
+    private String name;
+    private String description;
+    private String orderedSlideIds;
```
jdamerow: why is this a string and not a list?

vspace/.gitignore
```diff
@@ -1,4 +1,8 @@
 /target/
 /Users/
 /.settings/
+<<<<<<< HEAD
+/.DS_Store
+=======
 .DS_Store
+>>>>>>> 0a1c4ae8f231ae79b85b86a7bc6afb794e5af3e0
```
jdamerow: needs to be fixed

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/add.jsp
```diff
@@ -0,0 +1,84 @@
...
+			</select>
+		<form:input type="hidden" id="orderedSlides" path="orderedSlides"></form:input>
+	</div>
+	<button class="btn btn-primary btn-sm" type="submit" value="submit">Create
+		Sequence</button>
+	
+</form:form>
+
+</body>
+</html>
```
jdamerow: why are there closing body and html tags?

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
```diff
@@ -0,0 +1,44 @@
+<%@ page pageEncoding="UTF-8" %>
+<%@ taglib uri="http://java.sun.com/jsp/jstl/core" prefix="c"%>
+<%@ taglib uri="http://java.sun.com/jsp/jstl/functions" prefix="fn" %>
+<%@ taglib prefix="form" uri="http://www.springframework.org/tags/form"%>
+
+<body>
```
jdamerow: why is there a body tag?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SequenceManager.java
```diff
@@ -0,0 +1,55 @@
...
+import edu.asu.diging.vspace.core.factory.ISequenceFactory;
+import edu.asu.diging.vspace.core.model.ISequence;
+import edu.asu.diging.vspace.core.model.ISlide;
+import edu.asu.diging.vspace.core.model.impl.Sequence;
+import edu.asu.diging.vspace.core.services.ISequenceManager;
+import edu.asu.diging.vspace.web.staff.forms.SequenceForm;
+
+@Transactional
+@Service
+public class SequenceManager implements ISequenceManager {
```
jdamerow: let's write test cases for this class

vspace/src/main/java/edu/asu/diging/vspace/core/factory/impl/SequenceFactory.java
```diff
@@ -0,0 +1,32 @@
...
+public class SequenceFactory implements ISequenceFactory{
+    /*
+     * (non-Javadoc)
+     * 
+     * @see
+     * edu.asu.diging.vspace.core.factory.impl.IImageFactory#createImage(java.lang.
+     * String, java.lang.String)
+     */
+    @Override
+    public ISequence createSequence(IModule module, SequenceForm sequenceForm, List<ISlide> slides) {
```
jdamerow: let's write test cases

Neehaarika:
wrote test cases.

vspace/.gitignore
jdamerow: still here

vspace/src/main/webapp/WEB-INF/views/staff/modules/module.jsp
```diff
@@ -1,66 +1,90 @@
...
+	Created on <span class="date">${module.creationDate}</span> by
+	${module.createdBy}. <br> Modified on <span class="date">${module.modificationDate}</span>
+	by ${module.modifiedBy}.
 </div>
-<h5>Description:</h5>
-<p>${module.description}</p>
-	
-
 
 <body>
```
jdamerow: there shouldn't be a body tag

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SequenceFactoryTest.java
```diff
@@ -0,0 +1,60 @@
...
+    @InjectMocks
+    private SequenceFactory sequenceFactoryToTest = new SequenceFactory();
+    
+    @Before
+    public void init() {
+        MockitoAnnotations.initMocks(this);
+    }
+    
+    @Test
+    public void createSequence_Success() {
```
jdamerow: 'success' should be lowercase

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SequenceFactoryTest.java
```diff
@@ -0,0 +1,60 @@
...
+        sequenceForm.setDescription("sample description");
+        List<String> orderedSlides = new ArrayList<>();
+        orderedSlides.add("SLI1");
+        orderedSlides.add("SLI2");
+        sequenceForm.setOrderedSlides(orderedSlides);
+        
+        List<ISlide> slides = new ArrayList<>();
+        for(String slideId : sequenceForm.getOrderedSlides()) {
+            slides.add(slideManager.getSlide(slideId));
+        }
```
jdamerow: why are you doing this? slideManager.getSlide would return null since you haven't told mockito what to do.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SequenceFactoryTest.java
```diff
@@ -0,0 +1,60 @@
...
+        
+        ISequence newSequence = new Sequence();      
+        newSequence.setName(sequenceForm.getName());;
+        newSequence.setDescription(sequenceForm.getDescription());
+        newSequence.setSlides(slides);
+        newSequence.setModule(module);       
+        
+        ISequence actualSequence = sequenceFactoryToTest.createSequence(module, sequenceForm, slides);
+        Assert.assertEquals(newSequence.getId(), actualSequence.getId());
```
jdamerow: you're setting all these values for the new sequence, but you're only testing that the id is correct. you should make sure all values that you put are set correctly. And there is no point in creating a newSequence objects if it's just for comparing values. You can just compare the actual sequence value with the ones you've set earlier (e.g. using the form or string variables).

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SequenceManagerTest.java
```diff
@@ -0,0 +1,90 @@
...
+    private ModuleManager moduleManager;
+    
+    @Mock
+    private SlideManager slideManager;
+    
+    @Mock
+    private SequenceFactory sequenceFactory;
+    
+    @InjectMocks
+    private SequenceManager sequenceToTest = new SequenceManager();
```
jdamerow: you're not testing a sequence but the sequenceManager, so your variable name should reflect that.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SequenceManagerTest.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    @Test
+    public void test_getSequence_idExists() {
+        Sequence newSequence = new Sequence();
+        newSequence.setId("sequence1");
+        Optional<Sequence> mockSequence = Optional.of(newSequence);
+        Mockito.when(mockSequenceRepo.findById(newSequence.getId())).thenReturn(mockSequence);
+
+        ISequence iSequenceActual = sequenceToTest.getSequence(newSequence.getId());
```
jdamerow: why does the variable have an 'i' as prefix?

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SequenceFactoryTest.java
jdamerow: also, where is the 'test_' prefix?

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SequenceManagerTest.java
```diff
@@ -0,0 +1,90 @@
...
+        Assert.assertEquals(mockSequence.get().getId(), iSequenceActual.getId());
+        Assert.assertEquals(mockSequence.get().getName(), iSequenceActual.getName());
```
jdamerow: why don't you use ```newSequence``` instead of ```mockSequence.get()```?

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SequenceManagerTest.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+    
+    @Test
+    public void test_getSequence_idNotExists() throws Exception {
+        Mockito.when(mockSequenceRepo.findById(Mockito.anyString())).thenReturn(Optional.empty());
+        Assert.assertNull(sequenceToTest.getSequence(Mockito.anyString()));
```
jdamerow: ```Mockito.anyString()``` isn't really meant for this case and returns an empty string, which is kind of a special case. just pass an actual string.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SequenceManagerTest.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+    
+    @Test
+    public void test_storeSequence_Success() {
```
jdamerow: 'success'

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SequenceManagerTest.java
```diff
@@ -0,0 +1,90 @@
...
+        orderedSlides.add("SLI2");
+        seqForm.setOrderedSlides(orderedSlides);
+        
+        List<ISlide> slides = new ArrayList<>();
+        for(String slideId : seqForm.getOrderedSlides()) {
+            slides.add(slideManager.getSlide(slideId));
+        }
+        
+        Sequence newSequence = new Sequence(); 
+        Mockito.when((Sequence) sequenceFactory.createSequence(moduleManager.getModule(module.getId()), seqForm, slides)).thenReturn(newSequence);
```
jdamerow: why are you casting here? Does it require that?

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SequenceManagerTest.java
```diff
@@ -0,0 +1,90 @@
...
+        Sequence mockSequence = new Sequence();
+        Mockito.when(mockSequenceRepo.save(newSequence)).thenReturn(mockSequence);
+        
+        ISequence iSequenceActual = sequenceToTest.storeSequence(module.getId(), seqForm);
+        Assert.assertEquals(mockSequence.getId(), iSequenceActual.getId());
```
jdamerow: you never set mockSequence's id, so it would be null. What are you testing?

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SequenceManagerTest.java
```diff
@@ -0,0 +1,90 @@
...
+        Assert.assertEquals(mockSequence.getName(), iSequenceActual.getName());
```
jdamerow: same here.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SequenceManagerTest.java
Neehaarika: Can I not use either of them?

jdamerow: ya, it's just an unnecessary method call since you already have the object you need.

PR #78, opened 2019-06-20
Story/vspc 51

tjquinn1:


vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ContentBlockManager.java
```diff
@@ -125,4 +126,55 @@ public CreationReturnValue createImageBlock(String slideId, byte[] image, String
...
+            } catch (FileStorageException e) {
+                throw new ImageCouldNotBeStoredException(e);
+            }
+            slideContentImage.setParentPath(relativePath);
+            imageRepo.save((VSImage) slideContentImage);
+        }
+
+        imageBlock.setImage(slideContentImage);
+        imageBlockRepo.save((ImageBlock) imageBlock);
+    }
```
jdamerow: a lot of this looks like the code for creating an image block; let's put that into methods and get rid of the code duplication.

vspace/src/main/webapp/WEB-INF/views/staff/modules/slides/slide.jsp
```diff
@@ -89,96 +106,222 @@ $(document).ready(function() {
...
+            $('<div class="col-xs-1" style="margin-top: 1%"><a id="cancelTextBlock" class="btn" href="#"style="float: right;"><i class="fas fa-times"></i></a><a id="submitTextBlock" class="btn" href="#"style="float: right;"><i class="fas fa-check"></i></a></div>').insertAfter( "#newTextBlockDiv" );
+            $("#" + blockId + " p").remove();
+            // unbind events to prevent multiple instances of buttons and constant highlighting
+            $('.valueDiv').unbind('mouseenter mouseleave dblclick');
+            // remove highlighting if present
+            $(this).removeClass("hova");
+        } else {
+        	// Must use closest('div').children() instead of closet('img').
+        	blockId = $(e.target).closest('div').children().attr('id');
+        	console.log(blockId);
```
jdamerow: remove

vspace/src/main/webapp/WEB-INF/views/staff/modules/slides/slide.jsp
```diff
@@ -89,96 +106,222 @@ $(document).ready(function() {
...
+            $("#addImgAlert").show();
+            localStorage.setItem('imgBlockId', blockId);
```
jdamerow: is there a reason you not just adding the block id as attribute to the submit button or alert? using local storage seems a little overkill and has higher bugginess potential.

tjquinn1: I use the block ids in order to keep track of which block I am appending and deleting UI components from as well as in the form submission. I can remove it from the submissions but it will still be used else where to manipulate the UI. Let me know what you would prefer.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ContentBlockManager.java
```diff
@@ -81,31 +82,21 @@ public ITextBlock createTextBlock(String slideId, String text, Integer contentOr
...
-     * (non-Javadoc)
-     * 
-     * @see edu.asu.diging.vspace.core.services.impl.IContentBlockManager#
-     * createImageBlock(java.lang.String, java.util.Arrays, java.lang.String)
-     */
-    @Override
-    public CreationReturnValue createImageBlock(String slideId, byte[] image, String filename, Integer contentOrder) throws ImageCouldNotBeStoredException {
-        IVSImage slideContentImage = null;
-        ISlide slide = slideManager.getSlide(slideId);
+    private IVSImage saveImage(byte[] image, IVSImage slideContentImage, String filename) {
```
jdamerow: ```slideContentImage``` always seems to be null and even if it weren't you're replacing whatever it is in line 92. That doesn't make sense, why do you pass it at all? Just create it and return.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ContentBlockManager.java
```diff
@@ -81,31 +82,21 @@ public ITextBlock createTextBlock(String slideId, String text, Integer contentOr
...
             slideContentImage = imageFactory.createImage(filename, contentType);
             slideContentImage = imageRepo.save((VSImage) slideContentImage);
         }
+        return slideContentImage;
+    }
 
-        CreationReturnValue returnValue = new CreationReturnValue();
-        returnValue.setErrorMsgs(new ArrayList<>());
-
+    private IVSImage handleNullContent(byte[] image, IVSImage slideContentImage, String filename)
```
jdamerow: you actually handle the case that the content isn't null, don't you?

jdamerow: also, since you are passing ```slideContentImage``` there is no need to return it as well since you just modify it's attributes.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditTextBlockController.java
```diff
@@ -0,0 +1,36 @@
...
+import edu.asu.diging.vspace.core.model.impl.TextBlock;
+import edu.asu.diging.vspace.core.services.IContentBlockManager;
+
+@Controller
+public class EditTextBlockController {
+
+    @Autowired
+    private IContentBlockManager contentBlockManager;
+
+    @RequestMapping(value = "/staff/module/{moduleId}/slide/{id}/textcontent/edit", method = RequestMethod.POST)
```
jdamerow: I would make 'textcontent' just 'text' since it's also 'image' and not 'imagecontent'

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditTextBlockController.java
```diff
@@ -0,0 +1,36 @@
...
+    public ResponseEntity<String> editTextBlock(@PathVariable("id") String slideId,
+            @RequestParam("textBlockId") String blockId, @PathVariable("moduleId") String moduleId,
+            @RequestParam("textBlockDesc") String textBlockDesc) throws IOException {
+
+        ITextBlock textBlock = contentBlockManager.getTextBlock(blockId);
+        textBlock.setText(textBlockDesc);
+        contentBlockManager.updateTextBlock((TextBlock) textBlock);
+
+        return new ResponseEntity<String>(HttpStatus.OK);
```
jdamerow: editing an image, returns "ok". should be consistent. (I would not return anything when editing images, since "ok" is not valid json.

vspace/src/main/webapp/WEB-INF/views/staff/modules/slides/slide.jsp
```diff
@@ -5,28 +5,42 @@
 <script>
 //# sourceURL=click.js
 var contentCount = ${fn:length(slideContents)};
+
+function createImageBlock(reader) {
+	var imageblock = $('<img src="#" />');
```
jdamerow: indentation

vspace/src/main/webapp/WEB-INF/views/staff/modules/slides/slide.jsp
```diff
@@ -5,28 +5,42 @@
...
 	var formData = new FormData();
 	formData.append('file', file);
 	formData.append('contentOrder', contentCount);
+	if (localStorage.getItem('imgBlockId')){
+		formData.append('imageBlockId', localStorage.getItem('imgBlockId'));
+        var url = "<c:url value="/staff/module/${module.id}/slide/${slide.id}/image/" />" + localStorage.getItem('imgBlockId') + "?${_csrf.parameterName}=${_csrf.token}";
+        reader.onload = function () {
+        	imageblock = createImageBlock(reader);
+            $("#" + localStorage.getItem('imgBlockId')).replaceWith(imageblock);
+        }
```
jdamerow: indentation

vspace/src/main/webapp/WEB-INF/views/staff/modules/slides/slide.jsp
jdamerow: I would reevaluate if you really need local storage. You can probably get most things done without using local storage. What exactly are you keeping track of? can't you just add ids to the elements and use them when clicking on buttons to delete something?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ContentBlockManager.java
tjquinn1: I am not sure I follow you about about slideContentImage always being null. Do you mean when it is passed into the function? I passed it in so that I did not need to declare slideContentImage in the both the function scope and the outer scope.

vspace/src/main/webapp/WEB-INF/views/staff/modules/slides/slide.jsp
```diff
@@ -5,136 +5,197 @@
...
-		
-		success: function(data) {
-			// do nothing for now
-		},
-		error: function(data) {
-		}
-	});
+    var file = document.getElementById('file').files[0];
+    var reader  = new FileReader();
+    var file = document.getElementById('file').files[0];
```
jdamerow: duplicate line?

jdamerow:
Jenkins, test this please.

jdamerow:
When I double click on a text block and change the test and then cancel editing (clicking the 'x'), it keeps the modified text (although it doesn't save it).

PR #80, opened 2019-07-11
Story/vspc 20

Neehaarika:


jdamerow:
Jenkins, test this please.

jdamerow:
there are test failures.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SequenceManager.java
```diff
@@ -0,0 +1,55 @@
...
+    private SequenceRepository sequenceRepo;
+
+    @Override
+    public ISequence storeSequence(String moduleId, SequenceForm sequenceForm) {
+        List<ISlide> slides = new ArrayList<>();
+        for(String slideId : sequenceForm.getOrderedSlides()) {
+            slides.add(slideManager.getSlide(slideId));
+        }
+        ISequence sequence = sequenceFactory.createSequence(moduleManager.getModule(moduleId), sequenceForm, slides);
+        sequenceRepo.save((Sequence) sequence);
```
jdamerow: you probably want to return the sequence object that ```save``` returns. I believe the repo creates a new sequence object with id and does not set the id for the passed object.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SequenceFactoryTest.java
```diff
@@ -0,0 +1,53 @@
...
+import edu.asu.diging.vspace.web.staff.forms.SequenceForm;
+
+public class SequenceFactoryTest {
+
+    @InjectMocks
+    private SequenceFactory sequenceFactoryToTest = new SequenceFactory();
+    
+    @Before
+    public void init() {
+        MockitoAnnotations.initMocks(this);
```
jdamerow: since you are not using any mocks in this test class you can remove this line and the ```injectMocks``` annotation.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SequenceManagerTest.java
```diff
@@ -0,0 +1,113 @@
...
+    @Before
+    public void init() {
+        MockitoAnnotations.initMocks(this);
+    }
+
+    @Test
+    public void test_getSequence_idExists() {
+        Sequence newSequence = new Sequence();
+        newSequence.setId("sequence1");
+        Optional<Sequence> mockSequence = Optional.of(newSequence);
```
jdamerow: this is not a mocked sequence. just a sequence. the variable name is confusing.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SequenceManagerTest.java
```diff
@@ -0,0 +1,113 @@
...
+        sequenceForm.setOrderedSlides(slideIds);
+        
+        ISlide slide1 = new Slide();
+        slide1.setId("slide1");       
+        ISlide slide2 = new Slide();
+        slide2.setId("slide2");
+ 
+        List<ISlide> slides = new ArrayList<>();
+        slides.add(slide1);
+        slides.add(slide2);
```
jdamerow: I believe there is a ```Arrays.asList()``` method that just takes several objects and returns a list of them. This would turn these three lines into a one-liner.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SequenceManagerTest.java
```diff
@@ -0,0 +1,113 @@
...
+        Mockito.when(sequenceFactory.createSequence(module, sequenceForm, slides)).thenReturn(newSequence);
+        Mockito.when(mockSequenceRepo.save(newSequence)).thenReturn(newSequence);
+        
+        ISequence actualSequence = sequenceManagerToTest.storeSequence(module.getId(), sequenceForm);
+
+        Assert.assertEquals(actualSequence.getModule().getId(), module.getId());
+        Assert.assertEquals(actualSequence.getName(), sequenceForm.getName());
+        Assert.assertEquals(actualSequence.getDescription(), sequenceForm.getDescription());
+        Assert.assertEquals(actualSequence.getSlides(), slides);
+        Mockito.verify(mockSequenceRepo).save(newSequence);
```
jdamerow: after a sequence has been stored, it should have an id.

jdamerow:
Jenkins, test this please.

jdamerow:
everything else looks good!

jdamerow:
Jenkins, test this please.

vspace/.gitignore
```diff
@@ -1,4 +1,5 @@
 /target/
 /Users/
 /.settings/
-.DS_Store
+/.DS_Store
```
jdamerow: why this change?

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SequenceFactoryTest.java
jdamerow: ?

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SequenceManagerTest.java
```diff
@@ -0,0 +1,105 @@
...
+    @Test
+    public void test_getSequence_idExists() {
+        Sequence newSequence = new Sequence();
+        newSequence.setId("sequence1");
+        Optional<Sequence> sequence = Optional.of(newSequence);
+        Mockito.when(mockSequenceRepo.findById(newSequence.getId())).thenReturn(sequence);
+
+        ISequence actualSequence = sequenceManagerToTest.getSequence(newSequence.getId());
+        Assert.assertEquals(newSequence.getId(), actualSequence.getId());
+        Assert.assertEquals(newSequence.getName(), actualSequence.getName());
```
jdamerow: you are not setting the name of the sequence

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SequenceManagerTest.java
```diff
@@ -0,0 +1,105 @@
...
+        sequenceForm.setDescription("sample description");            
+        List<String> slideIds = Arrays.asList("slide1", "slide2");
+        sequenceForm.setOrderedSlides(slideIds);
+        
+        List<ISlide> slidesList = Arrays.asList(new Slide(), new Slide());
+        slidesList.get(0).setId("slide1");
+        slidesList.get(1).setId("slide2");
+            
+        Sequence newSequence = new Sequence();
+        newSequence.setId("seq01");
```
jdamerow: when ```storeSequence``` is called, the sequence does not have an id yet; the returned sequence has one (that's how you know it was stored). So  ```newSequence``` should not have an id. The sequence object returned by the repo, however, has.

jdamerow:
tests seem to be failing

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SequenceManagerTest.java
Neehaarika: but we are the ones who are telling mockito what to return when save method is hit on the repo right? it is returning newSequence object and Assert.assertNotNull(actualSequence.getId()); is failing as it returns null.

jdamerow: but let's say someone changes the code so that it checks if the ID is set after the factory is called and then returns null because it assumes the object is not a new object, then your test would fail although the code would still be correct. Your test case needs to mock what is actually happening (id = null before save, != null after save). Although, you probably want to test that the id the returned object has is the id that the save method returns, not a different one. Otherwise, if someone would change the code to set the id manually after save is called, your test would pass although the code would be wrong.

jdamerow: the easiest would probably to create two sequence objects, one returned by the factory, one returned by the repo. the other option is to use ```doAnswer``` as described here (with actually exactly our case): https://wahz.blogspot.com/2012/10/mockito-modifying-parameters.html

jdamerow:
Jenkins, test this please.

PR #82, opened 2019-07-23
Story/vspc 51

Neehaarika:


vspace/src/main/webapp/WEB-INF/views/staff/modules/slides/slide.jsp
```diff
@@ -282,9 +281,22 @@ $(document).ready(function() {
         // remove id from storage so its not there on refresh
         $(".open").removeClass("open");
     }
+    
+    function cancelAndCloseTextBox() {
+    	var description = document.getElementById("newTextBlock").defaultValue;;
```
jdamerow: I would just set the value of 'newTextBock' to the old value (so add a line after this to set the value of newTextBlock to description') and then call closeTextBox. Or put all the code below that is the same for 'closeTextBock' in it's own method and remove duplicate code.

jdamerow:
Jenkins test this please.

jdamerow:
Jenkins test this please.

PR #85, opened 2019-09-27
Story/vspc-56

ashmichheda:


vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditSequenceController.java
```diff
@@ -0,0 +1,40 @@
...
+import edu.asu.diging.vspace.core.model.ISequence;
+import edu.asu.diging.vspace.core.model.impl.Sequence;
+import edu.asu.diging.vspace.core.services.ISequenceManager;
+
+@Controller
+public class EditSequenceController {
+
+	
+	@Autowired
+	ISequenceManager sequenceManager;
```
jdamerow: should be private

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditSequenceController.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @RequestMapping(value = "/staff/module/{moduleId}/sequence/{sequenceId}/edit/title", method = RequestMethod.POST)
+    public ResponseEntity<String> saveTitle(@RequestParam("title") String title,
+            @PathVariable("moduleId") String moduleId, @PathVariable("sequenceId") String sequenceId) {
+    	ISequence sequence = sequenceManager.getSequence(sequenceId);
+    	sequence.setName(title);
+        sequenceManager.updateSequence((Sequence) sequence);
+        return new ResponseEntity<String>(HttpStatus.OK);
+    }
+}
```
jdamerow: indention is all over the place

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
```diff
@@ -10,13 +10,168 @@
...
+        
+    });
+   // ------- Edit/Save Description ends --------
+});
+
+
+
+</script>
+
+<!--  <h1>Sequence: ${sequence.name}</h1> -->
```
jdamerow: why is this commented out? if it's not needed anymore, remove.

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
```diff
@@ -10,13 +10,168 @@
...
+        processData : false,
+        data: formData,
+        enctype: 'multipart/form-data',
+        success: function(data) {
+            // replace text box with new description
+            $("#submitDescription").hide()
+            $("#cancelEditDescription").hide()
+            $("#editDescription").show()
+            var val = $("#newDescription").val();
+            console.log("Value of new description: "+val)
```
jdamerow: remove

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
```diff
@@ -10,13 +10,168 @@
...
+            data: formData,
+            enctype: 'multipart/form-data',
+            success: function(data) {
+                // replace text box with new description
+                $("#submitTitle").hide()
+                $("#cancelEditTitle").hide();
+                $("#editTitle").show()
+                var val = $("#newTitle").val();
+                $("#newTitle").closest('div').remove();
+                $("#title").text("Silde: " + val)
```
jdamerow: why does this say "Slide"?

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
```diff
@@ -10,13 +10,168 @@
...
+                $('.error').append(alert);
+            }
+        });
+      });
+   $("#cancelEditTitle").click(function(){
+        $("#submitTitle").hide()
+        $("#editTitle").show()
+        $("#cancelEditTitle").hide()
+        $("#newTitle").closest('div').remove();
+        $("#title").text("Silde: " + getTitleText)
```
jdamerow: why does this say "slide"?

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
```diff
@@ -10,13 +10,168 @@
...
+    >Cancel</button>
+</div>
 <div class="alert alert-light" role="alert">
   Created on <span class="date">${sequence.creationDate}</span> by ${sequence.createdBy}.<br>
   Modified on <span class="date">${sequence.modificationDate}</span> by ${sequence.modifiedBy}.
 </div>
-<h5>Description:</h5>
-<p>${sequence.description}</p>
+ <!-- <h5>Description:</h5>
+<p>${sequence.description}</p>  -->
```
jdamerow: remove if not needed

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SequenceManager.java
```diff
@@ -52,4 +52,9 @@ public ISequence getSequence(String sequenceId) {
         }
         return null;
     }
+
+    @Override
+    public void updateSequence(Sequence sequence) {
```
jdamerow: the interface should be used here. otherwise why would you cast in the next line?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditSequenceController.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Autowired
+    private ISequenceManager sequenceManager;
+
+    @RequestMapping(value = "/staff/module/{moduleId}/sequence/{sequenceId}/edit/description", method = RequestMethod.POST)
+    public ResponseEntity<String> saveDescription(@RequestParam("description") String description,
+            @PathVariable("moduleId") String moduleId, @PathVariable("sequenceId") String sequenceId) {
+        ISequence sequence = sequenceManager.getSequence(sequenceId);
+        sequence.setDescription(description);
+        sequenceManager.updateSequence((Sequence) sequence);
```
jdamerow: should not be cast here

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditSequenceController.java
```diff
@@ -0,0 +1,39 @@
...
+        return new ResponseEntity<String>(HttpStatus.OK);
+    }
+
+    @RequestMapping(value = "/staff/module/{moduleId}/sequence/{sequenceId}/edit/title", method = RequestMethod.POST)
+    public ResponseEntity<String> saveTitle(@RequestParam("title") String title,
+            @PathVariable("moduleId") String moduleId, @PathVariable("sequenceId") String sequenceId) {
+        ISequence sequence = sequenceManager.getSequence(sequenceId);
+        sequence.setName(title);
+        sequenceManager.updateSequence((Sequence) sequence);
```
jdamerow: same here

jdamerow:
Jenkins, test this please.

jdamerow:
When I edit the description, it changes it. but when I edit it again (without reloading the page) it still shows the old description.
- edit description
- click save
- click edit icon again
-> previous description (edit in step 1) is still in input field

jdamerow:
same for title field

jdamerow:
Jenkins, test this please.

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
```diff
@@ -1,23 +1,184 @@
...
+    
+    
+    
+    //------- edit title of Sequence starts --------
+    
+    $("#submitTitle").hide();
+    $("#cancelEditTitle").hide();
+    
+    var getTitleText = $("#title").text().split(": ")[1]
+    console.log("Value of title: "+getTitleText);
```
jdamerow: remove

jdamerow:
Nope, still the same problem. Also there is something weird with the title. It seems to get tabs appended.

jdamerow:
cancel now shows the new entered (cancelled) text, not the actual text.

ashmichheda:
While editing sequence Title, in order to retrieve the title value i have kept the value displayed while edit. Please let me know if this is okay, I will do the same change for **description**.
image

jdamerow:
That'll become very unwieldy for longer titles and especially for descriptions. What's the reasoning for that. It shouldn't be necessary to keep the old title/description around.

ashmichheda:
Fixed the the title and description of sequence. 
Fix - Saved the original values of title and description using value() function for retrieving it when clicked 'cancel' button.

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
```diff
@@ -1,23 +1,200 @@
...
+$(document).ready(function() { 
+ 
+	
+    //-------- edit contentblock description Sequence starts --------
+    $("#submitDescription").hide()
+    $("#cancelEditDescription").hide()	
+    
+    $("#editDescription").click(function() {
+    	var description = $("#description").text()
+    	document.getElementById("description").value = description;
```
jdamerow: we're using jquery. let's not mix jquery and plain js unless we have to.

jdamerow: also, why do you set the value field of the description element?

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
```diff
@@ -1,23 +1,200 @@
...
+        processData : false,
+        data: formData,
+        enctype: 'multipart/form-data',
+        success: function(data) {
+            // replace text box with new description
+            $("#submitDescription").hide()
+            $("#cancelEditDescription").hide()
+            $("#editDescription").show()
+            var val = $("#newDescription").val();
+            $('<p id="description"style="margin-top: .5rem; margin-bottom: .5rem;">val</p>').insertBefore( "#newDescription" );
```
jdamerow: why are you inserting a new element with id "description"? the old one is only hidden, isn't it?

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
```diff
@@ -1,23 +1,200 @@
...
+   });
+    
+    // ---- Cancel Edited Description ----
+    
+    $("#cancelEditDescription").click(function(){
+        $("#submitDescription").hide()
+        $("#editDescription").show()
+        $("#cancelEditDescription").hide()
+        var description = document.getElementById("description").value
+        $('<p id="description" style="margin-top: .5rem; margin-bottom: .5rem;">description</p>').insertBefore( "#newDescription" );
```
jdamerow: same here, what's with the old element with that id?

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
```diff
@@ -1,23 +1,200 @@
...
+    // ------- Edit/Save/Cancel Description ends --------
+    
+    
+    
+    //------- edit title of Sequence starts --------
+    
+    $("#submitTitle").hide();
+    $("#cancelEditTitle").hide();
+    
+    	$("#editTitle").click(function() {
```
jdamerow: indentation

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
```diff
@@ -1,23 +1,200 @@
...
+
+    	var getTitleText = $("#title").text().split(": ")[1].trim()
```
jdamerow: ugly, use data attributes or another cleaner way to get the title

jdamerow: also, bad variable name. get... sounds like a method

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
```diff
@@ -1,23 +1,200 @@
...
+        $('<div class="col-4"><input id="newTitle" class="form-control" type="text"></div>').insertAfter( "#title" );
+    	document.getElementById("title").value = getTitleText;
```
jdamerow: same as above, don't mix jquery and plain js

jdamerow: and why do you set the value attribute?

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
```diff
@@ -1,23 +1,200 @@
...
+            processData : false,
+            data: formData,
+            enctype: 'multipart/form-data',
+            success: function(data) {
+                // replace text box with new description
+                $("#submitTitle").hide()
+                $("#cancelEditTitle").hide();
+                $("#editTitle").show()
+                var val = $("#newTitle").val();
+                $("#newTitle").closest('div').remove();
```
jdamerow: why don't you give the div an id and remove it by id? much less error prone, because what if we need some extra layouting or something and need to add another div around the input field?

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
```diff
@@ -1,23 +1,200 @@
...
+      });
+    
+    
+   // ------  Edit Title Sequence ------
+   
+   $("#cancelEditTitle").click(function(){
+        $("#submitTitle").hide()
+        $("#editTitle").show()
+        $("#cancelEditTitle").hide()
+       	var value = document.getElementById("title").value
```
jdamerow: indentation

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
```diff
@@ -1,23 +1,200 @@
...
+        $("#newTitle").closest('div').remove();
```
jdamerow: same as above

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
```diff
@@ -1,23 +1,200 @@
...
-<p>${sequence.description}</p>
-    
+
+
+
+<!-- description -->
+<div style="margin-left: .1%;" class="row align-items-center">
+	<h4 style="margin-bottom: 0px;">
+		Description:
+		</h5>
```
jdamerow: you are opening an h4 but closing an h5

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
ashmichheda: In the code, value of the description is getting removed. Hence, I stored it in value attribute so that it can be retrieved later. Otherwise, it would be difficult to fetch original value.

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
ashmichheda: I have referred to the functionality implemented for Slides.jsp for it's title and description and have incorporated it for sequence.

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
jdamerow: I don't think it's proper HTML to give p elements a value attribute. Use data attributes instead.

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
jdamerow: comment still applied. if you end up with multiple elements with id "description" something is wrong.

ashmichheda: okay, I will use different id names, I tried with the one that is hidden, but it's not working.

jdamerow: it's not about using a different id but about reusing the existing element.

jdamerow:
what about my other comments about val attributes?

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
```diff
@@ -1,23 +1,208 @@
...
+        data: formData,
+        enctype: 'multipart/form-data',
+        success: function(data) {
+            // replace text box with new description
+            $("#submitDescription").hide()
+            $("#cancelEditDescription").hide()
+            $("#editDescription").show()
+            var val = $("#newDescription").val();
+            $("#editDescription").show()
+            $('<p id="description" style="margin-top: .5rem; margin-bottom: .5rem;">val</p>').insertBefore( "#newDescription" );
```
jdamerow: you never deleted the old description element, did you? wouldn't you end up with multiple? why are you not using the previous one?

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
```diff
@@ -1,23 +1,208 @@
...
+   });
+    
+    // ---- Cancel Edited Description ----
+    
+    $("#cancelEditDescription").click(function(){
+        $("#submitDescription").hide()
+        $("#editDescription").show()
+        $("#cancelEditDescription").hide()
+        var description = $("#description").data('value')
+        $('<p id="editedDescription" style="margin-top: .5rem; margin-bottom: .5rem;">'+description+'</p>').insertBefore( "#newDescription" );
```
jdamerow: why are you creating this if you delete in line 90?

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
```diff
@@ -1,23 +1,208 @@
...
+    //------- edit title of Sequence starts --------
+    
+    $("#submitTitle").hide();
+    $("#cancelEditTitle").hide();
+    
+    $("#editTitle").click(function() {
+    	// Using data attribute
+    	var sequenceTitle = $("#title").data('value')  // gets value without using trim function
+        $('<div class="col-4" id = "editSequenceTitle"><input id="newTitle" class="form-control" type="text"></div>').insertAfter( "#title" );
+    	$("#title").val(sequenceTitle)
```
jdamerow: why are you setting the value here?

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
```diff
@@ -1,23 +1,207 @@
...
+        enctype: 'multipart/form-data',
+        success: function(data) {
+            // replace text box with new description
+            $("#submitDescription").hide()
+            $("#cancelEditDescription").hide()
+            $("#editDescription").show()
+            var val = $("#newDescription").val();
+            $("#editDescription").show()
+            $("#description").show() 
+            $('#description').insertBefore( "#newDescription" );
```
jdamerow: why are you inserted newDescription here?

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
```diff
@@ -1,23 +1,207 @@
...
+            $("#description").text(val);
```
jdamerow: first set the text, then show it. otherwise the user see's the flicker of changing the description.

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
```diff
@@ -1,23 +1,207 @@
...
+            $("#description").data('value',val);
+            $("#newDescription").hide()
```
jdamerow: every time the user edits the description, you create a new textarea and add it to the html. that would result in many textareas that are not used, wouldn't it?

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
```diff
@@ -1,23 +1,207 @@
...
+    // ---- Cancel Edited Description ----
+    
+    $("#cancelEditDescription").click(function(){
+        $("#submitDescription").hide()
+        $("#editDescription").show()
+        $("#cancelEditDescription").hide()
+        var description = $("#description").data('value')
+        $('description').insertBefore( "#newDescription" );
+        $("#newDescription").remove()
+        $("#description").text(description);
```
jdamerow: the description shouldn't have changed, should it? so no need to set the text?

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
```diff
@@ -1,23 +1,207 @@
...
+   });
+    
+    // ---- Cancel Edited Description ----
+    
+    $("#cancelEditDescription").click(function(){
+        $("#submitDescription").hide()
+        $("#editDescription").show()
+        $("#cancelEditDescription").hide()
+        var description = $("#description").data('value')
+        $('description').insertBefore( "#newDescription" );
```
jdamerow: see above

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
jdamerow: ?

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
ashmichheda: Previously, the code was creating new description tag, while editing each time. Hence, as you suggested in the previous pull request to use the original description tag. I have reused the same. It's not creating a new one, just using the "description" tag initially created.

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
ashmichheda: Here the value changes, if I edit any value without saving it. Hence, I had to set the text tag.
I found out while testing it.

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
jdamerow: line 32 creates a new textarea, doesn't it?

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
jdamerow: if you never change the text of the div description, how can it change?

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
ashmichheda: Yes, because while editing we need a text-area to edit. As there is no already created text-area for description. Hence we are creating it. I will close it on submitting, to avoid multiple creations.

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
ashmichheda: Yes, there is no need to set the text. I will remove it.

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
jdamerow: if you delete newDescription in line 68, why inserting it here?

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
jdamerow: see above

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
jdamerow: ?

jdamerow: please clarify/fix

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
```diff
@@ -1,23 +1,206 @@
...
+      });
+    
+    
+   // ------  Edit Title Sequence ------
+   
+   $("#cancelEditTitle").click(function(){
+        $("#submitTitle").hide()
+        $("#editTitle").show()
+        $("#cancelEditTitle").hide()
+        var sequenceTitle = $("#title").val()
```
jdamerow: there is not value attribute for h2

ashmichheda: if you delete newDescription in line 68, why inserting it here?
-- I am not inserting new Description again. I am just added original description tag before "newDescription" using insertBefore and then deleting "newDescription"

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
jdamerow: why? description is already in your html, why would you insert it again?

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
jdamerow: see above (also, this doesn't refer to anything 'description' would refer to a tag, '#description' refers to an element with id description, so if this works, you can delete this line).

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
jdamerow: FIX THIS

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
jdamerow: why adding the original description again? it's already in your html.

jdamerow: h2 does not have a value attribute

ashmichheda:
var sequenceTitle = $("#title").data('value')  // gets value without using trim function
        $('<div class="col-4" id = "editSequenceTitle"><input id="newTitle" class="form-control" type="text"></div>').insertAfter( "#title" );
    	$("#title").val(sequenceTitle)


For the above, what needs to be fixed?

ashmichheda:
var sequenceTitle = $("#title").data('value')  // gets value without using trim function
        $('<div class="col-4" id = "editSequenceTitle"><input id="newTitle" class="form-control" type="text"></div>').insertAfter( "#title" );
    	$("#title").val(sequenceTitle)

ashmichheda:
When editTitle function is called, it sets the sequence value to "Sequence:", due to which I cannot retrieve the value.

$("#editTitle").click(function() {
    	// Using data attribute
    	var sequenceTitle = $("#title").data('value')  // gets value without using trim function
        $('<div class="col-4" id = "editSequenceTitle"><input id="newTitle" class="form-control" type="text"></div>').insertAfter( "#title" );
    	$("#title").val(sequenceTitle)
    	**$('#title').text('Sequence: ')**
        $("#newTitle").val(sequenceTitle)
        $("#editTitle").hide()
        $("#submitTitle").show()
        $("#cancelEditTitle").show()
        
    });

The line that is bold.
So, how can I added a data-attribute to set the value.
Is there any other way, I can retrieve it.

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
jdamerow: why is this here? what's the point of it? both #description and @newDescription are already in the dom tree.

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
jdamerow: same here

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
jdamerow: really?

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
```diff
@@ -1,23 +1,205 @@
...
+
+</script>
+
+
+<!-- title -->
+<div class="row align-items-center">
+	<h2 id="title" style="margin-bottom: 0%; margin-left: 1%;" data-value="${sequence.name}">
+		Sequence: ${sequence.name}
+		</h2>
+		<p id="titleName" style="margin-top: .5rem; font-size:35px; margin-bottom: .5rem;" data-value="${sequence.name}"></p>
```
jdamerow: why is this here?

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
```diff
@@ -1,23 +1,205 @@
...
+    $("#submitTitle").hide();
+    $("#cancelEditTitle").hide();
+    
+    $("#editTitle").click(function() {
+    	// Using data attribute
+    	var sequenceTitle = $("#titleName").data('value')  // gets value without using trim function
+        $('<div class="col-4" id = "editSequenceTitle"><input id="newTitle" class="form-control" type="text"></div>').insertAfter( "#title" );
+    	$("#title").val(sequenceTitle)
+    	$('#title').text('Sequence: ')
+        $("#newTitle").val(sequenceTitle)
```
jdamerow: why is there a title and a newtitle? what's that for? And since they are h2 and p, why do you use val() here?

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
ashmichheda: I used the p tag, since h2 tag cannot have val attribute. So in order to retrieve the Sequence name, I used p tag and got the value using data-attribute

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
ashmichheda: Since title is <h2> tag, while editing title, newTitle of the input form is added so that user can add any new title.

ashmichheda: Yes, I will remove the val()

jdamerow: I'm still confused why you don't use the h2 tag?

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
jdamerow: data attributes should be valid on any tag as far as I know

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
ashmichheda: Because, in the code,
When editTitle function is called, it sets the sequence value to "Sequence:", due to which I cannot retrieve the value from the h2 tag.

$("#editTitle").click(function() {
// Using data attribute
var sequenceTitle = $("#title").data('value') // gets value without using trim function
$('

').insertAfter( "#title" );
$("#title").val(sequenceTitle)
**$('#title').text('Sequence: ')**


Hence, I have added p tag to get the value back.

ashmichheda: I can only retrieve "Sequence :" as value, and not the actual value, e.g. **Sequence value: abcd**

jdamerow: I don't get it, is it a display issue? Why don't you do
```
$('#title').data('previous-title', sequenceTitle);
```
when start editing and then
```
var sequenceTitle = $("#title").data('previous-title')  // gets value
$("#title").text("Sequence: "+sequenceTitle)
```
?

ashmichheda: I can't use data attributes in title, since it's an H2 tag right?

ashmichheda: Hence, I am using a p tag with data attributes

jdamerow: https://www.w3schools.com/tags/att_data-.asp

ashmichheda: Okay, I will use data-attributes in h2 tag and remove the p tag.

jdamerow:
👍
---
Julia Damerow, PhD
Scientific Software Engineer
Global Biosocial Complexity Initiative
Digital Innovation Group
Arizona State University
email: jdamerow@asu.edu <>
web: http://diging.asu.edu, https://complexity.asu.edu/

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.jsp
```diff
@@ -1,23 +1,201 @@
...
+        
+   });
+    
+    // ---- Cancel Edited Description ----
+    
+    $("#cancelEditDescription").click(function(){
+        $("#submitDescription").hide()
+        $("#editDescription").show()
+        $("#cancelEditDescription").hide()
+        var description = $("#description").data('value')
```
jdamerow: where is that being used?

ashmichheda: Yes, its not being used. Will remove that.

jdamerow:
Whoohoo! ;)

ashmichheda:
:)

PR #86, opened 2019-10-03
Story/vspc 25

Neehaarika:


vspace/src/main/java/edu/asu/diging/vspace/config/PersistenceJPAConfig.java
```diff
@@ -85,7 +85,7 @@ Properties additionalProperties() {
         Properties properties = new Properties();
         properties.setProperty("hibernate.hbm2ddl.auto", "update");
         properties.setProperty("hibernate.dialect", "org.hibernate.dialect.MySQL57Dialect");
-        properties.setProperty("hibernate.show_sql", "false");
+        properties.setProperty("hibernate.show_sql", "true");
```
jdamerow: Let’s make this configurable through the Pom.

vspace/src/main/java/edu/asu/diging/vspace/core/factory/impl/ChoiceFactory.java
```diff
@@ -0,0 +1,36 @@
...
+     * 
+     * @see
+     * edu.asu.diging.vspace.core.factory.impl.ISlideFactory#createChoices(java.util.List<java.lang.String>)
+     */
+    @Override
+    public List<IChoice> createChoices(List<String> choiceIds) {
+        List<IChoice> choices = new ArrayList<IChoice>();
+        for(String choiceId: choiceIds) {
+            IChoice choice = new Choice();
+            choice.setSequence(sequenceManager.getSequence(choiceId));
```
jdamerow: Are choiceIds sequence ids? Or why do you call get sequence with a choice Id? Is so, very bad variable naming.

vspace/src/main/java/edu/asu/diging/vspace/core/factory/impl/SlideFactory.java
```diff
@@ -1,34 +1,61 @@
...
+    @Override
+    public IBranchingPoint createBranchingPoint(IModule module, SlideForm form) {
+        IBranchingPoint branchingPoint = new BranchingPoint();
+        branchingPoint.setName(form.getName());
+        branchingPoint.setDescription(form.getDescription());
+        branchingPoint.setModule(module);
+        branchingPoint.setContents(new ArrayList<IContentBlock>());
+       
+        List<IChoice> choices = choiceFactory.createChoices(form.getChoices());
+        branchingPoint.setChoices(choices);
```
jdamerow: These two lines are the only ones that are different from slides, right? I’d either create a method that you call that does the things that are the same or pass in what you want to create and then create different objects based on that.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ContentBlockManager.java
```diff
@@ -164,4 +178,20 @@ public ITextBlock getTextBlock(String textBlockId) {
...
     }
+    
+    /*
+     * (non-Javadoc)
+     * 
+     * @see edu.asu.diging.vspace.core.services.impl.IChoiceBlock#createTextBlock(java.
+     * lang.String, java.lang.String, java.lang.Integer)
+     */
+    @Override
+    public IChoiceBlock createChoiceBlock(String slideId, String choiceId, Integer contentOrder) {
```
jdamerow: same as before, is this really a choice id?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -5,34 +5,53 @@
...
+        slideRepo.save((Slide) slide);        
         return slide;
     }
 
+    @Override
+    public IBranchingPoint createBranchingPoint(IModule module, SlideForm slideForm) {
+        IBranchingPoint branchingPoint = slideFactory.createBranchingPoint(module, slideForm); 
+        for(IChoice choice: branchingPoint.getChoices()) {
+            choiceRepo.save((Choice)choice);
+        }
```
jdamerow: I forgot, have you tried if this works without explicitly saving the choices?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddChoiceBlockController.java
```diff
@@ -0,0 +1,33 @@
...
+    private IContentBlockManager contentBlockManager;
+
+    @RequestMapping(value = "/staff/module/{moduleId}/slide/{id}/choicecontent", method = RequestMethod.POST)
+    public ResponseEntity<ISequence> addChoiceBlock(@PathVariable("id") String slideId,
+            @PathVariable("moduleId") String moduleId, @RequestParam("content") String choiceContent,
+            @RequestParam("contentOrder") Integer contentOrder) throws IOException {
+
+        IChoiceBlock choiceBlock = contentBlockManager.createChoiceBlock(slideId, choiceContent, contentOrder);
+
+        return new ResponseEntity<>(choiceBlock.getSequence(), HttpStatus.OK);
```
jdamerow: thats odd. why are you returning the sequence here?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddSlideController.java
```diff
@@ -10,27 +10,39 @@
...
         return "staff/module/slide/add";
     }
 
     @RequestMapping(value = "/staff/module/{moduleId}/slide/add", method = RequestMethod.POST)
     public String addSlide(Model model, @PathVariable("moduleId") String moduleId, @ModelAttribute SlideForm slideForm,
             Principal principal) {
 
-        slideManager.createSlide(moduleId, slideForm);
+        IModule module = moduleManager.getModule(moduleId);
+        if(slideForm.getSlideType().substring(0, slideForm.getSlideType().length()-1).equals("slide")) {
```
jdamerow: why this? what does slide type contain?

jdamerow: also, do not hardcode strings!

jdamerow: And third, I would always use the most basic type in the else block. If someone submits "chicken" as type you want a slide to be created (if anything).

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SlideController.java
```diff
@@ -27,12 +29,13 @@
...
     @RequestMapping("/staff/module/{moduleId}/slide/{id}")
     public String listSlides(@PathVariable("id") String id, @PathVariable("moduleId") String moduleId, Model model) {
-
+        
         model.addAttribute("module", moduleManager.getModule(moduleId));
         model.addAttribute("slide", slideManager.getSlide(id));
         model.addAttribute("slideContents", contentBlockManager.getAllContentBlocks(id));
-
+        if(slideManager.getSlide(id) instanceof BranchingPoint) 
+            model.addAttribute("choices", ((IBranchingPoint)slideManager.getSlide(id)).getChoices());
```
jdamerow: please follow our style guide

vspace/src/main/webapp/WEB-INF/views/staff/modules/module.jsp
```diff
@@ -85,6 +85,9 @@ $(document).ready(function($) {
 						<div class="card" style="max-width: 18rem; margin-bottom:10px;">
 							<div align="left" class="card-body">
 								<a href="<c:url value="/staff/module/${module.id}/slide/${slide.id}" />">
+									<c:if test="${slide['class'].simpleName ==  'BranchingPoint'}">
```
jdamerow: indentation

vspace/src/main/java/edu/asu/diging/vspace/core/factory/impl/ChoiceBlockFactory.java
```diff
@@ -0,0 +1,23 @@
...
+
+import org.springframework.stereotype.Service;
+
+import edu.asu.diging.vspace.core.factory.IChoiceBlockFactory;
+import edu.asu.diging.vspace.core.model.IChoiceBlock;
+import edu.asu.diging.vspace.core.model.ISlide;
+import edu.asu.diging.vspace.core.model.impl.ChoiceBlock;
+
+@Service
+public class ChoiceBlockFactory implements IChoiceBlockFactory {
```
jdamerow: needs test class

vspace/src/main/java/edu/asu/diging/vspace/core/factory/impl/ChoiceFactory.java
```diff
@@ -0,0 +1,36 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+import edu.asu.diging.vspace.core.factory.IChoiceFactory;
+import edu.asu.diging.vspace.core.model.IChoice;
+import edu.asu.diging.vspace.core.model.impl.Choice;
+import edu.asu.diging.vspace.core.services.ISequenceManager;
+
+@Service
+public class ChoiceFactory implements IChoiceFactory{
```
jdamerow: test class

vspace/src/main/java/edu/asu/diging/vspace/core/factory/impl/SlideFactory.java
```diff
@@ -1,34 +1,61 @@
...
+    }
+    
+    /*
+     * (non-Javadoc)
+     * 
+     * @see
+     * edu.asu.diging.vspace.core.factory.impl.ISlideFactory#createBranchingPoint(edu.asu.diging.vspace.core.model.IModule, edu.asu.diging.vspace.web.staff.forms.SlideForm)
+     */
+    @Override
+    public IBranchingPoint createBranchingPoint(IModule module, SlideForm form) {
```
jdamerow: test case

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/ChoiceBlock.java
```diff
@@ -0,0 +1,34 @@
...
+import javax.persistence.OneToOne;
+
+import edu.asu.diging.vspace.core.model.IChoiceBlock;
+import edu.asu.diging.vspace.core.model.ISequence;
+
+@Entity
+public class ChoiceBlock extends ContentBlock implements IChoiceBlock {
+
+    @OneToOne(targetEntity = Sequence.class)
+    private ISequence sequence;
```
jdamerow: I'm confused, your choice blocks should contain choices, not sequences, shouldn't they?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
Neehaarika: it worked when used cascade=CascadeType.ALL. Which way should I go with?

jdamerow: since the choices without a branching point don't make sense, I'd cascade them all and avoid saving them individually.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddChoiceBlockController.java
Neehaarika: I will need both sequence id and name as well to display in the choiceBlock in front end. So, returning the object itself. Is that not the right way to do?

jdamerow: why don't your return the block?

Neehaarika:
addressed the comments and wrote test cases for mentioned classes.

vspace/src/main/java/edu/asu/diging/vspace/config/PersistenceJPAConfig.java
```diff
@@ -85,7 +85,6 @@ Properties additionalProperties() {
         Properties properties = new Properties();
         properties.setProperty("hibernate.hbm2ddl.auto", "update");
         properties.setProperty("hibernate.dialect", "org.hibernate.dialect.MySQL57Dialect");
-        properties.setProperty("hibernate.show_sql", "false");
```
jdamerow: let's keep that but make it configurable through the pom

vspace/src/main/java/edu/asu/diging/vspace/core/factory/impl/ChoiceFactory.java
```diff
@@ -0,0 +1,36 @@
...
+     * edu.asu.diging.vspace.core.factory.impl.ISlideFactory#createChoices(java.util.List<java.lang.String>)
+     */
+    @Override
+    public List<IChoice> createChoices(List<String> sequenceIds) {
+        List<IChoice> choices = new ArrayList<IChoice>();
+        for(String sequenceId: sequenceIds) {
+            IChoice choice = new Choice();
+            choice.setSequence(sequenceManager.getSequence(sequenceId));
+            choices.add(choice);
+           }
```
jdamerow: indentation

vspace/src/main/java/edu/asu/diging/vspace/core/factory/impl/SlideFactory.java
```diff
@@ -1,34 +1,49 @@
...
+     * edu.asu.diging.vspace.core.factory.impl.ISlideFactory#createSlide(edu.asu.diging.vspace.core.model.IModule, edu.asu.diging.vspace.web.staff.forms.SlideForm)
      */
     @Override
-    public ISlide createSlide(IModule module, SlideForm form) {
-        ISlide slide = new Slide();
+    public ISlide createSlide(String type, IModule module, SlideForm form) {
         
-        slide.setContents(new ArrayList<IContentBlock>());
+        ISlide slide;
+        if(type.equals("Slide")) {
```
jdamerow: do not hardcode strings. if there is no other way than doing it this way, use enums.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddSlideController.java
```diff
@@ -10,27 +10,39 @@
...
         return "staff/module/slide/add";
     }
 
     @RequestMapping(value = "/staff/module/{moduleId}/slide/add", method = RequestMethod.POST)
     public String addSlide(Model model, @PathVariable("moduleId") String moduleId, @ModelAttribute SlideForm slideForm,
             Principal principal) {
 
-        slideManager.createSlide(moduleId, slideForm);
+        IModule module = moduleManager.getModule(moduleId);
+        if(slideForm.getSlideType().equals("branchingPoint")) {
```
jdamerow: don't hard-code strings. use enums or if not possible otherwise, use constants.

vspace/src/main/webapp/WEB-INF/views/staff/modules/slides/add.jsp
```diff
@@ -10,13 +54,31 @@
...
+	     </form:select>
+ </div>
+  
+ <div class="form-group row" id="sequences" style="display:none;">
+    <label for="choices" class="col-md-2 col-form-label">Add Choices:</label>	
+            <select multiple="multiple" id="selectSequence" name="selectSequence" class="form-control form-control-sm" style="width: 68px;">
+				 <c:forEach items="${sequences}" var="sequence">
+					<option value='${sequence.id}' id="${sequence.id}">${sequence.name}</option>
+				 </c:forEach>
+			</select>
```
jdamerow: indentation

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ChoiceBlockFactoryTest.java
```diff
@@ -0,0 +1,31 @@
...
+import org.junit.Test;
+
+import edu.asu.diging.vspace.core.factory.impl.ChoiceBlockFactory;
+import edu.asu.diging.vspace.core.model.IChoiceBlock;
+import edu.asu.diging.vspace.core.model.impl.Choice;
+import edu.asu.diging.vspace.core.model.impl.Slide;
+
+public class ChoiceBlockFactoryTest {
+
+private ChoiceBlockFactory choiceBlockFactoryToTest = new ChoiceBlockFactory();
```
jdamerow: indentation

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/ChoiceBlock.java
```diff
@@ -0,0 +1,34 @@
...
+import javax.persistence.OneToOne;
+
+import edu.asu.diging.vspace.core.model.IChoice;
+import edu.asu.diging.vspace.core.model.IChoiceBlock;
+
+@Entity
+public class ChoiceBlock extends ContentBlock implements IChoiceBlock {
+
+    @OneToOne(targetEntity = Choice.class)
+    private IChoice choice;
```
jdamerow: what about the other choices? for now a choice block should simply show all choices of a branching point.

Neehaarika: No, we are displaying all the choices on the top and giving an option "Add Choice" to be able to add the choices(as each individual choice blocks) where ever we want them to be added. Isn't it?

jdamerow: hm, we definitely want an option to add all at once (since in most cases users probably want to add all in a block anyways and layouting that will be easier), but even if we won't do that now, keep that in mind. It probably makes most sense to let choiceblocks have a list of choices so they can be reused for that scenario, even if they right now only contain one choice. makes sense?

Neehaarika: Okay got it.

Neehaarika:
These changes are almost addressed but need minor fixes before we can close this task.

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/Choice.java
```diff
@@ -21,6 +30,15 @@
         strategy = "edu.asu.diging.vspace.core.data.IdGenerator")
     private String id;
 
+//    @OneToMany(targetEntity = ChoiceBlock.class)
+//    @JoinTable(name = "ChoiceBlck_Choice",
+//    joinColumns = {@JoinColumn(name = "ChoiceBlock_id")},
+//    inverseJoinColumns = {@JoinColumn(name = "choices_id")}
+//    )
```
jdamerow, line 37: remove if not needed

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/Choice.java
```diff
@@ -65,4 +83,12 @@ public ISequence getSequence() {
...
         this.sequence = sequence;
     }
+    
+//    public IChoiceBlock getChoiceBlock() {
+//        return choiceBlock;
+//    }
+//
+//    public void setChoiceBlock(IChoiceBlock choiceBlock) {
+//        this.choiceBlock = choiceBlock;
+//    }
```
jdamerow, line 93: why are getter/setter not needed?

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/Slide.java
```diff
@@ -23,6 +27,9 @@
 import edu.asu.diging.vspace.core.model.ISlide;
 
 @Entity
+@Inheritance(strategy = InheritanceType.JOINED)
```
jdamerow, line 30: does this work for existing slides?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -35,16 +40,29 @@
...
+    public ISlide createSlide(IModule module, SlideForm slideForm, SlideType type) {
+        ISlide slide = slideFactory.createSlide(module, slideForm, type);             
+        slideRepo.save((Slide) slide);        
+        return slide;
+    }
+
+    @Override
+    public IBranchingPoint createBranchingPoint(IModule module, SlideForm slideForm, SlideType type) {
+        ISlide branchingPoint = slideFactory.createSlide(module, slideForm, type);            
+        bpointRepo.save((BranchingPoint) branchingPoint);
```
jdamerow, line 62: you want to return the branching point that the save method returns (it'll have the id)

vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddChoiceBlockController.java
```diff
@@ -0,0 +1,32 @@
...
+
+@Controller
+public class AddChoiceBlockController {
+
+    @Autowired
+    private IContentBlockManager contentBlockManager;
+
+    @RequestMapping(value = "/staff/module/{moduleId}/slide/{id}/choicecontent", method = RequestMethod.POST)
+    public ResponseEntity<IChoiceBlock> addChoiceBlock(@PathVariable("id") String slideId,
+            @PathVariable("moduleId") String moduleId, @RequestParam("content") String choiceId,
```
jdamerow, line 25: is it content or a choice id?

jdamerow: Also, I think we said that the choice block would for now have all choices of the branching point.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SlideController.java
jdamerow: put `slideManager.getSlide(id)` into a varialbe. there is no reason to call the db twice (even though it might be cached)

vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddChoiceBlockController.java
prjadhav14: Currently, we are popping the screen showing all the available choices through which the user can select the one that should be added as content. 
should this option be removed and defaulted to link all the choices of branching point or user should be able to select the choices just that user should select all at a time from the branching points list?

jdamerow: jsut default to all choices; we'll add features to only add specific ones later.

vspace/src/main/java/edu/asu/diging/vspace/core/factory/impl/ChoiceBlockFactory.java
```diff
@@ -0,0 +1,33 @@
...
+    @Override
+    public IChoiceBlock createChoiceBlock(ISlide slide, Integer contentOrder, List<IChoice> choices) {
+        IChoiceBlock choiceBlock = new ChoiceBlock();
+        List<IChoice> choicesList= new ArrayList<IChoice>();
+        for(int i=0;i<choices.size();i++) {
+            choicesList.add(choices.get(i));
+        }
+        choiceBlock.setSlide(slide);
+        choiceBlock.setContentOrder(contentOrder);
+        choiceBlock.setChoices(choicesList);
```
jdamerow, line 30: I wouldn't set the choices here. since it should show all choices of a branching point, the block would have to be updated when new choices are added, existing one deleted, etc.

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/Slide.java
jdamerow: I'm fairly certain you don't use discriminator annotations with joined strategy. you don't need them, there's no need for a column that discriminates between classes. Also still: does this work with existing data?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -35,16 +40,29 @@
...
-        ISlide slide = slideFactory.createSlide(module, slideForm);
-        slideRepo.save((Slide) slide);
-        return slide;
-    }
-
+    @Override
+    public ISlide createSlide(IModule module, SlideForm slideForm, SlideType type) {
+        ISlide slide = slideFactory.createSlide(module, slideForm, type);             
+        slideRepo.save((Slide) slide);        
+        return slide;
```
jdamerow, line 56: return what "save" returns (or you might not have an id)

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -35,16 +40,29 @@
...
+    }
+
+    @Override
+    public IBranchingPoint createBranchingPoint(IModule module, SlideForm slideForm, SlideType type) {
+        ISlide branchingPoint = slideFactory.createSlide(module, slideForm, type);            
+        branchingPoint = bpointRepo.save((BranchingPoint) branchingPoint);        
+        return (IBranchingPoint) branchingPoint;
```
jdamerow, line 63: see above

vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddChoiceBlockController.java
```diff
@@ -0,0 +1,41 @@
...
+@Controller
+public class AddChoiceBlockController {
+
+    @Autowired
+    private IContentBlockManager contentBlockManager;
+    
+    @Autowired
+    private ISlideManager slideManager;
+
+    @RequestMapping(value = "/staff/module/{moduleId}/slide/{id}/choicecontent", method = RequestMethod.POST)
```
jdamerow, line 30: let's put a slash between `choice` and `content`

vspace/src/main/java/edu/asu/diging/vspace/core/model/display/SlideType.java
```diff
@@ -0,0 +1,6 @@
+package edu.asu.diging.vspace.core.model.display;
+
+public enum SlideType {
+
+    SLIDE, BRANCHINGPOINT;
```
jdamerow, line 5: add a `_` between branching and point

vspace/src/main/java/edu/asu/diging/vspace/core/factory/impl/ChoiceBlockFactory.java
prjadhav14: So, add choices option on staff slide screen should allow user to add/remove choices to slide?

jdamerow: the user should be able to add choice blocks. And a choice block would for now always show all choices. The user could add two or more choice blocks (e.g. beginning and end of slide) and they would just show the same choices.

jdamerow: but there need to be choice blocks, as in a next step we would allow the user to select a layout and the layout would specify where what blocks go.

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/Choice.java
```diff
@@ -12,6 +12,7 @@
 import edu.asu.diging.vspace.core.model.ISequence;
 
 @Entity
+//@Inheritance(strategy = InheritanceType.JOINED)
```
jdamerow, line 15: remove

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/Slide.java
jdamerow: what happens if you remove this? why do you need this? the tables are already setup to use the single table strategy, why this change?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ContentBlockManager.java
```diff
@@ -207,4 +218,27 @@ public ITextBlock getTextBlock(String textBlockId) {
...
+     * (non-Javadoc)
+     * 
+     * @see edu.asu.diging.vspace.core.services.impl.IChoiceBlock#createTextBlock(java.
+     * lang.String, java.lang.String, java.lang.Integer)
+     */
+    @Override
+    public IChoiceBlock createChoiceBlock(String slideId, List<IChoice> choices, Integer contentOrder) {
+        IChoiceBlock choiceBlock = choiceBlockFactory.createChoiceBlock(slideManager.getSlide(slideId), contentOrder, choices); 
+        choiceBlock = choiceBlockRepo.save((ChoiceBlock)choiceBlock);
+        return choiceBlock;
```
jdamerow, line 241: merge with line 240

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
jdamerow: merge with line 55

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
jdamerow: merge with line 62

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -34,17 +39,36 @@
...
     }
 
+    @Override
+    public IBranchingPoint updateBranchingPoint(IBranchingPoint branchingPoint) {
+        branchingPoint = bpointRepo.save((BranchingPoint) branchingPoint);        
+        return (IBranchingPoint) branchingPoint;
```
jdamerow, line 69: merge with line 68 or don't return anything (when updating you already have an id)

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/ChoiceBlock.java
```diff
@@ -0,0 +1,48 @@
...
+import edu.asu.diging.vspace.core.model.IChoice;
+import edu.asu.diging.vspace.core.model.IChoiceBlock;
+
+@Entity
+public class ChoiceBlock extends ContentBlock implements IChoiceBlock {
+
+    @ManyToMany(targetEntity = Choice.class)
+    private List<IChoice> choices;
+
+    private boolean showsAll;
```
jdamerow, line 17: where is the logic that uses this property?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ContentBlockManager.java
```diff
@@ -207,4 +218,27 @@ public ITextBlock getTextBlock(String textBlockId) {
...
+    /*
+     * (non-Javadoc)
+     * 
+     * @see edu.asu.diging.vspace.core.services.impl.IChoiceBlock#createTextBlock(java.
+     * lang.String, java.lang.String, java.lang.Integer)
+     */
+    @Override
+    public IChoiceBlock createChoiceBlock(String slideId, List<IChoice> choices, Integer contentOrder, boolean showsAll) {
+        IChoiceBlock choiceBlock = choiceBlockFactory.createChoiceBlock(slideManager.getSlide(slideId), contentOrder, choices); 
+        choiceBlock.setShowsAll(showsAll);
```
jdamerow, line 240: if "showsAll" is selected, you don't need to store the choices with the block

vspace/src/main/java/edu/asu/diging/vspace/core/factory/impl/ChoiceBlockFactory.java
```diff
@@ -0,0 +1,31 @@
...
+    
+    /* (non-Javadoc)
+     * @see edu.asu.diging.vspace.core.factory.impl.IChoiceBlockFactory#createChoiceBlock(edu.asu.diging.vspace.core.model.impl.ISlide, java.lang.Integer,
+     * edu.asu.diging.vspace.core.model.impl.IChoice)
+     */    
+    @Override
+    public IChoiceBlock createChoiceBlock(ISlide slide, Integer contentOrder, List<IChoice> choices) {
+        IChoiceBlock choiceBlock = new ChoiceBlock();
+        List<IChoice> choicesList= new ArrayList<IChoice>();
+        choicesList.addAll(choices);
```
jdamerow, line 25: i'm pretty sure you don't need the new list anymore, do you?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddSlideController.java
```diff
@@ -10,27 +11,51 @@
...
             Principal principal) {
 
-        slideManager.createSlide(moduleId, slideForm);
+        IModule module = moduleManager.getModule(moduleId);
+        SlideType type = slideForm.getType().isEmpty() ? null : SlideType.valueOf(slideForm.getType());
+        if(type.equals(SlideType.BRANCHING_POINT)) {
+            IBranchingPoint branchingPoint = slideManager.createBranchingPoint(module, slideForm, type);
+            List<IChoice> choices = branchingPoint.getChoices();
+            int contentOrder=branchingPoint.getContents().size()+1;
+            IChoiceBlock choiceBlock = contentBlockManager.createChoiceBlock(branchingPoint.getId(), choices, contentOrder, false);
```
jdamerow, line 55: the default should be "true" to show all choices even when they are updated

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditSlideController.java
```diff
@@ -36,4 +54,47 @@
...
+    @RequestMapping(value="/staff/module/{moduleId}/slide/{slideId}/edit", method=RequestMethod.POST)
+    public String save(@ModelAttribute SlideForm slideForm, @PathVariable("moduleId") String moduleId, @PathVariable("slideId") String slideId) {
+        ISlide slide = slideManager.getSlide(slideId);
+        slide.setName(slideForm.getName());
+        slide.setDescription(slideForm.getDescription());
+        SlideType type = slideForm.getType().isEmpty() ? null : SlideType.valueOf(slideForm.getType());
+        if(type.equals(SlideType.BRANCHING_POINT)) {
+            List<IChoice> choices = choiceFactory.createChoices(slideForm.getChoices());
+            IBranchingPoint branchingPoint = (IBranchingPoint) slide;
+            branchingPoint.setChoices(choices);
```
jdamerow, line 93: what about the previously created choices?

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ChoiceFactoryTest.java
```diff
@@ -0,0 +1,58 @@
...
+        newSequences.add(newSequence2);
+      
+        Mockito.when(mockSequenceManager.getSequence("sequence1")).thenReturn(newSequence1);
+        Mockito.when(mockSequenceManager.getSequence("sequence2")).thenReturn(newSequence2);
+
+        List<IChoice> actualChoices = choiceFactoryToTest.createChoices(sequenceIds);
+        int i = 0;
+        for(IChoice actualChoice : actualChoices) {
+            Assert.assertEquals(newSequences.get(i), actualChoice.getSequence());
+            Assert.assertEquals(newSequences.get(i).getId(), actualChoice.getSequence().getId());
```
jdamerow, line 54: if the sequence objects are the same, their ids are the same

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditSlideController.java
prjadhav14: we are providing an option on UI for a user to select and deselect the sequences as choices while editing the slide. When the user changes the list and clicks on save, deselected items will be deleted from Branching choice options and selected ones will be added to the list.

jdamerow: where are the previous ones being deleted? you create new choices, set them in the slide, but what happens with the existing choices? Since you don't have orphanremoval set, I don't think hibernate does that for you?

prjadhav14: got it, It is just deleting and adding the links in the join table by creating new choice ids leaving the existing ones hanging in the Choices List.  I will fix this.

prjadhav14:
Closing the pull request to fix linking and delinking of choices.

vspace/src/main/java/edu/asu/diging/vspace/core/factory/impl/ChoiceBlockFactory.java
```diff
@@ -0,0 +1,34 @@
...
+     */    
+    @Override
+    public IChoiceBlock createChoiceBlock(ISlide slide, Integer contentOrder, List<IChoice> choices, boolean showsAll) {
+        IChoiceBlock choiceBlock = new ChoiceBlock();
+        choiceBlock.setSlide(slide);
+        choiceBlock.setContentOrder(contentOrder);
+        if(showsAll) {
+            choiceBlock.setShowsAll(showsAll);
+        }
+        else {
```
jdamerow, line 28: code style

vspace/src/main/java/edu/asu/diging/vspace/core/factory/impl/ChoiceBlockFactory.java
```diff
@@ -0,0 +1,34 @@
...
+            choiceBlock.setChoices(choices);
+            choiceBlock.setShowsAll(showsAll);
```
jdamerow, line 30: just set showsAll outside the if and only test if choices need to be set

vspace/src/main/java/edu/asu/diging/vspace/core/factory/impl/ChoiceFactory.java
```diff
@@ -0,0 +1,36 @@
...
+     */
+    @Override
+    public List<IChoice> createChoices(List<String> sequenceIds) {
+        List<IChoice> choices = new ArrayList<IChoice>();
+        for(String sequenceId: sequenceIds) {
+            IChoice choice = new Choice();
+            choice.setSequence(sequenceManager.getSequence(sequenceId));
+            choices.add(choice);
+        }
+        return choices;
```
jdamerow, line 34: this can probably be a one liner using `stream`, `map`, and `collect`

vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddChoiceBlockController.java
```diff
@@ -0,0 +1,46 @@
...
+    @Autowired
+    private IContentBlockManager contentBlockManager;
+
+    @Autowired
+    private ChoiceRepository choiceRepo;
+
+    @RequestMapping(value = "/staff/module/{moduleId}/slide/{id}/choice/content", method = RequestMethod.POST)
+    public ResponseEntity<IChoiceBlock> addChoiceBlock(@PathVariable("id") String slideId,
+            @PathVariable("moduleId") String moduleId,
+            @RequestParam("contentOrder") Integer contentOrder, @RequestParam("selectedChoices") List<String> selectedChoice, @RequestParam("showsAll") boolean showsAll) throws IOException {
```
jdamerow, line 35: variable name should be plural

vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddChoiceBlockController.java
```diff
@@ -0,0 +1,46 @@
...
+        List<IChoice> choices = new ArrayList<IChoice>();
+        if(!showsAll) {
+            for(String choiceID : selectedChoice) {
+                Optional<Choice> choice = choiceRepo.findById(choiceID);
+                choices.add(choice.get());
+            }
```
jdamerow, line 41: this should be done in manager (business logic)

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditSlideController.java
```diff
@@ -36,4 +55,57 @@
...
+        SlideForm slideForm = new SlideForm();
+        slideForm.setName(slide.getName());
+        slideForm.setDescription(slide.getDescription());
+        if(slide instanceof BranchingPoint) {
+            slideForm.setType(SlideType.BRANCHING_POINT.toString());
+            IBranchingPoint branchingPoint = (IBranchingPoint) slide;
+            List<String> choices = new ArrayList<String>();
+            for(IChoice choice : branchingPoint.getChoices()) {
+                choices.add(choice.getSequence().getId());
+            }
```
jdamerow, line 71: should be a one liner (stream, map, collect)

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditSlideController.java
```diff
@@ -36,4 +55,57 @@
...
+    }
+
+    @RequestMapping(value="/staff/module/{moduleId}/slide/{slideId}/edit", method=RequestMethod.POST)
+    public String save(@ModelAttribute SlideForm slideForm, @PathVariable("moduleId") String moduleId, @PathVariable("slideId") String slideId) {
+        ISlide slide = slideManager.getSlide(slideId);
+        slide.setName(slideForm.getName());
+        slide.setDescription(slideForm.getDescription());
+        SlideType type = slideForm.getType().isEmpty() ? null : SlideType.valueOf(slideForm.getType());
+        if(type.equals(SlideType.BRANCHING_POINT)) {
+            List<ISequence> choiceSequences = new ArrayList<ISequence>();
```
jdamerow, line 92: what is this list for?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditSlideController.java
```diff
@@ -36,4 +55,57 @@
...
+        if(type.equals(SlideType.BRANCHING_POINT)) {
+            List<ISequence> choiceSequences = new ArrayList<ISequence>(); 
+            List<IChoice> existingChoices=((IBranchingPoint)slide).getChoices();
+            for(IChoice choice : existingChoices) {
+                choiceSequences.add(choice.getSequence());
+            }
+            List<IChoice> choices = choiceFactory.createChoices(slideForm.getChoices());
+            for(IChoice choice : choices) {
+                if(!choiceSequences.contains(choice.getSequence())) {
+                    existingChoices.add(choice);
```
jdamerow, line 100: what about choices that have been removed?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditSlideController.java
```diff
@@ -36,4 +55,57 @@
...
+                }
+            }
+            IBranchingPoint branchingPoint = (IBranchingPoint) slide;
+            branchingPoint.setChoices(existingChoices);
```
jdamerow, line 104: most of the above code should be done in the manager

vspace/src/main/java/edu/asu/diging/vspace/core/factory/impl/ChoiceBlockFactory.java
```diff
@@ -0,0 +1,33 @@
...
+     * @see edu.asu.diging.vspace.core.factory.impl.IChoiceBlockFactory#createChoiceBlock(edu.asu.diging.vspace.core.model.impl.ISlide, java.lang.Integer,
+     * edu.asu.diging.vspace.core.model.impl.IChoice)
+     */    
+    @Override
+    public IChoiceBlock createChoiceBlock(ISlide slide, Integer contentOrder, List<IChoice> choices, boolean showsAll) {
+        IChoiceBlock choiceBlock = new ChoiceBlock();
+        choiceBlock.setSlide(slide);
+        choiceBlock.setContentOrder(contentOrder);
+        choiceBlock.setShowsAll(showsAll);
+        List<IChoice> choicesList = new ArrayList<IChoice>(choices);
```
jdamerow, line 27: merge with line 29

vspace/src/main/java/edu/asu/diging/vspace/core/factory/impl/ChoiceFactory.java
jdamerow: merge with line 30

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -34,17 +42,63 @@
...
+        return bpointRepo.save((BranchingPoint) branchingPoint);
     }
 
+    @Override
+    public void updateBranchingPoint(IBranchingPoint branchingPoint, List<String> editedChoices) {
+        List<String> existingChoiceSequences = new ArrayList<String>(); 
+        List<IChoice> existingChoices=branchingPoint.getChoices();
+        for(IChoice choice : existingChoices) {
+            existingChoiceSequences.add(choice.getSequence().getId());
+        }
```
jdamerow, line 75: can be a one-liner with stream, map and collect

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -34,17 +42,63 @@
...
+        List<String> deletedChoices = (List<String>) CollectionUtils.subtract(existingChoiceSequences, editedChoices);
+        List<String> addedChoices = (List<String>) CollectionUtils.subtract(editedChoices,existingChoiceSequences);
+
+        if(addedChoices.size()>0) {
```
jdamerow, line 79: I don't think you need this if

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -34,17 +42,63 @@
...
+            List<IChoice> choices = choiceFactory.createChoices(addedChoices);
+            for(IChoice choice : choices) {
+                existingChoices.add(choice);
+            }
+        }
+        List<IChoice> choicestoDelete = new ArrayList<IChoice>();
+        if(deletedChoices.size()>0) {
```
jdamerow, line 86: or this if

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -34,17 +42,63 @@
...
+            choicestoDelete = existingChoices.stream().filter(choice -> deletedChoices.contains(choice.getSequence().getId())).collect(Collectors.toList());
+            for(String choice : deletedChoices) {
+                existingChoices.removeIf(choices -> choices.getSequence().getId().equals(choice));
```
jdamerow, line 89: bad variable name `choices`. It's just one choice at a time.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -34,17 +42,63 @@
...
+        List<String> deletedChoices = (List<String>) CollectionUtils.subtract(existingChoiceSequences, editedChoices);
+        List<String> addedChoices = (List<String>) CollectionUtils.subtract(editedChoices,existingChoiceSequences);
+
+        if(addedChoices.size()>0) {
+            List<IChoice> choices = choiceFactory.createChoices(addedChoices);
+            for(IChoice choice : choices) {
+                existingChoices.add(choice);
+            }
+        }
+        List<IChoice> choicestoDelete = new ArrayList<IChoice>();
```
jdamerow, line 85: needs proper camelcase

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -34,17 +42,63 @@
...
+        if(deletedChoices.size()>0) {
+            choicestoDelete = existingChoices.stream().filter(choice -> deletedChoices.contains(choice.getSequence().getId())).collect(Collectors.toList());
+            for(String choice : deletedChoices) {
+                existingChoices.removeIf(choices -> choices.getSequence().getId().equals(choice));
+            }
+        }
+        branchingPoint.setChoices(existingChoices);
+        bpointRepo.save((BranchingPoint) branchingPoint);
+
+        if(choicestoDelete.size()>0) {
```
jdamerow, line 95: again, I don't think you need this

vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddChoiceBlockController.java
jdamerow: still true

vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddSlideController.java
```diff
@@ -19,19 +27,36 @@
...
 
-        slideManager.createSlide(moduleId, slideForm);
-        
+        IModule module = moduleManager.getModule(moduleId);
+        SlideType type = slideForm.getType().isEmpty() ? null : SlideType.valueOf(slideForm.getType());
+        if(type.equals(SlideType.BRANCHING_POINT)) {
+            IBranchingPoint branchingPoint = slideManager.createBranchingPoint(module, slideForm, type);
+            List<IChoice> choices = branchingPoint.getChoices();
+            int contentOrder=branchingPoint.getContents().size()+1;
+            IChoiceBlock choiceBlock = contentBlockManager.createChoiceBlock(branchingPoint.getId(), choices, contentOrder, true);
```
jdamerow, line 55: same here, I woudl do all of this in the manager (there should be just one line here)

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ChoiceFactoryTest.java
jdamerow: ?

jdamerow: which means, remove line 53

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ContentBlockManager.java
```diff
@@ -207,4 +219,35 @@ public ITextBlock getTextBlock(String textBlockId) {
...
+    
+    /*
+     * (non-Javadoc)
+     * 
+     * @see edu.asu.diging.vspace.core.services.impl.IChoiceBlock#createTextBlock(java.
+     * lang.String, java.lang.String, java.lang.Integer)
+     */
+    @Override
+    public IChoiceBlock createChoiceBlock(IBranchingPoint branchingPoint, List<String> selectedChoices, boolean showsAll) {
+        //List<IChoice> choices = branchingPoint.getChoices();
```
jdamerow, line 240: remove if not needed

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ContentBlockManager.java
```diff
@@ -207,4 +219,35 @@ public ITextBlock getTextBlock(String textBlockId) {
...
+        List<IChoice> choices = new ArrayList<IChoice>();
+        if(!showsAll) {
+            for(String choiceID : selectedChoices) {
+                IChoice choice = slideManager.getChoice(choiceID);
+                choices.add(choice);
+            }
```
jdamerow, line 246: once again, this can be a one-liner using the stream api ;)

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -34,17 +42,52 @@
...
+        List<IChoice> existingChoices=branchingPoint.getChoices();
+        List<String> existingChoiceSequences=existingChoices.stream().map(choiceSequence -> choiceSequence.getSequence().getId()).collect(Collectors.toList());
+        List<String> deletedChoices = (List<String>) CollectionUtils.subtract(existingChoiceSequences, editedChoices);
+        List<String> addedChoices = (List<String>) CollectionUtils.subtract(editedChoices,existingChoiceSequences);
+        List<IChoice> newlyAddedChoices = choiceFactory.createChoices(addedChoices);
+        for(IChoice addedChoice : newlyAddedChoices) {
+            existingChoices.add(addedChoice);
+        }
+        List<IChoice> choicesToDelete = new ArrayList<IChoice>();
+        choicesToDelete = existingChoices.stream().filter(choice -> deletedChoices.contains(choice.getSequence().getId())).collect(Collectors.toList());
```
jdamerow, line 80: merge with line 79

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -34,17 +42,52 @@
...
+        for(String deletedChoice : deletedChoices) {
+            existingChoices.removeIf(choice -> choice.getSequence().getId().equals(deletedChoice));
```
jdamerow, line 82: can't you do:
```
removeIf(choice -> deletedChoices.contains(choice.getSequence.getId()))
```
and get rid of the loop

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -34,17 +42,52 @@
...
+        return bpointRepo.save((BranchingPoint) branchingPoint);
     }
 
+    @Override
+    public void updateBranchingPoint(IBranchingPoint branchingPoint, List<String> editedChoices) {
+        List<IChoice> existingChoices=branchingPoint.getChoices();
+        List<String> existingChoiceSequences=existingChoices.stream().map(choiceSequence -> choiceSequence.getSequence().getId()).collect(Collectors.toList());
+        List<String> deletedChoices = (List<String>) CollectionUtils.subtract(existingChoiceSequences, editedChoices);
+        List<String> addedChoices = (List<String>) CollectionUtils.subtract(editedChoices,existingChoiceSequences);
+        List<IChoice> newlyAddedChoices = choiceFactory.createChoices(addedChoices);
```
jdamerow, line 75: it would be a lot easier to read if you call lists that contain ids "...Ids" and lists that contain choices "...Choices"

jdamerow:
Jenkins, test this please.

jdamerow:
for the frontend:
- branching point choices have to be at least deletable if not editable (we can put editing into a new ticket; if deleting it will be a lot of work we can put that in a new ticket as well, but I'd prefer it to be done here (otherwise, once a choice block has been created, errors can't be corrected anymore))
- after adding a new choice block, the new block is empty:
image
- in the block drop down menu it should say "Choices (Links to other sequences)" not "Choice"

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -34,17 +42,49 @@
...
 
+    @Override
+    public void updateBranchingPoint(IBranchingPoint branchingPoint, List<String> editedChoiceSequenceIds) {
+        List<IChoice> existingChoices=branchingPoint.getChoices();
+        List<String> existingChoiceSequenceIds=existingChoices.stream().map(choiceSequence -> choiceSequence.getSequence().getId()).collect(Collectors.toList());
+        List<String> deletedChoiceSequenceIds = (List<String>) CollectionUtils.subtract(existingChoiceSequenceIds, editedChoiceSequenceIds);
+        List<String> addedChoiceSequenceIds = (List<String>) CollectionUtils.subtract(editedChoiceSequenceIds,existingChoiceSequenceIds);
+        List<IChoice> newlyAddedChoices = choiceFactory.createChoices(addedChoiceSequenceIds);
+        for(IChoice addedChoice : newlyAddedChoices) {
+            existingChoices.add(addedChoice);
```
jdamerow, line 77: can you use `addAll` instead of this loop?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -34,17 +42,49 @@
...
+        }
+        List<IChoice> choicesToDelete = existingChoices.stream().filter(choice -> deletedChoiceSequenceIds.contains(choice.getSequence().getId())).collect(Collectors.toList());
+        existingChoices.removeIf(choice -> deletedChoiceSequenceIds.contains(choice.getSequence().getId()));
+        branchingPoint.setChoices(existingChoices);
+        bpointRepo.save((BranchingPoint) branchingPoint);
+        for(IChoice choice:choicesToDelete) {
+            choiceRepo.deleteById(choice.getId());
+        }
```
jdamerow, line 85: have you tried using `deleteAll` instead of the loop?

vspace/src/main/webapp/WEB-INF/views/staff/modules/slides/edit.jsp
```diff
@@ -0,0 +1,96 @@
...
+
+<script>
+//# sourceURL=click.js
+$(window).on("load", function() {
+    var vals = [];
+    var choices = [];
+    var selectedchoices = ${choices};
+    var selectedType =  $('select[id="type"]');
+    console.log(selectedType.val());
+    console.log(selectedchoices);
```
jdamerow, line 15: remove console output

vspace/src/main/webapp/WEB-INF/views/staff/modules/slides/edit.jsp
```diff
@@ -0,0 +1,96 @@
...
+    if (selectedType.val() == "BRANCHING_POINT"){
+            $('#sequences').show();
+            $('#selectSequence').multiSelect({                  
+                dblClick : true,
+                afterInit : function(container) {
+                    $("#selectSequence").find("option").each(function() {
+                        console.log($(this).val());
```
jdamerow, line 22: remove

vspace/src/main/webapp/WEB-INF/views/staff/modules/slides/edit.jsp
```diff
@@ -0,0 +1,96 @@
...
+                        if (String(choices[i]) === String(value)) {
+                            choices.splice(i, 1);
+                            break;
+                        }
+                    }
+                    $("#choices").val(choices);
+                }
+            });
+            $(selectedchoices).each(function() {
+                console.log($(this).val());
```
jdamerow, line 47: remove

jdamerow:
Jenkins, test this please.

jdamerow:
two finally things:
- the add choice modal, doesn't reset. I see the previous selection when I add another choice block.
- the "All" and only selected choices options need to be separate. If I want to show all current choices, but not update them if I add new ones, I can't really do that. In other words, there should be an "Always show all choices" option and then a second option that let's me choose which ones.
image

jdamerow:
Jenkins, test this please.

jdamerow:
Jenkins, test this please.

jdamerow:
one last thing and then it should be good: if the user doesn't select anything, they shouldn't be able to create a new choice block. and by default "Always show all choices" should be selected.

vspace/src/main/webapp/WEB-INF/views/staff/modules/slides/slide.jsp
```diff
@@ -475,7 +475,18 @@ $(document).ready(function() {
         });
         $("#textBlockText").val('')
     });
-   	
+    
+    $('input:checkbox').click(function() {
+        if ($(this).is(':checked')) {
+            console.log("In If block");
```
jdamerow, line 481: remove

vspace/src/main/webapp/WEB-INF/views/staff/modules/slides/slide.jsp
```diff
@@ -475,7 +475,18 @@ $(document).ready(function() {
...
+        	$('#submitChoices').prop("disabled", false);
+        } else {
+            console.log("In else block");
```
jdamerow, line 484: remove

PR #88, opened 2019-11-08
Story/vspc 54

Neehaarika:


vspace/src/main/java/edu/asu/diging/vspace/web/staff/DeleteSlideController.java
```diff
@@ -0,0 +1,27 @@
...
+import edu.asu.diging.vspace.core.services.ISlideManager;
+
+@Controller
+public class DeleteSlideController {
+    
+    @Autowired
+    private ISlideManager slideManager;
+
+    @RequestMapping(value = "/staff/module/{id}/slide/remove/{slideId}", method = RequestMethod.DELETE)
+    public ResponseEntity<?> deleteModuleLink(@PathVariable("id") String moduleId,
```
jdamerow: why is this method called "deleteModuleLink"?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DeleteSlideController.java
```diff
@@ -0,0 +1,27 @@
...
+import edu.asu.diging.vspace.core.exception.SlideDoesNotExistException;
+import edu.asu.diging.vspace.core.services.ISlideManager;
+
+@Controller
+public class DeleteSlideController {
+    
+    @Autowired
+    private ISlideManager slideManager;
+
+    @RequestMapping(value = "/staff/module/{id}/slide/remove/{slideId}", method = RequestMethod.DELETE)
```
jdamerow: since you're making a DELETE request, the url does not need to contain "remove"

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DeleteSlideController.java
```diff
@@ -0,0 +1,27 @@
...
+    public ResponseEntity<?> deleteModuleLink(@PathVariable("id") String moduleId,
+            @PathVariable("slideId") String slideId) throws SlideDoesNotExistException {
```
jdamerow: the exception needs to be handled here.

jdamerow: also, where is that actually thrown?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DeleteSlideController.java
```diff
@@ -0,0 +1,27 @@
...
+        
+        slideManager.deleteSlide(slideId);
```
jdamerow: the user should see a warning if the slide is used in a sequence

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -46,4 +54,27 @@ public ISlide getSlide(String slideId) {
     public void updateSlide(Slide slide) {
         slideRepo.save((Slide) slide);
     }
+    
+    @Override
+    public void deleteSlideById(String slideId) throws SlideDoesNotExistException {
+        try {
+            slideRepo.delete((Slide) getSlide(slideId));
+        } catch (IllegalArgumentException | EmptyResultDataAccessException exception) {
```
jdamerow: I don't think the delete method throws an emptyresultdataaccessexception. Are you sure about that? the javadoc says otherwise I believe. Why do you catch that here?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -46,4 +54,27 @@ public ISlide getSlide(String slideId) {
...
+        }
+    }
+
+	@Override
+	public void deleteSlidesFromSequence(String slideId, String moduleId) throws SlideDoesNotExistException {
+
+		List<Sequence> sequences = sequenceRepo.findSequencesForModule(moduleId);
+		// Remove slides that refer Sequence
+		for (int i = 0; i < sequences.size(); i++) {
+			if (sequences.get(i).getSlides().get(i).getId().equals(slideId)) {
```
jdamerow: what is this doing? why are you only looking at the i'th slide of any i'th sequence?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DeleteSlideController.java
```diff
@@ -0,0 +1,49 @@
...
+		} catch (SlideDoesNotExistException slideDoesNotExistException) {
+			logger.error("Could not delete slide.", slideDoesNotExistException);
+			return new ResponseEntity<String>("Sorry, unable to delete the slide. Please try later",
+					HttpStatus.NOT_FOUND);
+		}
+		return new ResponseEntity<String>("Ok", HttpStatus.OK);
+	}
+
+	@RequestMapping(value = "/staff/module/{id}/slide/{slideId}/{hasSequence}", method = RequestMethod.DELETE)
+	public ResponseEntity<String> deleteSlideFromSequence(@PathVariable("id") String moduleId,
```
jdamerow: why is that a separate method? the controller should just delete a slide. the manager makes sure it's delete everywhere it needs to.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
ashmichheda: The sequences is the total sequences belonging to a module. So to delete reference of a slide belonging to that specific sequence, I have to loop through it.

jdamerow: I'm confused. "So to delete reference of a slide belonging to that specific sequence", what specific sequence? Right now you take the first sequence and check if the first slide has a specific slide id, then you take the second sequence and check if the second slide has a specific slide id, then you take the third sequence and its third slide, etc. Why?

ashmichheda: Because, findSequencesForModule() is giving me all the sequences belonging to the module and not Slide. 
Hence, I am looping each sequence object from a list of sequences object and matching it with the selected slide, to delete it.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -46,4 +54,36 @@ public ISlide getSlide(String slideId) {
...
+							}
+						}
+						
+					}
+				}
+			}
+		} catch (IllegalArgumentException exception) {
+			throw new SlideDoesNotExistException(exception);
+		}
+	}
```
jdamerow: formatting!!!!

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -46,4 +54,36 @@ public ISlide getSlide(String slideId) {
     public void updateSlide(Slide slide) {
         slideRepo.save((Slide) slide);
     }
+    
+	@Override
+	public void deleteSlideById(String slideId, String moduleId, String flag) throws SlideDoesNotExistException {
+		try {
+
+			if (flag.equals("0")) {
```
jdamerow: why do you need this?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -46,4 +54,36 @@ public ISlide getSlide(String slideId) {
...
+				slideRepo.delete((Slide) getSlide(slideId));
+			}
+			// Slide part of another sequence
+			else {
+				List<Sequence> sequences = sequenceRepo.findSequencesForModule(moduleId);
+				int sizeOfSequences = sequences.size();
+				for (int i = 0; i < sizeOfSequences; i++) {
```
jdamerow: use for each loops and it'll make this a lot cleaner

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -46,4 +54,36 @@ public ISlide getSlide(String slideId) {
...
+					
+					if (sequences.get(i).getSlides().size() > 0) {
+						int sizeOfSlides = sequences.get(i).getSlides().size();
+						
+						for(int j = 0; j < sizeOfSlides; j++) {
```
jdamerow: for each loop

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -46,4 +54,36 @@ public ISlide getSlide(String slideId) {
...
+							if (sequences.get(i).getSlides().get(j).getId().equals(slideId)) {
+								sequences.get(i).getSlides().remove(j);
+								sequenceRepo.save(sequences.get(i));
+								slideRepo.delete((Slide) getSlide(slideId));
```
jdamerow: why do you delete the slide here? what it it's still in another sequence? that should be the last line in thee method

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -46,4 +54,36 @@ public ISlide getSlide(String slideId) {
...
+			if (flag.equals("0")) {
+				slideRepo.delete((Slide) getSlide(slideId));
+			}
+			// Slide part of another sequence
+			else {
+				List<Sequence> sequences = sequenceRepo.findSequencesForModule(moduleId);
+				int sizeOfSequences = sequences.size();
+				for (int i = 0; i < sizeOfSequences; i++) {
+					
+					if (sequences.get(i).getSlides().size() > 0) {
```
jdamerow: you don't need this. you can iterate over empty lists, just no over null.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -46,4 +54,36 @@ public ISlide getSlide(String slideId) {
     public void updateSlide(Slide slide) {
         slideRepo.save((Slide) slide);
     }
+    
+	@Override
+	public void deleteSlideById(String slideId, String moduleId, String flag) throws SlideDoesNotExistException {
```
jdamerow: what is the flag for?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DeleteSlideController.java
```diff
@@ -0,0 +1,37 @@
...
+			@PathVariable("slideId") String slideId, @PathVariable("flag") String flag) {
+		try {
+			slideManager.deleteSlideById(slideId, moduleId, flag);
+		} catch (SlideDoesNotExistException slideDoesNotExistException) {
+			logger.error("Could not delete slide.", slideDoesNotExistException);
+			return new ResponseEntity<String>("Sorry, unable to delete the slide. Please try later",
+					HttpStatus.NOT_FOUND);
+		}
+		return new ResponseEntity<String>("Ok", HttpStatus.OK);
+	}
```
jdamerow: formatting!!!

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DeleteSlideController.java
```diff
@@ -0,0 +1,37 @@
...
+    private ISlideManager slideManager;
+
+	@RequestMapping(value = "/staff/module/{id}/slide/{slideId}/{flag}", method = RequestMethod.DELETE)
+	public ResponseEntity<String> deleteSlide(@PathVariable("id") String moduleId,
+			@PathVariable("slideId") String slideId, @PathVariable("flag") String flag) {
+		try {
+			slideManager.deleteSlideById(slideId, moduleId, flag);
+		} catch (SlideDoesNotExistException slideDoesNotExistException) {
+			logger.error("Could not delete slide.", slideDoesNotExistException);
+			return new ResponseEntity<String>("Sorry, unable to delete the slide. Please try later",
```
jdamerow: no, don't try again later. the slide doesn't exist, why would trying again later help with anything?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
ashmichheda: I am using the flag, to handle the condition when slide has any sequences.
When flag is 0 - The slide has no sequences.
Else, - It has sequences, and it deletes the references from those sequences.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
ashmichheda: Yes, it's a part of the sequence, so I am deleting it's references in the loop itself.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
jdamerow: why do you need that? the else block takes care of that no matter what.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
jdamerow: this line will be executed every time a sequence includes the slide; so if two sequences  contain the slide you call delete slide as soon as the first sequence contains the slide (which will probably raise an exception since the slide is still in another sequence).

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
ashmichheda: But, then 
When I am deleting a slide that has no sequence, for the below code

List<Sequence> sequences = sequenceRepo.findSequencesForModule(moduleId);

Won't I be getting sequence list as size 0, I will have to add checks then, if the size is 0 then just delete the slide, else delete the references.
Else, while looping it might throw an exception?

ashmichheda: I will try without the flag and check.

jdamerow: you can loop over an empty list (as long as the list is not null). `for (Slide slide : slides) {}` works just fine for empty lists.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
ashmichheda: Yes, I got it. I will do the changes.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
ashmichheda: Okay

jdamerow:
conflicts

ashmichheda:
I removed the conflicts.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -46,4 +53,25 @@ public ISlide getSlide(String slideId) {
...
+                    if(slideIds.getId().equals(slideId)) {
+                        sequence.getSlides().remove(slideIds);
+                        sequenceRepo.save(sequence);
+                        continue outer;
+                    }
+                }
+            }
+            slideRepo.delete((Slide) getSlide(slideId));
+
+        } catch(IllegalArgumentException exception) {
```
jdamerow: try catch should be around the code that actually throws it

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -46,4 +53,25 @@ public ISlide getSlide(String slideId) {
...
     }
+
+    @Override
+    public void deleteSlideById(String slideId, String moduleId) throws SlideDoesNotExistException {
+
+        try {
+            List<Sequence> sequences = sequenceRepo.findSequencesForModule(moduleId);
+            outer: for(Sequence sequence : sequences) {
+                for(ISlide slideIds : sequence.getSlides()) {
+                    if(slideIds.getId().equals(slideId)) {
```
jdamerow: why is this variable called `slideIds`? It's a slide, isn't it?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -46,4 +53,25 @@ public ISlide getSlide(String slideId) {
...
+                        sequence.getSlides().remove(slideIds);
```
jdamerow: are you sure this works and doesn't throw an exception? typically you have to use an iterator since you can't modify a list that you are currently iterating over.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -46,4 +53,25 @@ public ISlide getSlide(String slideId) {
...
+                        sequenceRepo.save(sequence);
+                        continue outer;
```
jdamerow: I wouldn't do that. Right now it's not possible to have a slide appear more than once in a sequence, but it might make sense to change that.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ModuleManager.java
```diff
@@ -67,8 +68,12 @@ public IModule getModule(String id) {
     }
 
     @Override
-    public List<ISlide> getModuleSlides(String moduleId) {
-        return new ArrayList<>(slideRepo.findSlidesForModule(moduleId));
+    public HashMap<ISlide, Integer> getModuleSlides(String moduleId) {
```
jdamerow: is that used anywhere?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
ashmichheda: Yes, it's working. 
on cases with slide not attached to any Sequence. And in cases where slides are attached to a Sequence.

ashmichheda: Yes, it was throwing concurrent modification exception before. After i added continue outer, didn't throw exception and working as expected.

jdamerow: that's not really a solution though (because of my next comment). an iterator should fix this though.

ashmichheda: okay.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ModuleManager.java
ashmichheda: Yes it is used

vspace/src/main/webapp/WEB-INF/views/staff/modules/module.jsp
```diff
@@ -7,15 +7,50 @@
...
+	        	$("#"+slideId).closest('.slide').remove();
+	        },
+	        error: function(data) {
+	        	var alert = $('<div class="alert alert-danger alert-dismissible fade show" role="alert"><p>We are sorry but something went wrong. Please try again later.</p><button type="button" class="close" data-dismiss="alert" aria-label="Close"><span aria-hidden="true">&times;</span></button></div>');
+	        	$('.error').append(alert);
+	        }
+	    });
+	} else {
+		slideIdToDelete = slideId;
+		$("#deleteSlideAlert").show();
```
jdamerow: let's make it an ajax call to check if a slide is used in a sequence, otherwise the info might very out of date (e.g. someone works on something else on the page or takes a break and then comes back). Someone else might have added the slide to sequence or removed it in the meantime. To reduce the risk, you should check right before you're about to delete the slide and then show a warning.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ModuleManager.java
jdamerow: this is pretty ugly and avoidable (see my comment below). Also, I believe it is used in the sequence controller, which I don't think you have adapted.

vspace/src/main/java/edu/asu/diging/vspace/core/data/SequenceRepository.java
```diff
@@ -13,7 +13,7 @@
 @JaversSpringDataAuditable
 public interface SequenceRepository extends PagingAndSortingRepository<Sequence, String> {
     
-    @Query("SELECT d FROM Sequence d WHERE d.module.id = ?1")
-    public List<Sequence> findSequencesForModule(String moduleId);
-    
+	@Query("SELECT d FROM Sequence d WHERE d.module.id = ?1")
+	public List<Sequence> findSequencesForModule(String moduleId);
+
```
jdamerow: why is the indentation changed here?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
jdamerow: formatting!

vspace/src/main/webapp/WEB-INF/views/staff/modules/module.jsp
ashmichheda: Okay

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -46,4 +53,46 @@ public ISlide getSlide(String slideId) {
...
+
+        List<Sequence> sequences = sequenceRepo.findSequencesForModule(moduleId);
+        for(Sequence sequence : sequences) {
+
+            int sizeOfSlides = sequence.getSlides().size();
+            for(int i = 0; i< sizeOfSlides; i++) {
+                if(sequence.getSlides().get(i).getId().equals(slideId)) {
+                    sequence.getSlides().remove(i);
+                    sequenceRepo.save(sequence);
+                }
```
jdamerow: so, i'm fairly certain this will have issues.

you are deleting elements from a list you're iterating over. So assume you remove the slide with index 0. Next iteration you take slide with index 1. At this point, the slide that previously had index 1, now has index 0 (because slide with index 0 was delete and all slides move one forward). Also, this will probably create an index out of bounds exception since if before deleting a slide the list had 3 slides, it now has 2 and a slide with i=2 will now not exist anymore. You can try it out easily:
```
        List<String> list = new ArrayList<>();
        list.add("aa");
        list.add("ab");
        list.add("bc");
        list.add("cd");
        
        int s = list.size();
        for(int i=0; i< s; i++) {
            System.out.println(i);
            if(list.get(i).startsWith("a")) {
                list.remove(i);
                System.out.println("deleted");
            }
        }
```
As I said before, you should use an iterator.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -46,4 +53,46 @@ public ISlide getSlide(String slideId) {
...
+    @Override
+    public boolean checkSlideHasSequence(String slideId, String moduleId) {
+
+        List<Sequence> sequences = sequenceRepo.findSequencesForModule(moduleId);
+        boolean slideHasSequence = false;
+        for(Sequence sequence : sequences) {
+            int sizeOfSlides = sequence.getSlides().size();
+            for(int i = 0; i< sizeOfSlides; i++) {
+                if(sequence.getSlides().get(i).getId().equals(slideId)) {
+                    slideHasSequence = true;
```
jdamerow: just return true here.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -46,4 +53,46 @@ public ISlide getSlide(String slideId) {
...
+                    break;
+                }
+            }
+            if(slideHasSequence)
```
jdamerow: unnecessary if you incorporate above change

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -46,4 +53,46 @@ public ISlide getSlide(String slideId) {
...
+    }
+
+    @Override
+    public boolean checkSlideHasSequence(String slideId, String moduleId) {
+
+        List<Sequence> sequences = sequenceRepo.findSequencesForModule(moduleId);
+        boolean slideHasSequence = false;
+        for(Sequence sequence : sequences) {
+            int sizeOfSlides = sequence.getSlides().size();
+            for(int i = 0; i< sizeOfSlides; i++) {
```
jdamerow: just use a for each loop here, it's so much easier to read and maintain!

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DeleteSlideController.java
```diff
@@ -0,0 +1,50 @@
...
+            slideManager.deleteSlideById(slideId, moduleId);
+        } catch (SlideDoesNotExistException slideDoesNotExistException) {
+            logger.error("Could not delete slide.", slideDoesNotExistException);
+            return new ResponseEntity<String>("Sorry, unable to delete the slide. The slide does not exist.",
+                    HttpStatus.NOT_FOUND);
+        }
+        return new ResponseEntity<String>("Ok", HttpStatus.OK);
+    }
+
+    @RequestMapping(value = "/staff/module/{id}/slide/{slideId}/sequences", method = RequestMethod.POST)
```
jdamerow: this path implies it's returning a list of sequences of some kind, so either return a list of sequence ids or something or change the path. I'd opt for version one (returning a list of sequence ids (and possible sequence titles)) since this will allow to e.g. highlight the sequences the slide is contained or list them in another way.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DeleteSlideController.java
```diff
@@ -0,0 +1,50 @@
...
+    public ResponseEntity<String> checkSlideInSequence(@PathVariable("id") String moduleId,
+            @PathVariable("slideId") String slideId) {
+
+        boolean checkSlideHasSequence = slideManager.checkSlideHasSequence(slideId, moduleId);
+        if (checkSlideHasSequence)
+            return new ResponseEntity<String>("1", HttpStatus.OK);
+        else
+            return new ResponseEntity<String>("0", HttpStatus.OK);
```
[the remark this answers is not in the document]
ashmichheda: I want to return 1 or 0, but still the response should be a success. Hence, I have return HttpStatus as 'ok' on if and else conditions.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
ashmichheda: Even if i am using iterator, I will still need to reference an index to retrieve the element.

ashmichheda: I am doing something like this.

int i = 0;
            Iterator<Sequence> sequenceIterator = sequences.iterator();
            while(sequenceIterator.hasNext()) {
                if(sequenceIterator.next().getSlides().get(i).getId().equals(slideId)){
                    sequenceIterator.next().getSlides().remove(i);
                    sequenceRepo.save(sequence);
                    sequenceIterator = sequences.iterator();
                    
                }
            }

ashmichheda: Also, I am not removing the list object. I am removing the slide object from the sequence list.
Hence, the loop should work as expected.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ModuleManager.java
jdamerow: where? it's not used in any of your changes.

vspace/src/main/java/edu/asu/diging/vspace/core/services/ISlideManager.java
```diff
@@ -1,17 +1,23 @@
...
-    ISlide createSlide(String moduleId, SlideForm slideForm);
+	ISlide createSlide(String moduleId, SlideForm slideForm);
 
-    void updateSlide(Slide slide);
+	void updateSlide(Slide slide);
 
-   
+	void deleteSlideById(String slideId, String moduleId) throws SlideDoesNotExistException;
+	
+	List<Sequence> checkSlideHasSequence(String slideId, String moduleId);
```
jdamerow: bad naming: method name that starts with `check...` implies it return true or false, but you return a list

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -46,4 +55,45 @@ public ISlide getSlide(String slideId) {
...
         slideRepo.save((Slide) slide);
     }
+
+    @Override
+    public void deleteSlideById(String slideId, String moduleId) throws SlideDoesNotExistException {
+
+        List<Sequence> sequences = sequenceRepo.findSequencesForModule(moduleId);
+        for(Sequence sequence : sequences) {
+            List<ISlide> slidesList = sequence.getSlides();
+            Iterator<ISlide> slideIterator = slidesList.iterator();
```
jdamerow: merge with line 64

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -46,4 +55,45 @@ public ISlide getSlide(String slideId) {
...
+
+        List<Sequence> sequences = sequenceRepo.findSequencesForModule(moduleId);
+        List<Sequence> sequenceSlides = new ArrayList<>();
+        for(Sequence sequence : sequences) {
+            List<ISlide> slidesList = sequence.getSlides();
+            Iterator<ISlide> slideIterator = slidesList.iterator();
+            while(slideIterator.hasNext()) {
+                if(slideIterator.next().getId().equals(slideId)){
+                    sequenceSlides.add(sequence);
+                    return sequenceSlides;
```
jdamerow: why do you return here?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -46,4 +55,45 @@ public ISlide getSlide(String slideId) {
...
+    }
+
+    @Override
+    public List<Sequence> checkSlideHasSequence(String slideId, String moduleId) {
+
+        List<Sequence> sequences = sequenceRepo.findSequencesForModule(moduleId);
+        List<Sequence> sequenceSlides = new ArrayList<>();
+        for(Sequence sequence : sequences) {
+            List<ISlide> slidesList = sequence.getSlides();
+            Iterator<ISlide> slideIterator = slidesList.iterator();
```
jdamerow: merge with line 88

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DeleteSlideController.java
```diff
@@ -0,0 +1,54 @@
...
+    @RequestMapping(value = "/staff/module/{id}/slide/{slideId}/sequences", method = RequestMethod.GET)
+    public ResponseEntity<List<Sequence>> checkSlideInSequence(@PathVariable("id") String moduleId,
+            @PathVariable("slideId") String slideId) {
+
+        List<Sequence> slideSequences = slideManager.checkSlideHasSequence(slideId, moduleId);
+        if (slideSequences.size() > 0) {
+            return new ResponseEntity<>(slideSequences, HttpStatus.OK);
+        } 
+        else {
+            return new ResponseEntity<>(slideSequences, HttpStatus.OK);
```
jdamerow: why do you have an if/else statement if you return the exact same thing?

vspace/src/main/java/edu/asu/diging/vspace/core/services/ISlideManager.java
```diff
@@ -1,17 +1,23 @@
...
-    ISlide createSlide(String moduleId, SlideForm slideForm);
+	ISlide createSlide(String moduleId, SlideForm slideForm);
 
-    void updateSlide(Slide slide);
+	void updateSlide(Slide slide);
 
-   
+	void deleteSlideById(String slideId, String moduleId) throws SlideDoesNotExistException;
+	
+	List<Sequence> getSlideSequences(String slideId, String moduleId);
```
jdamerow: wrong indentation

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -46,4 +55,42 @@ public ISlide getSlide(String slideId) {
...
+
+            slideRepo.delete((Slide) getSlide(slideId));
+
+        } catch(IllegalArgumentException exception) {
+            throw new SlideDoesNotExistException(exception);
+        }
+    }
+
+    @Override
+    public List<Sequence> getSlideSequences(String slideId, String moduleId) {
```
jdamerow: needs test cases

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -46,4 +55,42 @@ public ISlide getSlide(String slideId) {
     public void updateSlide(Slide slide) {
         slideRepo.save((Slide) slide);
     }
+
+    @Override
+    public void deleteSlideById(String slideId, String moduleId) throws SlideDoesNotExistException {
```
jdamerow: needs test cases

jdamerow:
Jenkins, test this please.

jdamerow:
going to a module throws an exception
```
javax.el.PropertyNotFoundException: Property [key] not found on type [edu.asu.diging.vspace.core.model.impl.Slide]
	javax.el.BeanELResolver$BeanProperties.get(BeanELResolver.java:260)
	javax.el.BeanELResolver$BeanProperties.access$300(BeanELResolver.java:212)
	javax.el.BeanELResolver.property(BeanELResolver.java:347)

```

ashmichheda:
> going to a module throws an exception
> 
> ```
> javax.el.PropertyNotFoundException: Property [key] not found on type [edu.asu.diging.vspace.core.model.impl.Slide]
> 	javax.el.BeanELResolver$BeanProperties.get(BeanELResolver.java:260)
> 	javax.el.BeanELResolver$BeanProperties.access$300(BeanELResolver.java:212)
> 	javax.el.BeanELResolver.property(BeanELResolver.java:347)
> ```

I think this is because, I reverted the changes for getModuleSlides(). That is why the property key, is not found.
I need the changes made.

vspace/src/main/webapp/WEB-INF/views/staff/modules/module.jsp
```diff
@@ -7,15 +7,64 @@
...
+	        type: 'DELETE',
+	        cache       : false,
+	        contentType : false,
+	        success: function(data) {
+	        	$("#"+slideId).closest('.slide').remove();
+	        },
+	        error: function(data) {
+	        	var alert = $('<div class="alert alert-danger alert-dismissible fade show" role="alert"><p>We are sorry but something went wrong. Please try again later.</p><button type="button" class="close" data-dismiss="alert" aria-label="Close"><span aria-hidden="true">&times;</span></button></div>');
+	        	$('.error').append(alert);
+	        }
```
jdamerow: Indentation

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SlideManagerTest.java
```diff
@@ -0,0 +1,85 @@
...
+        Sequence sequenceObj = new Sequence();
+        sequenceObj.setId("SEQ000000004");
+        sequenceObj.setSlides(slidesList);
+        List<Sequence> sequencesList = new ArrayList<>();
+        sequencesList.add(sequenceObj);
+        Mockito.when(sequenceRepo.findSequencesForModule(moduleId)).thenReturn(sequencesList);
+        
+        List<Sequence> actualSequenceSlideList = slideManagerToTest.getSlideSequences(slideId, moduleId);
+        
+        Assert.assertEquals(actualSequenceSlideList.get(0).getSlides().get(0).getId(), sequencesList.get(0).getSlides().get(0).getId());
```
jdamerow: just create a variable at the beginning for the slide (before adding it to slideList) and then call get id on that slide object rather than this long expression `list.get(0).getSlides()...`. makes it hard to read.

jdamerow: also, actually, the method does not need to make sure that the slides are in the sequence, does it? It should needs to return the correct sequence object. So you're actually not testing what the method should do.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SlideManagerTest.java
```diff
@@ -0,0 +1,85 @@
...
+        Assert.assertEquals(actualSequenceSlideList.size(), sequencesList.size()); 
+    }
+
```
jdamerow: test case for slide with no sequences is missing.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SlideManagerTest.java
```diff
@@ -0,0 +1,85 @@
...
+    public void test_deleteSlideById() throws SlideDoesNotExistException {
+        String slideId = "SLI0000000040";
+        String moduleId = "MOD000000010";
+
+        List<ISlide> slidesList = Arrays.asList(new Slide());
+        slidesList.get(0).setId("SLI0000000040");
+        
+        Sequence sequenceObj = new Sequence();
+        sequenceObj.setId("SEQ000000004");
+        sequenceObj.setSlides(slidesList);
```
jdamerow: since you need a sequence object with a slide more often, you should create that in the setup method.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -46,4 +55,45 @@ public ISlide getSlide(String slideId) {
...
+
+        List<Sequence> sequences = sequenceRepo.findSequencesForModule(moduleId);
+        for(Sequence sequence : sequences) {
+            
+            List<ISlide> sequenceSlides = new ArrayList<ISlide> (sequence.getSlides());
+            Iterator<ISlide> slideIterator = sequenceSlides.iterator();
+            while(slideIterator.hasNext()) {
+                //System.out.println("Slide ID: "+slideIterator.next().getId());
```
jdamerow: remove

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SlideManagerTest.java
```diff
@@ -0,0 +1,85 @@
...
+        List<Sequence> sequencesList = new ArrayList<>();
+        sequencesList.add(sequenceObj);
+        Mockito.when(sequenceRepo.findSequencesForModule(moduleId)).thenReturn(sequencesList);
+        
+        ISlide slide = new Slide();
+        slide.setId(slideId);
+        Mockito.doReturn(slide).when(slideManagerToTestSpy).getSlide(Mockito.any());
+        
+        slideManagerToTestSpy.deleteSlideById(slideId, moduleId);
```
jdamerow: you don't need a spy object here, just mock the getById call on the repository. This way even if the implementation changes and the delete method calls the repository directly, the test won't fail.

jdamerow:
Jenkins, test this please.

jdamerow:
clicking on the delete button for a slide in a sequence doesn't do anything

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SlideManagerTest.java
ashmichheda: Without using spy, I tried to call using the repo directly, but it threw an exception as null data was getting returned.
To mock methods in the same test class, answers in the net suggested using spy object.

jdamerow: I'm not 100% what you mean but if you mock `getById` (or whicheever method is called by `getSlide`) to return something, it won't throw a null pointer.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SlideManagerTest.java
```diff
@@ -0,0 +1,102 @@
...
+        // Positive scenario - Slide present in Sequence
+        
+        sequenceObj.setId(sequenceId);
+        sequenceObj.setSlides(slidesList);
+        List<Sequence> sequencesList = new ArrayList<>();
+        sequencesList.add(sequenceObj);
+        Mockito.when(sequenceRepo.findSequencesForModule(moduleId)).thenReturn(sequencesList);
+        List<Sequence> slideSequencePresent = slideManagerToTest.getSlideSequences(slideId, moduleId);
+        
+        if(slideSequencePresent.size() > 0) {
```
jdamerow: what iv slideSequencePresent is not greater than 0? Shouldn't the testcase fail?

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SlideManagerTest.java
```diff
@@ -0,0 +1,102 @@
...
+        sequenceObj.setId(sequenceId);
+        sequenceObj.setSlides(slidesList);
+        List<Sequence> sequencesList = new ArrayList<>();
+        sequencesList.add(sequenceObj);
+        Mockito.when(sequenceRepo.findSequencesForModule(moduleId)).thenReturn(sequencesList);
+        
+        // using mockito spy object
+        ISlide slide = new Slide();
+        slide.setId(slideId);
+        Mockito.doReturn(slide).when(slideManagerToTestSpy).getSlide(Mockito.any());
```
jdamerow: why do you need a spy object here?

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SlideManagerTest.java
```diff
@@ -0,0 +1,102 @@
...
+            Assert.assertEquals(slideSequencePresent.size(), sequencesList.size()); 
+        }
+        
+        // Negative Scenario - Slide not present in Sequence
+        List<Sequence> actualSequenceSlideListNotPresent = slideManagerToTest.getSlideSequences(slideId, moduleId);
+        Assert.assertNotSame(actualSequenceSlideListNotPresent, slideSequencePresent);
+    }
+    
+    @Test
+    public void test_deleteSlideById() throws SlideDoesNotExistException {
```
jdamerow: you want at least four test cases: slide with id doesn't exist, slide is in no sequence, slide is in sequence, slide is in 4 sequences.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SlideManagerTest.java
ashmichheda: I am using spy object to mock the method of the same class. Otherwise, it's calling the actual method to get data which is null.

jdamerow: you should do that on the manager that's being tested, you don't need a spy for that

ashmichheda: Yes, I am doing it on SlideManager that is being tested.
I tried to mock it normally, but it was still hitting DB and returning null. I checked on net, it suggested using Spy object, to mock method of the same class.

jdamerow: you should mock the repository method that getSlide calls, so that get slides returns soemthing meaningful.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SlideManagerTest.java
```diff
@@ -0,0 +1,124 @@
...
+        sequenceObj.setId(sequenceId);
+        sequenceObj.setSlides(slidesList);
+        List<Sequence> sequencesList = new ArrayList<>();
+        sequencesList.add(sequenceObj);
+        Mockito.when(sequenceRepo.findSequencesForModule(moduleId)).thenReturn(sequencesList);
+        
+        // using mockito spy object
+        ISlide slide = new Slide();
+        slide.setId(slideId);
+        Mockito.doReturn(slide).when(slideManagerToTestSpy).getSlide(Mockito.any());
```
jdamerow: same as before, why is that using a spy object?

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SlideManagerTest.java
```diff
@@ -0,0 +1,124 @@
...
+        slideManagerToTestSpy.deleteSlideById(slideId, moduleId);  
+        Mockito.verify(slideRepo).delete((Slide) slide);
+    }
+    
+    @Test
+    public void test_deleteSlideById_slideIdNotPresent() throws SlideDoesNotExistException {
+    	
+    	
+    	Mockito.when(slideRepo.findById(Mockito.anyString())).thenReturn(Optional.empty());
+        Assert.assertNull(slideManagerToTest.getSlide(Mockito.anyString()));
```
jdamerow: why are you asserting this here?

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SlideManagerTest.java
```diff
@@ -0,0 +1,124 @@
...
+    	
+    	sequenceObj.setId(sequenceId);
```
jdamerow: indentation

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SlideManagerTest.java
```diff
@@ -0,0 +1,124 @@
...
+        sequenceObj.setSlides(slidesList);
+        List<Sequence> sequencesList = new ArrayList<>();
+        sequencesList.add(sequenceObj);
+        Mockito.when(sequenceRepo.findSequencesForModule(moduleId)).thenReturn(sequencesList);
+        
+        // using mockito spy object
+        ISlide slide = new Slide();
+        slide.setId(slideIdNotPresent);
+        Mockito.doReturn(slide).when(slideManagerToTestSpy).getSlide(Mockito.any());
```
jdamerow: same as above, why a spy?

ashmichheda: If I use the actualy SlideManager object, that is slideManagerToTest and then call it's class defined methods, while debugging it shows not a mock object and tries to hit actual DB, thereby returning null.

Hence I am using a spy object.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -1,49 +1,98 @@
...
+        Slide slideObj = (Slide) getSlide(slideId);
+        for(Sequence sequence : sequences) {
+			for(int i = 0; i<sequence.getSlides().size(); i++) {
+				String slideIdToCompare = sequence.getSlides().get(i).getId();
+				if(slideIdToCompare.equals(slideId)) {
+					sequence.getSlides().remove(slideObj);
+					sequenceRepo.save(sequence);
+				}
+			}
+        }
```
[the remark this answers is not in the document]
jdamerow: indentation is still wrong

vspace/src/main/webapp/WEB-INF/views/staff/modules/module.jsp
jdamerow: as I said, indentation

vspace/src/main/webapp/WEB-INF/views/staff/modules/module.jsp
```diff
@@ -7,15 +7,64 @@
...
+	        success: function(data) {
+	        	if(data.length > 0){
+	        		$("#deleteSlideAlert").show();
+		        	slideIdToDelete = slideId;
+	        	}
+	        	else{
+	        		deleteSlide(slideId);
+	        	}	
+	        }
+	    });
```
jdamerow: same here

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SlideManagerTest.java
```diff
@@ -0,0 +1,112 @@
...
+	@Mock
+	private SlideRepository slideRepo;
+
+	@InjectMocks
+	private SlideManager slideManagerToTest = new SlideManager();
+
+	// setting common used variables and Objects
+	String slideId, moduleId, sequenceId, slideIdNotInSequence;
+	List<ISlide> slidesList = Arrays.asList(new Slide());
+	Sequence sequenceObj;
```
jdamerow: should all be private. not as important here as in the main code, but still a good practice.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SlideManagerTest.java
```diff
@@ -0,0 +1,112 @@
...
+		sequenceObj = new Sequence() ;
+		slideId = "SLI000000002";
+		slideIdNotInSequence = "SLI000000219";
+		moduleId = "MOD000000002";
+		sequenceId = "SEQ000000004";
+		slidesList.get(0).setId(slideId);
+	}    
+
+	@Test
+	public void test_slideSequence_slideExistsInSequence() {
```
jdamerow: in the code you call `getSlideSequences`. please read our style guidelines for tests.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SlideManagerTest.java
```diff
@@ -0,0 +1,112 @@
...
+		Assert.assertEquals(actualSlideIdSequence, expectedSlideIdSequence);
+		Assert.assertEquals(slideSequencePresent.size(), sequencesList.size()); 
+
+	}
+
+	@Test
+	public void test_slideSequence_slideNotExistsInSequence() {
+
+		List<Sequence> slideSequencePresent = slideManagerToTest.getSlideSequences(slideId, moduleId);
+		List<Sequence> actualSequenceSlideListNotPresent = slideManagerToTest.getSlideSequences(slideIdNotInSequence, moduleId);
```
jdamerow: you need to make sure that what the manager returns is what you expect. if `getSlideSequences` shouldn't return anything for the provided parameters, that's what you need to test (e.g. returned list is empty or null), not that it is not the same as for ids that should return something. there could be plenty of ways `assertNotSame` would be true but the method is still doing the wrong thing.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SlideManagerTest.java
```diff
@@ -0,0 +1,112 @@
...
+		Mockito.when(slideRepo.findById(Mockito.anyString())).thenReturn(Optional.empty());
+
+		sequenceObj.setId(sequenceId);
+		sequenceObj.setSlides(slidesList);
+		List<Sequence> sequencesList = new ArrayList<>();
+		sequencesList.add(sequenceObj);
+		Mockito.when(sequenceRepo.findSequencesForModule(moduleId)).thenReturn(sequencesList);
+
+		slideManagerToTest.deleteSlideById(slideId, moduleId);
+	}
```
jdamerow: what about the cases when the slide is used in sequences?

jdamerow:
Jenkins, test this please.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SlideManagerTest.java
ashmichheda: that's the first method in the Test file.

jdamerow:
clicking on the delete button just throws a javascript error. I believe you need the fix that Prashant put in his code for sequences (I think it was a json ignore annotation for the start sequence or module property).

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SlideManagerTest.java
ashmichheda: I am passing slideId which is not in sequence which is returning a null list.
And I am passing slideId which is there in sequence which is returning data.
So, I am checking them.
What other check do I need to make?

jdamerow:
https://github.com/diging/virtual-spaces-2.0/blob/451232150eceffa708a45b6582e9233731d0444d/vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/Module.java#L38

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SlideManagerTest.java
jdamerow: there should be one for when slide is not in a sequence, one when slide doesn't exist, and one for slides being in two (or more sequences)

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SlideManagerTest.java
ashmichheda: Okay, I will change it.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SlideManagerTest.java
jdamerow: if `getSlideSequences` should return null for the given id, then you need to assert exactly that.

ashmichheda: Is Assert.assertNotSame not a proper way to check?

ashmichheda: It's returning a list of size 0, hence I am checking that.

jdamerow: you're not checking if the returned list has size 0 but that the returned list is not the same as a different list. if the returned list has 2 elements and you check if it is not the same as a list that has 1 element the test would pass although the method doesn't do what it's supposed to be doing.

ashmichheda: Oh okay.

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/Sequence.java
```diff
@@ -31,6 +33,8 @@
     @OneToOne(targetEntity = Module.class)
     private IModule module;
 
+    //-------- @JsonIgnore used as this Slide will be returned in a controller
```
jdamerow: I'm pretty sure that you need this in the controller?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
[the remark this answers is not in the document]
jdamerow: ?

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/Sequence.java
ashmichheda: Do I need to change?

jdamerow:
the confirmation modal should pop up in the middle of the screen and the rest should be greyed out

vspace/src/main/webapp/WEB-INF/views/staff/modules/module.jsp
```diff
@@ -125,8 +176,27 @@ $(document).ready(function($) {
...
+  
+<div id="deleteSlideAlert" class="modal fade" role="dialog" >
+	<div class="modal-dialog">
+	<div class="modal-content">
+	<div class="modal-header">
+		<h4 class="modal-title" id="deleteModalTitle">Confirm Deletion?</h4>
+		<button type="button" id="closeSlide" class="close" data-dismiss="modal" aria-hidden="true">×</button>
+	</div>
+	<div class="modal-body">
+		<p> This Slide is a part of another Sequence. Are you sure you want to delete it?</p>
```
jdamerow, line 194: This should say:
"This slide is part of X sequence(s). Are you sure you want to delete it?"

where X is the number of sequences it's part of. Even better, it should list the sequences it's part of.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SlideManagerTest.java
```diff
@@ -0,0 +1,147 @@
...
+        sequenceObj.setId(sequenceId);
+        sequenceObj.setSlides(slidesList);
+        List<Sequence> sequencesList = new ArrayList<>();
+        sequencesList.add(sequenceObj);
+        Mockito.when(sequenceRepo.findSequencesForModule(moduleId)).thenReturn(sequencesList);
+
+        List<Sequence> slideSequencePresent = slideManagerToTest.getSlideSequences(slideId, moduleId);
+        List<Sequence> actualSequenceSlideListNotPresent = slideManagerToTest.getSlideSequences(slideIdNotInSequence,
+                moduleId);
+        Assert.assertNotEquals(slideSequencePresent, actualSequenceSlideListNotPresent);
```
jdamerow, line 87: I'm confused what are you comparing here? If the slide is not in any sequence, you expect the getSlideSequence to return an empty list, don't you? Why would you compare that to a different list? Espeically one you are also geetting from the method you want to test. If there's a bug, you can't trust any of these.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SlideManagerTest.java
```diff
@@ -0,0 +1,147 @@
...
+        sequenceObj.setId(sequenceId);
+        sequenceObj.setSlides(slidesList);
+
+        sequenceObjOther.setId(sequenceId);
+
+        List<Sequence> sequencesList = new ArrayList<>();
+        sequencesList.add(sequenceObj);
+        Mockito.when(sequenceRepo.findSequencesForModule(moduleId)).thenReturn(sequencesList);
+
+        slideManagerToTest.deleteSlideById(slideId, moduleId);
```
jdamerow, line 121: where is your assert or verify statement? what are you testing?

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SlideManagerTest.java
```diff
@@ -0,0 +1,147 @@
...
+        Mockito.when(sequenceRepo.findSequencesForModule(moduleId)).thenReturn(sequencesList);
+
+        Slide slide = new Slide();
+        slide.setId(slideId);
+        Optional<Slide> slideObj = Optional.of(slide);
+        Mockito.when(slideRepo.findById(slide.getId())).thenReturn(slideObj);
+        slideManagerToTest.deleteSlideById(slideId, moduleId);
+        Mockito.verify(slideRepo).delete((Slide) slide);
+
+        slideManagerToTest.deleteSlideById(slideId, moduleId);
```
jdamerow, line 145: why do you call this again here?

vspace/src/main/webapp/WEB-INF/views/staff/modules/module.jsp
```diff
@@ -5,17 +5,64 @@
...
+	        contentType : false,
+	        success: function(data) {
+	        	if(data.length > 0){
+	        		var keys = Object.keys(data);
+	        		for(var i = 0; i < keys.length;i++){
+	        			$("#sequenceData").text(data[keys[i]].name);
+	        		}
+	        		
+	        		$('#deleteSlideAlert').modal('show'); // popup #myModal id modal
+		        	slideIdToDelete = slideId;
```
jdamerow, line 45: the slide id should be stored in the modal, not as global variable

vspace/src/main/webapp/WEB-INF/views/staff/modules/module.jsp
jdamerow: I really need you to fix the indentation. why is the first line (and all the following) of the function indented by like 12 spaces or something? same thing above.

vspace/src/main/webapp/WEB-INF/views/staff/modules/module.jsp
```diff
@@ -5,17 +5,64 @@
...
+	$("#closeSlide").click(function (){
+		slideIdToDelete = 0;
+		 $('#deleteSlideAlert').modal('hide'); 
+	});
+	
+	$("#cancelSlideDelButton").click(function () {
+		slideIdToDelete = 0;
+		$("#deleteSlideAlert").hide();
 	});
+	$("#deleteSlideAlert").draggable();
```
jdamerow, line 65: the alert does not need to be draggable

vspace/src/main/webapp/WEB-INF/views/staff/modules/module.jsp
```diff
@@ -125,8 +181,31 @@ $(document).ready(function($) {
...
+        <div class="modal-content">
+            <div class="modal-header">
+                <h4 class="modal-title" id="deleteModalTitle">Confirm
+                    Deletion?</h4>
+                <button type="button" id="closeSlide" class="close" data-dismiss="modal" aria-hidden="true">×</button>
+            </div>
+            <div class="modal-body">
+                <p>
+                    This Slide is a part of <span id="sequenceData"></span>
+                    Sequence. Are you sure you want to delete it?
```
jdamerow, line 202: this will be a list of sequences, right? Make it an actual html list and remove 'Sequence'.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SlideManagerTest.java
```diff
@@ -0,0 +1,143 @@
...
+
+        sequenceObj.setId(sequenceId);
+        sequenceObj.setSlides(slidesList);
+        List<Sequence> sequencesList = new ArrayList<>();
+        sequencesList.add(sequenceObj);
+        Mockito.when(sequenceRepo.findSequencesForModule(moduleId)).thenReturn(sequencesList);
+        List<Sequence> slideSequencePresent = slideManagerToTest.getSlideSequences(slideId, moduleId);
+
+        String actualSlideIdSequence = slideSequencePresent.get(0).getSlides().get(0).getId();
+        String expectedSlideIdSequence = sequencesList.get(0).getSlides().get(0).getId();
```
jdamerow, line 68: the expected slide id is `slideId`, not need for this line

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SlideManagerTest.java
```diff
@@ -0,0 +1,143 @@
...
+
+        Assert.assertEquals(actualSlideIdSequence, expectedSlideIdSequence);
```
jdamerow, line 70: I believe the order is expected value, actual value, not the otheer way around.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SlideManagerTest.java
```diff
@@ -0,0 +1,143 @@
...
+        Mockito.when(slideRepo.findById(Mockito.anyString())).thenReturn(Optional.empty());
+
+        sequenceObj.setId(sequenceId);
+        sequenceObj.setSlides(slidesList);
+
+        List<Sequence> sequencesList = new ArrayList<>();
+        sequencesList.add(sequenceObj);
+        Mockito.when(sequenceRepo.findSequencesForModule(moduleId)).thenReturn(sequencesList);
+        slideManagerToTest.deleteSlideById(slideId, moduleId);
+        Mockito.doThrow(SlideDoesNotExistException.class).when(slideRepo).delete(null);
```
jdamerow, line 118: a couple of things, first I"m pretty sure the repo doesn't throw a `SlideDoesNotExistException`. And second, you should not try to delete null. Check that first. Third, if you expect an exception to be thrown you need to have your test case expect that so the test passes.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SlideManagerTest.java
```diff
@@ -0,0 +1,143 @@
...
+        sequencesList.add(sequenceObj);
+        sequencesList.add(sequenceObjOther);
+        Mockito.when(sequenceRepo.findSequencesForModule(moduleId)).thenReturn(sequencesList);
+
+        Slide slide = new Slide();
+        slide.setId(slideId);
+        Optional<Slide> slideObj = Optional.of(slide);
+        Mockito.when(slideRepo.findById(slide.getId())).thenReturn(slideObj);
+        slideManagerToTest.deleteSlideById(slideId, moduleId);
+        Mockito.verify(slideRepo).delete((Slide) slide);
```
jdamerow, line 140: a) you also want to make sure the slide has been removed from the sequences.
b) you probably want another slide object in at least one of the sequences to make sure not all slides are being deleted.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SlideManagerTest.java
```diff
@@ -0,0 +1,159 @@
...
+        int slidesListSizeBefore = sequencesList.get(1).getSlides().size();
+        Mockito.when(sequenceRepo.findSequencesForModule(moduleId)).thenReturn(sequencesList);
+
+        ISlide slideObj = slidesList.get(0);
+        Mockito.when(slideRepo.findById(slideObj.getId())).thenReturn(Optional.of((Slide) slideObj));
+        slideManagerToTest.deleteSlideById(slideId, moduleId);
+        Mockito.verify(slideRepo).delete((Slide) slideObj);
+        int slidesListSizeAfter = sequencesList.get(1).getSlides().size();
+
+        Assert.assertNotEquals(slidesListSizeBefore, slidesListSizeAfter);
```
jdamerow, line 157: you want to make sure that the sequences do not contain the slide anymore

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -1,49 +1,104 @@
...
+        Slide slideObj = (Slide) getSlide(slideId);
+        if(slideObj == null) {
+            return;
+        }
+        for (Sequence sequence : sequences) {
+            for (int i = 0; i < sequence.getSlides().size(); i++) {
+                String slideIdToCompare = sequence.getSlides().get(i).getId();
+                if (slideIdToCompare.equals(slideId)) {
+                    sequence.getSlides().remove(slideObj);
+                    sequenceRepo.save(sequence);
```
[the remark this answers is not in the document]
ashmichheda: If there is slide sequence a - b - a then there should be 3 entries to DB table?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -1,49 +1,102 @@
...
+                sequence.getSlides().removeAll(slideObjToRemove);
+                sequenceRepo.save(sequence);
+            }
+        }
+        try {
+            
+            slideRepo.delete((Slide) getSlide(slideId));
+            
+        } catch (IllegalArgumentException exception) {
+            logger.error("Slide Id cannot be null. "+exception);
```
jdamerow, line 83: this needs to be checked in the beginning. if id is null, this should never happen.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SlideManagerTest.java
```diff
@@ -0,0 +1,159 @@
...
+        String slideIdBeforeDeletion = sequencesList.get(1).getSlides().get(0).getId();
+        Mockito.when(sequenceRepo.findSequencesForModule(moduleId)).thenReturn(sequencesList);
+
+        ISlide slideObj = slidesList.get(0);
+        Mockito.when(slideRepo.findById(slideObj.getId())).thenReturn(Optional.of((Slide) slideObj));
+        slideManagerToTest.deleteSlideById(slideId, moduleId);
+        Mockito.verify(slideRepo).delete((Slide) slideObj);
+        String slideIdAfterDeletion = sequencesList.get(1).getSlides().get(0).getId();
+
+        Assert.assertNotEquals(slideIdBeforeDeletion, slideIdAfterDeletion);
```
jdamerow, line 157: if the method changed the order of the objects, this would pass although the method has a bug. why don't you test if the the object is still in the list?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SlideManager.java
```diff
@@ -1,49 +1,107 @@
...
+                sequence.getSlides().removeAll(slideObjToRemove);
+                sequenceRepo.save(sequence);
+            }
+        }
+        try {
+            
+            slideRepo.delete((Slide) getSlide(slideId));
+            
+        } catch (IllegalArgumentException exception) {
+            logger.error("Unable to delete Slide. "+exception);
```
jdamerow, line 88: I believe `error` takes a string and the exception as second argument. I don't think it'll print the full stacktrace otherwise.

jdamerow:
Jenkins, test this please.

jdamerow:
after that should be good

jdamerow:
Jenkins, test this please.

jdamerow:
hurray!

PR #92, opened 2020-02-14
Story/vspc 52

Sagar-k93:
[VSPC 52] Fixed delete buttons on the Slides for image and text blocks.


vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ContentBlockManager.java
```diff
@@ -133,7 +135,48 @@ public CreationReturnValue createImageBlock(String slideId, byte[] image, String
...
+     */
+
     @Override
+    public void deleteTextBlockById(String id) throws BlockDoesNotExistException {
+        if (id == null) {
+            return;
+        }
+        try {
+            textBlockRepo.deleteById(id);         
+        } catch (EmptyResultDataAccessException e) {
```
jdamerow: Are you sure this exception is being thrown here?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ContentBlockManager.java
```diff
@@ -133,7 +135,48 @@ public CreationReturnValue createImageBlock(String slideId, byte[] image, String
...
+     */
+
+    @Override
+    public void deleteImageBlockById(String id) throws BlockDoesNotExistException {
+        if (id == null) {
+            return;
+        }
+        try {
+            imageBlockRepo.deleteById(id);
+        } catch (IllegalArgumentException | EmptyResultDataAccessException e) {
```
jdamerow: since id can't be null, IllegalArgumentException can't be thrown and for the other exception, I don't think it is thrown by this method?

vspace/src/main/webapp/WEB-INF/views/staff/modules/slides/slide.jsp
```diff
@@ -123,20 +125,23 @@ function uploadImage() {
            	var img = imageblock.find('img');
            	if(data != ''){
            		img.attr('id', data);
+           		var divBlock = document.getElementById("current");
```
jdamerow: please use jquery

vspace/src/main/webapp/WEB-INF/views/staff/modules/slides/slide.jsp
```diff
@@ -123,20 +125,23 @@ function uploadImage() {
...
+           		divBlock.querySelector("input[id='deleteImageId']").value = data;
+           		divBlock.setAttribute('id', data);
            	}
            	else{
            		img.attr('id', imgID);
+           		var divBlock = document.getElementById("current");
```
jdamerow: please use jquery

vspace/src/main/webapp/WEB-INF/views/staff/modules/slides/slide.jsp
```diff
@@ -359,7 +453,7 @@ $(document).ready(function() {
 $(window).on('load', function () {
 	var divWindow = $(".valueDiv");
 	var images = $(".imgDiv");
-	resizeImage(images);
+	/* resizeImage(images); */
```
jdamerow: remove if not needed

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ContentBlockManagerTest.java
```diff
@@ -0,0 +1,74 @@
...
+    public void test_deleteTextBlockById_success() throws BlockDoesNotExistException {
+        String textBlockId = "2";
+        managerToTest.deleteTextBlockById(textBlockId);
+        Mockito.verify(textBlockRepo).deleteById(textBlockId);
+    }
+
+    @Test(expected = BlockDoesNotExistException.class)
+    public void test_deleteTextBlockById_forNonExistentId() throws BlockDoesNotExistException {
+        String textBlockId = "notARealId";
+        Mockito.doThrow(EmptyResultDataAccessException.class).when(textBlockRepo).deleteById(textBlockId);
```
jdamerow: is that really thrown?

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ContentBlockManagerTest.java
```diff
@@ -0,0 +1,74 @@
...
+        managerToTest.deleteTextBlockById(textBlockId);
+    }
+
+    @Test
+    public void test_deleteTextBlockById_whenIdIsNull() throws BlockDoesNotExistException {
+        String textBlockId = "2";
```
jdamerow: textblockid should be null

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ContentBlockManagerTest.java
```diff
@@ -0,0 +1,74 @@
...
+    public void test_deleteImageBlockById_success() throws BlockDoesNotExistException {
+        String imageBlockId = "2";
+        managerToTest.deleteImageBlockById(imageBlockId);
+        Mockito.verify(imageBlockRepo).deleteById(imageBlockId);
+    }
+
+    @Test(expected = BlockDoesNotExistException.class)
+    public void test_deleteImageBlockById_forNonExistentId() throws BlockDoesNotExistException {
+        String imageBlockId = "notARealId";
+        Mockito.doThrow(EmptyResultDataAccessException.class).when(imageBlockRepo).deleteById(imageBlockId);
```
jdamerow: same as above

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ContentBlockManagerTest.java
```diff
@@ -0,0 +1,74 @@
...
+        managerToTest.deleteImageBlockById(imageBlockId);
+    }
+
+    @Test
+    public void test_deleteImagetBlockById_whenIdIsNull() throws BlockDoesNotExistException {
+        String imageBlockId = "2";
+        managerToTest.deleteImageBlockById(null);
+        Mockito.verify(imageBlockRepo, Mockito.never()).deleteById(imageBlockId);
```
jdamerow: same as above

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ContentBlockManager.java
Sagar-k93: If someone was to edit the id of the text block in the console of browser, then it would throw the error. It can also be thrown if somehow data gets deleted from db while the user has the page opened and then user tries to delete it through UI. Highly unlikely scenarios, but can happen. This code was already present in the branch from prior commits by Taylor I believe, so I did not make changes to these. Removing it will stop showing a generic message on the UI that something went wrong in case something does go wrong.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ContentBlockManagerTest.java
Sagar-k93: Updated.

vspace/src/main/webapp/WEB-INF/views/staff/modules/slides/slide.jsp
Sagar-k93: removed from code.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ContentBlockManagerTest.java
Sagar-k93: As mentioned above, it can happen that the block no longer exists but still present on UI.

vspace/src/main/webapp/WEB-INF/views/staff/modules/slides/slide.jsp
Sagar-k93: Updated to jquery.

vspace/src/main/webapp/WEB-INF/views/staff/modules/slides/slide.jsp
Sagar-k93: Updated to jquery.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ContentBlockManager.java
Sagar-k93: I removed IllegalArgumentException.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ContentBlockManager.java
Sagar-k93: Based on the comments above, please let me know if I should remove EmptyResultDataAccessException or not.

jdamerow: did you look into the javadoc? I don't think this exception is thrown by this method. (it's not a question of could it be thrown but does the method ever actually throw it based on its implementation).

jdamerow:
no need to comment on my comments, unless you have question or want to discuss something. i get an email for every comment, which isn't really helpful if it just says "done" ;)

PR #94, opened 2020-02-21
Story/vspc 58

prjadhav14:


vspace/src/main/java/edu/asu/diging/vspace/web/ExhibitionModuleController.java
```diff
@@ -1,23 +1,48 @@
...
+import edu.asu.diging.vspace.core.model.ISlide;
 import edu.asu.diging.vspace.core.services.IModuleManager;
+import edu.asu.diging.vspace.core.services.impl.ContentBlockManager;
 
 @Controller
 public class ExhibitionModuleController {
 
-    @Autowired
-    private IModuleManager moduleManager;
+	@Autowired
```
jdamerow: wrong indentation, please follow our style guide

vspace/src/main/java/edu/asu/diging/vspace/web/ExhibitionModuleController.java
```diff
@@ -1,23 +1,48 @@
...
+			List<ISlide> slides = module.getStartSequence().getSlides();
+			Iterator<ISlide> iter=slides.iterator();
+			while(iter.hasNext()) {
+				ISlide slide=iter.next();
+				slideContents.add(contentBlockManager.getAllContentBlocks(slide.getId()));
+			}
+			model.addAttribute("slideContents", slideContents);
+		}
+		catch(Exception e) {
+			e.printStackTrace();
```
jdamerow: nope, nope, nope, in so many ways. don't just catch all exceptions, catch specific ones. and why would an exception be thrown here? Handle that case appropriately. Show an error message or take otheer measures that will make sense from a user perspective. And this is not how you log an exception.

vspace/src/main/java/edu/asu/diging/vspace/web/ExhibitionModuleController.java
```diff
@@ -1,23 +1,48 @@
...
+		model.addAttribute("module", module);
+		try {
+			List<List<IContentBlock>> slideContents=new ArrayList<List<IContentBlock>>();
+			List<ISlide> slides = module.getStartSequence().getSlides();
+			Iterator<ISlide> iter=slides.iterator();
+			while(iter.hasNext()) {
+				ISlide slide=iter.next();
+				slideContents.add(contentBlockManager.getAllContentBlocks(slide.getId()));
+			}
+			model.addAttribute("slideContents", slideContents);
```
jdamerow: why do you add the contents of all slides here? You only need to show the contents of the first slide? Also, you might just want to redirect from this method to a controller that shows a slide (in this case the first in the sequence), so you can reuse it for all the following slides.

vspace/src/main/java/edu/asu/diging/vspace/web/ExhibitionModuleController.java
```diff
@@ -1,23 +1,66 @@
...
-        model.addAttribute("module", moduleManager.getModule(id));
-
+        IModule module = moduleManager.getModule(id);
+        model.addAttribute("module", module);
+        if (module.getStartSequence() == null) {
+            return "module";
+        } else {
+            model.addAttribute("startSequenceId", module.getStartSequence().getId());
+            model.addAttribute("slides", module.getStartSequence().getSlides());
+            model.addAttribute("slideContents", contentBlockManager.getAllContentBlocks(module.getStartSequence().getSlides().get(0).getId()));
```
jdamerow: this will throw an error when the start sequence doesn't have any slides

vspace/src/main/java/edu/asu/diging/vspace/web/ExhibitionModuleController.java
```diff
@@ -1,23 +1,66 @@
...
+            return "module";
```
jdamerow: in this case you just want to redirect to the controller that handles slides

vspace/src/main/java/edu/asu/diging/vspace/web/ExhibitionModuleController.java
```diff
@@ -1,23 +1,66 @@
...
+        }
+    }
+    @RequestMapping(value = "/exhibit/module/{moduleId}/sequence/{sequenceId}/slide/{slideId}/content", method = RequestMethod.GET)
+    public String slide(Model model, @PathVariable("slideId") String slideId,@PathVariable("moduleId") String moduleId) {
```
jdamerow: needs to be in its own controller

vspace/src/main/java/edu/asu/diging/vspace/web/ExhibitionModuleController.java
```diff
@@ -1,23 +1,66 @@
...
+        if (module.getStartSequence() == null) {
+            return "module";
+        } else {
+            model.addAttribute("startSequenceId", module.getStartSequence().getId());
+            model.addAttribute("slides", module.getStartSequence().getSlides());
+            model.addAttribute("slideContents", contentBlockManager.getAllContentBlocks(module.getStartSequence().getSlides().get(0).getId()));
+            return "module";
+        }
+    }
+    @RequestMapping(value = "/exhibit/module/{moduleId}/sequence/{sequenceId}/slide/{slideId}/content", method = RequestMethod.GET)
```
jdamerow: this does not need the `/content` suffix

vspace/src/main/java/edu/asu/diging/vspace/web/ExhibitionModuleController.java
```diff
@@ -1,23 +1,66 @@
...
+    public String slide(Model model, @PathVariable("slideId") String slideId,@PathVariable("moduleId") String moduleId) {
+        IModule module = moduleManager.getModule(moduleId);
+        model.addAttribute("module", module);
+        model.addAttribute("startSequenceId", module.getStartSequence().getId());
+        model.addAttribute("firstSlide", module.getStartSequence().getSlides().get(0).getId());
```
jdamerow: will throw exception if there are no slides in the sequence

vspace/src/main/java/edu/asu/diging/vspace/web/ExhibitionModuleController.java
```diff
@@ -1,23 +1,66 @@
...
+        model.addAttribute("currentSlide", slideId);
+        List<IContentBlock> slideContents = contentBlockManager.getAllContentBlocks(slideId);
+        model.addAttribute("slideContents", slideContents);
         return "module";
+
+    }
+    @RequestMapping(value = "/exhibit/module/{moduleId}/sequence/{id}/slides", method = RequestMethod.GET)
+    public ResponseEntity<List<ISlide>> getSequenceSlides(Model model, @PathVariable("moduleId") String moduleId,
+            @PathVariable("id") String sequenceId) {
+        List<ISlide> slides = sequenceManager.getSequence(sequenceId).getSlides();
```
jdamerow: what if the sequence with that id doesn't exist?

vspace/src/main/webapp/WEB-INF/views/exhibition/module.jsp
```diff
@@ -1,12 +1,108 @@
...
+				if (images[i].width > divWindow.width()) {
+					images[i].width = 800;
+				} else {
+					$(".valueDiv").css("width", images[i].width);
+				}
+			}
+		}
+	});
+	$(document).ready(function($) {
+		$.ajax({
```
jdamerow: why do you do an ajax call here instead of rendering it in the template?

vspace/src/main/java/edu/asu/diging/vspace/web/ExhibitionSequencesController.java
```diff
@@ -0,0 +1,48 @@
...
+	private IModuleManager moduleManager;
+
+	@Autowired
+	private ISequenceManager sequenceManager;
+
+	@RequestMapping(value = "/exhibit/module/{moduleId}/sequence/{sequenceId}")
+	public String sequence(Model model, @PathVariable("sequenceId") String sequenceId, @PathVariable("moduleId") String moduleId) {
+		IModule module = moduleManager.getModule(moduleId);
+		model.addAttribute("module", module);
+		List<ISequence> list=moduleManager.getModuleSequences(moduleId);
```
jdamerow: since this variable contains sequences, why is called `list`?

vspace/src/main/java/edu/asu/diging/vspace/web/ExhibitionSequencesController.java
```diff
@@ -0,0 +1,48 @@
...
+		boolean flag=true;
```
jdamerow: bad varialbe name

vspace/src/main/java/edu/asu/diging/vspace/web/ExhibitionSequencesController.java
```diff
@@ -0,0 +1,48 @@
...
+		for(ISequence sequence:list) {
+			if(sequence.getId().equals(sequenceId)) {
+				flag=false;
+				break;
+			}
+		}
```
jdamerow: Instead I would try to use `stream` and `filter` (filter by id). It's a lot more elegant and would probably be just a one-liner.

vspace/src/main/java/edu/asu/diging/vspace/web/ExhibitionSequencesController.java
```diff
@@ -0,0 +1,48 @@
...
+		if(flag) {
+			model.addAttribute("error","Sequence does not belong to selected module.");
+			return "module";
+		}
+		else if(sequenceManager.getSequence(sequenceId).getSlides().size()==0) {
+			model.addAttribute("error","No slides to display in selected sequence for module.");
+			return "module";
+		}
+		String firstSlideId=sequenceManager.getSequence(sequenceId).getSlides().get(0).getId();
```
jdamerow: put `sequenceManager.getSequence(sequenceId).getSlides()` into a variable so you don't have to call it twice.

vspace/src/main/java/edu/asu/diging/vspace/web/ExhibitionSlideController.java
```diff
@@ -0,0 +1,73 @@
...
+	@Autowired
+	private ISequenceManager sequenceManager;
+
+	@RequestMapping(value = "/exhibit/module/{moduleId}/sequence/{sequenceId}/slide/{slideId}", method = RequestMethod.GET)
+	public String slide(Model model, @PathVariable("slideId") String slideId,@PathVariable("moduleId") String moduleId, @PathVariable("sequenceId") String sequenceId) {
+		IModule module = moduleManager.getModule(moduleId);
+		model.addAttribute("module", module);
+		model.addAttribute("startSequenceId", module.getStartSequence().getId());
+		model.addAttribute("firstSlide", module.getStartSequence().getSlides().get(0).getId());
+		model.addAttribute("slides", sequenceManager.getSequence(sequenceId).getSlides());
```
jdamerow: what if either of those don't exist (e.g. no start sequence, no slides in sequence or no sequence)?

vspace/src/main/java/edu/asu/diging/vspace/web/ExhibitionSlideController.java
```diff
@@ -0,0 +1,73 @@
...
+		model.addAttribute("currentSequenceId", sequenceId);
+		model.addAttribute("currentSlide", slideId);
+		model.addAttribute("currentSlideCon", sildeManager.getSlide(slideId));
```
jdamerow: why do you put the slide id in the model, if the slide is in there already?

vspace/src/main/java/edu/asu/diging/vspace/web/ExhibitionSlideController.java
```diff
@@ -0,0 +1,73 @@
...
+		model.addAttribute("currentSlide", slideId);
+		model.addAttribute("nextSlide", nextSlideId);
+		model.addAttribute("prevSlide", prevSlideId);
+		model.addAttribute("currentSlideCon", sildeManager.getSlide(slideId));
+		model.addAttribute("startSequenceId", module.getStartSequence().getId());
+		model.addAttribute("numOfSlides", slides.size());
+		model.addAttribute("currentNumOfSlide", index+1);
+		return "slideShow";
+
+	}
```
jdamerow: why are there two methods that bot seem to show a slide?

vspace/src/main/webapp/WEB-INF/views/exhibition/module.jsp
```diff
@@ -1,12 +1,131 @@
...
+<script
+	src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.14.7/umd/popper.min.js"
+	integrity="sha384-UO2eT0CpHqdSJQ6hJty5KVphtPhzWj9WO1clHTMGa3JDZwrnQq4sF86dIHNDz0W1"
+	crossorigin="anonymous">
+	
+</script>
+<script
+	src="https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/js/bootstrap.min.js"
+	integrity="sha384-JjSmVgyd0p3pXB1rRibZUAYoIIy6OrQ6VrjIEaFf/nJGzIxFDsf4x0xIM+B07jRM"
+	crossorigin="anonymous">
```
jdamerow: isn't bootstrap already loaded?

vspace/src/main/webapp/WEB-INF/views/exhibition/module.jsp
```diff
@@ -1,12 +1,131 @@
...
+
+<link id="applicationStylesheet" rel="stylesheet"
+	href="<c:url value="/resources/extra/Module_1.css" />">
+<script id="applicationScript"
+	src="<c:url value="/resources/extra/Module_1.js" />"></script>
+
+<link rel="stylesheet" type="text/css" id="applicationStylesheet"
+	href="<c:url value="/resources/extra/Module_1___slideshow___1.css" />">
+<script id="applicationScript"
+	src="<c:url value="/resources/extra/Module_1___slideshow___1.js" />"></script>
```
jdamerow: what exactly do these files contain? do we need all that?

vspace/src/main/java/edu/asu/diging/vspace/web/ExhibitionSlideController.java
prjadhav14: One method is to load the home page of modules, just the first slide and the second method is to start slideShow as managing content on one jsp and handling visibilities is getting difficult.

jdamerow: I'm confused, what is the slideshow in this case?

prjadhav14: there is a button called text version, on click of which the slideShow starts wherein a user can switch between all the slides in the sequence and on the same page user can come back to the module screen which has only the first slide of the start sequence.

vspace/src/main/java/edu/asu/diging/vspace/web/ExhibitionModuleController.java
jdamerow: ?

vspace/src/main/webapp/WEB-INF/views/exhibition/module.jsp
```diff
@@ -1,12 +1,169 @@
...
-<div class="container">
-<div class="row">
-<div class="col-sm">
-<h1>${module.name}</h1>	
+<link
+	href="<c:url value="/resources/bootstrap/css/bootstrap.min.css" />"
+	rel="stylesheet">
+<script
+	src="https://ajax.googleapis.com/ajax/libs/jquery/3.2.0/jquery.min.js"></script>
+<script
```
jdamerow: aren't bootstrap and jquery already included?

jdamerow:
please resolve conflicts

prjadhav14:
Resolved the conflicts for tiles-def.xml

vspace/src/main/java/edu/asu/diging/vspace/web/ExhibitionSequencesController.java
```diff
@@ -0,0 +1,44 @@
...
+    @Autowired
+    private ISequenceManager sequenceManager;
+
+    @RequestMapping(value = "/exhibit/module/{moduleId}/sequence/{sequenceId}")
+    public String sequence(Model model, @PathVariable("sequenceId") String sequenceId,
+            @PathVariable("moduleId") String moduleId) {
+        IModule module = moduleManager.getModule(moduleId);
+        model.addAttribute("module", module);
+        List<ISequence> sequences = moduleManager.getModuleSequences(moduleId);
+        List<ISlide> slides = sequenceManager.getSequence(sequenceId).getSlides();
```
jdamerow: what if sequence doesn't exist? this will throw an exception, won't it?

vspace/src/main/java/edu/asu/diging/vspace/web/ExhibitionSlideController.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @Autowired
+    private ISequenceManager sequenceManager;
+
+    @RequestMapping(value = "/exhibit/module/{moduleId}/sequence/{sequenceId}/slide/{slideId}", method = RequestMethod.GET)
+    public String slide(Model model, @PathVariable("slideId") String slideId, @PathVariable("moduleId") String moduleId,
+            @PathVariable("sequenceId") String sequenceId) {
+        IModule module = moduleManager.getModule(moduleId);
+        model.addAttribute("module", module);
+        if (module.getStartSequence() == null) {
```
jdamerow: what if module doesn't exist?

vspace/src/main/java/edu/asu/diging/vspace/web/ExhibitionSlideController.java
```diff
@@ -0,0 +1,66 @@
...
+            model.addAttribute("error", "Sorry, this module has not been configured yet.");
+        } else {
+            String startSequenceId = module.getStartSequence().getId();
+            model.addAttribute("startSequenceId", startSequenceId);
+            List<ISlide> sequenceSlides = sequenceManager.getSequence(sequenceId).getSlides();
```
jdamerow: what if the sequence doesn't exist?

jdamerow:
Jenkins, test this please.

vspace/src/main/webapp/WEB-INF/views/exhibition/module.jsp
```diff
@@ -1,12 +1,165 @@
...
+												class="Icon_ionic_ios_arrow_back_Class"
+												d="M 17.28610038757324 23.69940376281738 L 30.53693389892578 10.45899486541748 C 31.51693534851074 9.478996276855469 31.51693534851074 7.894316673278809 30.53693389892578 6.92474365234375 C 29.55693244934082 5.944746017456055 27.97225761413574 5.95517110824585 26.99225997924805 6.92474365234375 L 11.9795093536377 21.92706680297852 C 11.03078651428223 22.87578582763672 11.00993728637695 24.39791297912598 11.90652942657471 25.3779125213623 L 26.98183441162109 40.48448944091797 C 27.47183418273926 40.97449111938477 28.11821746826172 41.21427917480469 28.75417137145996 41.21427917480469 C 29.39012908935547 41.21427917480469 30.0365104675293 40.97449111938477 30.52651405334473 40.48448944091797 C 31.50650978088379 39.50448989868164 31.50650978088379 37.91981506347656 30.52651405334473 36.95023727416992 L 17.28610038757324 23.69940376281738 Z">
+									</path>
+								</svg>
+									</div>
+								</a>
+							</c:if>
+							<c:if test="${nextSlide !=  ''}">
+								<a
+									href="/vspace/exhibit/module/${module.id}/sequence/${currentSequenceId}/slide/${nextSlide}">
```
jdamerow: this is not how you create links (this will raise a not found (see my other ocmments)).

jdamerow:
and a couple of other thigns:
- the menu bar on the left should be collapsed by default
- the next/prev buttons should be smaller and on the top (above the branching point choice links)
- something is wrong with the top bar:
image

- the module name should always be visible (maybe in the grey top bar that also has the left menu link?)
- the text version link shouldn't be there. we can't create branching points yet, so there shouldn't be any branching point choices on the public page yet.
- clicking on the next button, shows a not found error.
- there is no link to get back to the space the user came from.
- the font is different than from the rest of the exhibition, which would be fine if that's the new font for all the public pages, but I don't like it very much... ;) please choose a different one or use the old one.
- the light blue of the left menu icon doesn't really work with the red.
- let's scroll only the content of the slide and keep everything else (so that the next and prev link are always visible)

vspace/src/main/java/edu/asu/diging/vspace/web/ExhibitionSequencesController.java
prjadhav14: In this case, control will end in ExhibitionModuleController and the user will be shown a message saying "This module is not yet configured".

jdamerow: someone can still bookmark for example a url that goes directly to a sequence and then this will throw an exception.

vspace/src/main/java/edu/asu/diging/vspace/web/ExhibitionModuleController.java
```diff
@@ -6,18 +6,38 @@
...
+        ISpace space = spaceManager.getSpace(spaceId);
+        if (space == null) {
+            throw new SpaceNotFoundException(spaceId);
+        }
+        IModule module = moduleManager.getModule(id);
+        model.addAttribute("module", module);
+        if (module == null) {
+            throw new ModuleNotFoundException(id);
+        } else if (module.getStartSequence() == null) {
+            throw new ModuleNotConfiguredException(id);
```
jdamerow, line 38: that shouldn't be a exception; it's not such an unexpected situation.

vspace/src/main/java/edu/asu/diging/vspace/web/ExhibitionSequencesController.java
```diff
@@ -0,0 +1,65 @@
...
+        IModule module = moduleManager.getModule(moduleId);
+        if (module == null) {
+            throw new ModuleNotFoundException(moduleId);
+        }
+        model.addAttribute("module", module);
+        if (module.getStartSequence() == null) {
+            throw new ModuleNotConfiguredException(moduleId);
+        }
+        List<ISequence> sequences = moduleManager.getModuleSequences(moduleId);
+        boolean sequenceExist = sequences.stream().anyMatch(sequence -> sequence.getId().equals(sequenceId));
```
jdamerow, line 54: thinking about this, this is probably more efficiently done as a repo method.

vspace/src/main/java/edu/asu/diging/vspace/web/ExhibitionSlideController.java
```diff
@@ -0,0 +1,110 @@
...
+            throw new ModuleNotFoundException(moduleId);
+        }
+        if (module.getStartSequence() == null) {
+            throw new ModuleNotConfiguredException(moduleId);
+        }
+        String startSequenceId = module.getStartSequence().getId();
+        model.addAttribute("startSequenceId", startSequenceId);
+
+        List<ISequence> sequences = moduleManager.getModuleSequences(moduleId);
+        boolean sequenceExist = sequences.stream().anyMatch(sequence -> sequence.getId().equals(sequenceId));
```
jdamerow, line 66: see above

vspace/src/main/java/edu/asu/diging/vspace/web/exception/GlobalExceptionHandler.java
```diff
@@ -0,0 +1,79 @@
...
+import javax.servlet.http.HttpServletRequest;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.web.bind.annotation.ControllerAdvice;
+import org.springframework.web.bind.annotation.ExceptionHandler;
+import org.springframework.web.servlet.ModelAndView;
+
+@ControllerAdvice
+public class GlobalExceptionHandler {
```
jdamerow, line 12: I wouldn't call it global, since it's not globally handling exceptions. Just for the browser part.

vspace/src/main/java/edu/asu/diging/vspace/web/exception/GlobalExceptionHandler.java
```diff
@@ -0,0 +1,79 @@
...
+            SlidesInSequenceNotFoundException ex) {
+        ModelAndView modelAndView = new ModelAndView();
+        modelAndView.addObject("error", ex.getMessage());
+        logger.info("SpaceNotFoundException Occured:: URL=" + request.getRequestURL());
+        modelAndView.addObject("url", request.getRequestURL());
+        modelAndView.setViewName("module");
+        return modelAndView;
+    }
+
+}
```
jdamerow, line 79: you're swallowing the exceptions; they need to be printed in the logs.

vspace/src/main/java/edu/asu/diging/vspace/web/exception/ModuleNotConfiguredException.java
```diff
@@ -0,0 +1,14 @@
+package edu.asu.diging.vspace.web.exception;
+
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+@ResponseStatus(reason = "Module Not Configured")
+public class ModuleNotConfiguredException extends Exception {
+    private static final long serialVersionUID = -3332292346834265371L;
```
jdamerow, line 8: let's start with 1

jdamerow: same for all the others

vspace/src/main/java/edu/asu/diging/vspace/web/exception/GlobalExceptionHandler.java
```diff
@@ -0,0 +1,79 @@
...
+        modelAndView.addObject("url", request.getRequestURL());
+        modelAndView.setViewName("module");
+        return modelAndView;
+    }
+
+    @ExceptionHandler({ SpaceNotFoundException.class })
+    protected ModelAndView handleSpaceNotFoundException(HttpServletRequest request,
+            SlidesInSequenceNotFoundException ex) {
+        ModelAndView modelAndView = new ModelAndView();
+        modelAndView.addObject("error", ex.getMessage());
```
jdamerow, line 72: since these will be internationalized eventually, I wouldn't use the message from the exception here.

jdamerow:
Jenkins, test this please.

jdamerow:
- The left menu should be collapsed by default
- there is something wrong with multiple images
image

- error pages should be proper error pages, let @adhvikmadhav help you with that (maybe including some graphics)
- the next and prev button shouldn't be there if there is no next/prev
- back to sapce menu should be a link
- menu has still issues with the icon aligment
image

jdamerow:
Jenkins, test this please.

jdamerow:
Nothing seemed to have changed. The images are still layered above each other, and it doesn't look like the mockup Adhvik made. Please check with Adhvik or the mockup again.

jdamerow:
Jenkins, test this please.

jdamerow:
Jenkins, test this please.

jdamerow:
why is there still a top menu bar? also, the menu on the left is miss aligned (why are the links stacked to the right???). font is way to big for the layout and what's with that misaligned bullet point? also, there should be return link to the space that is not in the menu but an icon somewhere where it makes sense.

jdamerow:
also, the side menu should only have sequences and slides, no space. link to space should be somewhere else.

prjadhav14:
I have kept language picker in file to test only, I will comment that section of code later.

jdamerow:
Jenkins, test this please.

jdamerow:
Please get rid of the top menu bar. Make a login icon in the top right corner. Also, the exhibition name has a different font than the rest. Also, let's get rid of the side menu. That doesn't work where it is right now. We'll think about that in another ticket.

@adhvikmadhav can you work the exhibition title into the layout please?

jdamerow:
Jenkins, test this please.

jdamerow:
Just two more things and I think it should be good:
- images are blown up when their are too small:
image
- the footer needs to stay at the bottom of the page. it's fine if it's always there, but right now it moves up if the content is short which just looks weird
image

prjadhav14:
> Just two more things and I think it should be good:
> 
> * images are blown up when their are too small:
>   image
> * the footer needs to stay at the bottom of the page. it's fine if it's always there, but right now it moves up if the content is short which just looks weird
>   image

Images real height and width is not getting stored when we add image to the slide hence, image blown up issue needs to be fixed from slide creation perspective, after which the height and width parameters can be used to show image in actual size in which it was uploaded to the server.

jdamerow:
Have you tried not providing a width and height for the image? or even better provide max-width and height? the browser should take care of the image size by itself.

prjadhav14:
> Have you tried not providing a width and height for the image? or even better provide max-width and height? the browser should take care of the image size by itself.

Yes, I tried that. As per my observation it works fro smaller size images but if the image size is more than the block provided to display it, image crosses the boundaries, and overlaps on other content on the screen. I will try again and provide an update.

prjadhav14:
> > Have you tried not providing a width and height for the image? or even better provide max-width and height? the browser should take care of the image size by itself.
> 
> Yes, I tried that. As per my observation it works fro smaller size images but if the image size is more than the block provided to display it, image crosses the boundaries, and overlaps on other content on the screen. I will try again and provide an update.

The fix worked with some additional CSS changes. :)

jdamerow:
Looks good, just one last thing, the slide title needs to be added to the slides.

jdamerow:
Jenkins, test this please.

jdamerow:
Jenkins, test this please.

jdamerow:
Hurray! Good work @prjadhav14 and @adhvikmadhav!

PR #97, opened 2020-02-28
VSPC-67

Sagar-k93:
Added back button on slides page to send user back to modules page.


jdamerow:
Jenkins, test this please.

jdamerow:
lets also put that button somewhere on the top so you don't have to scroll down

PR #107, opened 2020-03-20
Story/vspc 71

Sagar-k93:


vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
```diff
@@ -42,4 +50,22 @@ public void setExhibition(JoinPoint jp) {
             }
         }
     }
+    
+    @Around("execution(public * edu.asu.diging.vspace.web..ExhibitionSpaceController.*(..))")
```
jdamerow, line 54: you want this to happen for all controllers for the public side, don't you? not just the exhibitionspacecontroller.

vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
```diff
@@ -42,4 +50,22 @@ public void setExhibition(JoinPoint jp) {
...
+    public Object showExhibition(ProceedingJoinPoint jp) throws Throwable {
+        Object[] args = jp.getArgs();
+        Exhibition exhibition = (Exhibition) exhibitionManager.getStartExhibition();
+        AuthenticationFacade authFacade = new AuthenticationFacade();
```
jdamerow, line 58: this is a spring managed class, you should not instantiate it.

vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
```diff
@@ -42,4 +50,22 @@ public void setExhibition(JoinPoint jp) {
...
+        if (exhibition.getMode().equals(ExhibitionModes.ACTIVE.getValue())) {
+            return jp.proceed();
+        }
+        else if(!(authFacade.getAuthenticatedUser() == null)){
+           ((Model) args[1]).addAttribute("showModal","true");
```
jdamerow, line 63: the modal might not always be the second argument, this will throw an exception if it is not.

vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
```diff
@@ -42,4 +50,22 @@ public void setExhibition(JoinPoint jp) {
...
+            return jp.proceed();
+        }
+        else {  
+            jp.proceed();
```
jdamerow, line 67: why do you proceed here if you return a specific page?

vspace/src/main/java/edu/asu/diging/vspace/core/model/ExhibitionModes.java
```diff
@@ -0,0 +1,30 @@
...
+	
+	public static List<String> getAllValues() {
+		List<String> valuesList = new ArrayList<>();
+		for(ExhibitionModes val : ExhibitionModes.values()) {
+			valuesList.add(val.getValue());
+		}
+		
+		return valuesList;
+	}
+}
```
jdamerow, line 30: Please follow our styleguide. I will look at this PR again when that's done.

vspace/src/main/webapp/WEB-INF/views/exhibition/space.jsp
```diff
@@ -123,7 +137,22 @@ function drawLinks() {
...
-
+<div id="exhibitionDownModal" class="modal" tabindex="-1"
+	role="dialog">
+	<div class="modal-dialog" role="document">
+		<div class="modal-content">
+			<div class="modal-header">
+				<h5 class="modal-title">Exhibition Not Active</h5>
+			</div>
+			<div class="modal-body">
+				<h6>${exhibitionConfig.modeMessage}</h6>
```
jdamerow, line 148: same as above

vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
```diff
@@ -22,24 +29,46 @@
...
+        Exhibition exhibition = (Exhibition) exhibitionManager.getStartExhibition();
+        if(exhibition.getMode().equals(ExhibitionModes.ACTIVE.getValue())) {
+            return jp.proceed();
+        }
+        else if(!(authFacade.getAuthenticatedUser()==null) && index>-1) {
+            ((Model) args[index]).addAttribute("showModal", "true");
+            return jp.proceed();
+        }
+        else {
+            ((Model) args[index]).addAttribute("modeMessage", exhibition.getModeMessage());
```
jdamerow, line 70: see comment in jsp

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/Exhibition.java
```diff
@@ -29,6 +29,10 @@
     private ISpace startSpace;
 
     private String title;
+    
+    private String mode;
+    
+    private String modeMessage;
```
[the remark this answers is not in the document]
Sagar-k93: For Offline mode, we allow the user to set an explicit message. How would we store that in enum then?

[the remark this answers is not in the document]
jdamerow: that can be an additional field, but you should still store the enum.

vspace/src/main/java/edu/asu/diging/vspace/core/model/ExhibitionModes.java
```diff
@@ -0,0 +1,29 @@
...
+
+    public String getValue() {
+        return value;
+    }
+
+    public static List<ExhibitionModes> getAllModes (){
+        List<ExhibitionModes> modesList = new ArrayList<>();
+            modesList.add(ExhibitionModes.ACTIVE);
+            modesList.add(ExhibitionModes.MAINTENANCE);
+            modesList.add(ExhibitionModes.OFFLINE);
```
jdamerow, line 26: what's up with the indentation?

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/Exhibition.java
```diff
@@ -29,6 +30,10 @@
     private ISpace startSpace;
 
     private String title;
+    
+    private ExhibitionModes mode;
```
jdamerow, line 34: you want to make sure the enum is mapped as a string, so we don't have to worry about changing the order of the values.
`@Enumerated(EnumType.STRING)`

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ExhibitionConfigurationController.java
```diff
@@ -60,26 +61,33 @@ public String showExhibitions(Model model) {
...
         }
-
         exhibition.setStartSpace(startSpace);
         exhibition.setTitle(title);
+        exhibition.setMode(exhibitMode);
+        if(exhibitMode.equals(ExhibitionModes.OFFLINE) && !(customMessage.equals("")
+            || customMessage.equals(ExhibitionModes.OFFLINE.getValue()))) {
+            exhibition.setCustomMessage(customMessage);
+        } else {
+            exhibition.setCustomMessage("");
```
jdamerow, line 85: why don't you leave that with the default (`null`) in this case?

Sagar-k93: I wanted to reset the message in case the exhibition mode is changed to Active / maintenance from Offline. Should I store the old message even after the offline mode is removed ? In that case, it will show up in the config page while setting the exhibition to Offline at a later stage.

jdamerow: I would keep it. Chances are that the maintenance message won't change much (once the user settled on one they would probably reuse it for the next time).

jdamerow:
Jenkins, test this please.

jdamerow:
throws a nullpointer
```
java.lang.NullPointerException
	edu.asu.diging.vspace.core.aspects.ExhibitionDataAspect.showExhibition(ExhibitionDataAspect.java:62)
	sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	java.lang.reflect.Method.invoke(Method.java:498)
```

jdamerow:
Jenkins, test this please.

vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
```diff
@@ -22,24 +31,64 @@
...
+        Object[] args = jp.getArgs();
+        MethodSignature signature = (MethodSignature) jp.getSignature();
+        int indexofModel = (Arrays.asList(signature.getParameterTypes())).indexOf(Model.class);
+        int indexofSpaceId = (Arrays.asList(signature.getParameterTypes())).indexOf(String.class);
+        Exhibition exhibition = (Exhibition) exhibitionManager.getStartExhibition();
+        ISpace space = spaceManager.getSpace(args[indexofSpaceId].toString());
+        if(exhibition==null) {
+            return "home";
+        }
+        if(exhibition.getMode().equals(ExhibitionModes.ACTIVE)) {
```
jdamerow, line 72: this line throws a null pointer when there is no mode yet (which is true for all existing exhibitions)

vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
```diff
@@ -22,24 +31,64 @@
...
             }
         }
     }
+
+    @Around("execution(public * edu.asu.diging.vspace.web.publicview..*Controller.*(..))")
+    public Object showExhibition(ProceedingJoinPoint jp) throws Throwable {
+        Object[] args = jp.getArgs();
+        MethodSignature signature = (MethodSignature) jp.getSignature();
+        int indexofModel = (Arrays.asList(signature.getParameterTypes())).indexOf(Model.class);
+        int indexofSpaceId = (Arrays.asList(signature.getParameterTypes())).indexOf(String.class);
```
jdamerow, line 66: I see two issues with this:
- this method is called for any method of controllers in the publicview package, right? so, controllers that handle the module pages will be intercepted too, but this line will return a module id not a space id
- often controller methods have several String parameters, unless the space id is the first one, this will fail. I would check the parameter name to make sure it's 'id' or 'spaceId'.

jdamerow:
please resolve conflicts

vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
```diff
@@ -29,17 +54,86 @@ public void setExhibition(JoinPoint jp) {
...
                 }
+            }  
+        }
+    }
+
+    @Around("execution(public * edu.asu.diging.vspace.web.publicview..*Controller.*(..))")
+    public Object showExhibition(ProceedingJoinPoint jp) throws Throwable {
+        Object[] args = jp.getArgs();
+        MethodSignature signature = (MethodSignature) jp.getSignature();
+        int indexofModel = (Arrays.asList(signature.getParameterTypes())).indexOf(Model.class);
```
jdamerow, line 75: needs to b proper camel case

vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
```diff
@@ -29,17 +54,86 @@ public void setExhibition(JoinPoint jp) {
...
+        Exhibition exhibition = (Exhibition) exhibitionManager.getStartExhibition(); 
+        if(exhibition==null) {
+            return "home";
```
jdamerow, line 78: that seems wrong, you probably want the actual controller take over from here so that the correct message is shown.

vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
```diff
@@ -29,17 +54,86 @@ public void setExhibition(JoinPoint jp) {
...
+        }
+        if(exhibition.getMode() == null) {
+            return jp.proceed();
+        }
+        Map<String, String> ids = getIds(args, signature);
+        String spaceId = ids.getOrDefault("spaceId", "");
```
jdamerow, line 84: strings should not be hard coded. why don't you use the id prefix enum here?

vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
```diff
@@ -29,17 +54,86 @@ public void setExhibition(JoinPoint jp) {
...
+    }
+    
+    public Object redirectRequest(ProceedingJoinPoint jp, String spaceId, String moduleId, int modelIndex, Exhibition exhibition) throws Throwable{
+        ISpace space = spaceManager.getSpace(spaceId);
+        IModule module = moduleManager.getModule(moduleId);
+        Object[] args = jp.getArgs();
+        if(exhibition.getMode().equals(ExhibitionModes.ACTIVE)) {
+            if(space==null && module==null) {
+                return "notFound";
+            }
```
jdamerow, line 100: how is this different from the szenario that is checked before this method is called (lines 86-89)?

vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
```diff
@@ -29,17 +54,86 @@ public void setExhibition(JoinPoint jp) {
...
+            else {
+                return jp.proceed();
+            }
+        }
+        if(authFacade.getAuthenticatedUser()!=null && modelIndex>-1) {
+            if(space==null && module==null) {
```
jdamerow, line 106: why do you check the modelIndex variable?

jdamerow: also, how is that different from the previous times not found was returned?

vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
```diff
@@ -29,17 +54,86 @@ public void setExhibition(JoinPoint jp) {
...
+                return "notFound";
+            }
+            else {
+                ((Model) args[modelIndex]).addAttribute("showModal", "true");
+                return jp.proceed();
+            }
+        }
+        if(exhibition.getMode()==ExhibitionModes.OFFLINE && !exhibition.getCustomMessage().equals("")) {
+            ((Model) args[modelIndex]).addAttribute("modeValue", exhibition.getCustomMessage());
+        } else {
```
jdamerow, line 116: so this else block will always be executed when either the mode is not offline or there is no custom message? that seems wrong. also, it would be much cleaner to have a tertiary expression set the custom message and then add the attribute once.

vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
```diff
@@ -29,17 +54,86 @@ public void setExhibition(JoinPoint jp) {
...
+    public Map<String, String> getIds(Object[] args, MethodSignature signature) {
+        Map<String, String> res = new HashMap<>();
+        for(int i=0; i <signature.getParameterTypes().length; ++i) {
+            if(signature.getParameterTypes()[i].equals(String.class)) {
+                if(((String) args[i]).length() > 2) {
+                    String prefix = ((String) args[i]).substring(0,3);
+                    if(prefix.equalsIgnoreCase(IdPrefix.SPACEID.getValue())) {
+                        res.put("spaceId", (String) args[i]);
+                    }
+                    else if(prefix.equalsIgnoreCase(IdPrefix.MODULEID.getValue())) {
```
jdamerow, line 131: code style

vspace/src/main/java/edu/asu/diging/vspace/core/model/ExhibitionModes.java
```diff
@@ -0,0 +1,29 @@
...
+    public String getValue() {
+        return value;
+    }
+
+    public static List<ExhibitionModes> getAllModes (){
+        List<ExhibitionModes> modesList = new ArrayList<>();
+        modesList.add(ExhibitionModes.ACTIVE);
+        modesList.add(ExhibitionModes.MAINTENANCE);
+        modesList.add(ExhibitionModes.OFFLINE);
+        return modesList;
```
jdamerow, line 27: i'm pretty sure enums come with a method to get all values, don't they?

vspace/src/main/java/edu/asu/diging/vspace/core/model/IdPrefix.java
```diff
@@ -0,0 +1,17 @@
+package edu.asu.diging.vspace.core.model;
+
+public enum IdPrefix {
```
jdamerow, line 3: these should also be used in the ID annotation of the domain classes, to avoid changing one but not the other.

vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
Sagar-k93: If (exhibition == null) returns to controller, then controller loads slides even if no exhibition exists. Home shows that there is no exhibition at the moment.

vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
Sagar-k93: This checks if the user is not logged in and if logged in and incorrect space/module Ids, then we show notFound page. If Id is correct we show the exhibition but with a flag to show the pop-up message. Since Active is handled before this if-else, only offline or maintenance modes will apply here.

vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
Sagar-k93: The else block will only handle maintenance as Active is handled before and Offline is handled in if. I will implement tertiary statement for offline.

vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
jdamerow: It should redirect to the start of the exhibition in that case. If we change something on the ’not setup yet’ screen, you don’t want to redo all of that in the aspect.

vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
jdamerow: since it doesn't matter if the user is logged in or not for the case that neither the space or the module exists, you should just have one check and return null.

vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
```diff
@@ -29,17 +54,86 @@ public void setExhibition(JoinPoint jp) {
...
+    public Object redirectRequest(ProceedingJoinPoint jp, String spaceId, String moduleId, int modelIndex, Exhibition exhibition) throws Throwable{
+        ISpace space = spaceManager.getSpace(spaceId);
+        IModule module = moduleManager.getModule(moduleId);
+        Object[] args = jp.getArgs();
+        if(exhibition.getMode().equals(ExhibitionModes.ACTIVE)) {
+            if(space==null && module==null) {
+                return "notFound";
+            }
+            else {
+                return jp.proceed();
```
jdamerow, line 102: so if space is null but module is not null, we proceed? that doesn't sound right.

vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
jdamerow: I don't think there is a reason to return 'notFound' more than once. You should be able to write the code so that it's very clear under which conditions it return not found, when it simply proceeds, and when it shows an alert.

Sagar-k93: For Offline or Maintenance, it matters if the user is logged in or not, If they are and the module/space does not exist, we show does not exist page, if they are not logged in and space/module does not exist we still show offline / maintenance page.

jdamerow: but if the exhibition is active, it doesn't matter.

jdamerow: now I see, I think you just need to clean up your code and add some comments. One if statement that returns notfound (e.g. is active or is logged in and space == null or module == nul)

vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
```diff
@@ -48,7 +63,78 @@ public void setExhibition(JoinPoint jp) {
...
+    
+    // Method to get the space or module id's from the request body.
+    public Map<String, String> getIds(Object[] args, MethodSignature signature) { 
+        Map<String, String> res = new HashMap<>();
+        for(int i=0; i <signature.getParameterTypes().length; ++i) {
+            if(signature.getParameterTypes()[i].equals(String.class)) {
+                if(((String) args[i]).length() > 2) {
+                    String prefix = ((String) args[i]).substring(0,3);
+                    if(prefix.equalsIgnoreCase(IdPrefix.SPACEID.getValue())) {
+                        res.put(IdPrefix.SPACEID.getValue(), (String) args[i]);
```
jdamerow, line 130: I would use the enum as key to make it easier to read.

vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
```diff
@@ -48,7 +63,78 @@ public void setExhibition(JoinPoint jp) {
...
+    }
+    
+    // Based on exhibition mode, get the redirect page or pass control to controller.  
+    public Object redirectRequest(ProceedingJoinPoint jp, String spaceId, String moduleId, int modelIndex, Exhibition exhibition) throws Throwable{
+        ISpace space = spaceManager.getSpace(spaceId);
+        IModule module = moduleManager.getModule(moduleId);
+        Object[] args = jp.getArgs();
+        // For exhibition being active, if space or module is not found, show not found page or else give control to controller.
+        if(exhibition.getMode().equals(ExhibitionModes.ACTIVE)) {
+            if(space==null && module==null) {
```
jdamerow, line 97: your comment says if space **or** module is not found but here you use `&&`. **Or** seems right.

vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
```diff
@@ -48,7 +63,78 @@ public void setExhibition(JoinPoint jp) {
...
+                return "notFound";
+            }
+            return jp.proceed();
+        }
+        // If the user is logged in, if space or module invalid, show not found page or else add attribute to show pop up
+        // notifying user about status of exhibition. ModelIndex checks if there is a model to add attribute to.
+        if(authFacade.getAuthenticatedUser()!=null && modelIndex>-1) {
```
jdamerow, line 104: this should be part of the condition in line 96 (exhiibtion is active or user is authenticated and there is no model).

vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
```diff
@@ -48,7 +63,78 @@ public void setExhibition(JoinPoint jp) {
...
+        }
+        // When exhibition is offline, if custom message is setup, show that to user, or else show default message, or else
+        // show maintenance default message if exhibition is set to maintenance mode.
+        if(exhibition.getMode()==ExhibitionModes.OFFLINE) {
+            String modeValue = exhibition.getCustomMessage().equals("") == false ? exhibition.getCustomMessage() : exhibition.getMode().getValue();
+            ((Model) args[modelIndex]).addAttribute("modeValue", modeValue);
+        } else {
+            ((Model) args[modelIndex]).addAttribute("modeValue", exhibition.getMode().getValue());
+        }
+        return "maintenance";
```
jdamerow, line 119: this is still difficult to read and unnecessary complex. Try this:

- check if mode is offline or maintenance (and no logged in user) if so set message and return maintenance
- check if space or module are null, if so return null
- check if user is logged in and offline or maintenance if so, show message
- otherwise proceed

you probably don't even need the mode==null check as this would be handled by the above points

vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
```diff
@@ -48,7 +63,104 @@ public void setExhibition(JoinPoint jp) {
...
+            return "notFound";
+        }
+        // If user is logged in and exhibition is not active, show exhibition with pop up message.
+        if(authFacade.getAuthenticatedUser()!=null && !exhibition.getMode().equals(ExhibitionModes.ACTIVE)) {
+            ((Model) args[modelIndex]).addAttribute("showModal", "true"); 
+            return jp.proceed();
+        }
+        else {
+            return jp.proceed();
+        }
```
jdamerow, line 121: get rid of else block and only have one return statement (they are both the same).

vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
```diff
@@ -48,7 +63,104 @@ public void setExhibition(JoinPoint jp) {
...
+//        // When exhibition is offline, if custom message is setup, show that to user, or else show default message, or else
+//        // show maintenance default message if exhibition is set to maintenance mode.
+//        if(exhibition.getMode()==ExhibitionModes.OFFLINE) {
+//            String modeValue = exhibition.getCustomMessage().equals("") == false ? exhibition.getCustomMessage() : exhibition.getMode().getValue();
+//            ((Model) args[modelIndex]).addAttribute("modeValue", modeValue);
+//        } else {
+//            ((Model) args[modelIndex]).addAttribute("modeValue", exhibition.getMode().getValue());
+//        }
+//        return "maintenance";
+    }
```
jdamerow, line 142: remove if not needed

vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
```diff
@@ -48,7 +63,104 @@ public void setExhibition(JoinPoint jp) {
...
+        if(exhibition.getMode().equals(ExhibitionModes.OFFLINE)) {
+            String modeValue = exhibition.getCustomMessage().equals("") == false ? exhibition.getCustomMessage() : exhibition.getMode().getValue();
+            ((Model) args[modelIndex]).addAttribute("modeValue", modeValue);
+        }
+        // If exhibition is set to maintenance, set the default message.
+        if(exhibition.getMode().equals(ExhibitionModes.MAINTENANCE)) {
+            ((Model) args[modelIndex]).addAttribute("modeValue", exhibition.getMode().getValue());
+        }
+        // If user is not logged in and exhibition is not active, show maintenance page.
+        if(authFacade.getAuthenticatedUser()==null && !exhibition.getMode().equals(ExhibitionModes.ACTIVE)) {
```
jdamerow, line 107: you've used `exhibition.getMode()` so often now, that I would put it in a variable

vspace/src/main/java/edu/asu/diging/vspace/core/model/ExhibitionModes.java
jdamerow: ?

vspace/src/main/java/edu/asu/diging/vspace/core/model/IPrefix.java
```diff
@@ -0,0 +1,7 @@
+package edu.asu.diging.vspace.core.model;
+
+public interface IPrefix {
+
+    public static String spacePrefix = "SPA";
+    public static String modulePrefix = "MOD";
```
jdamerow, line 6: constants need to use CONSTANT_CASE

jdamerow:
also, please resolve conflicts

vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
```diff
@@ -48,7 +63,83 @@ public void setExhibition(JoinPoint jp) {
...
+        }
+        return jp.proceed();
+    }
+    
+
+
+
+
+    // Method to get the space or module id's from the request body.
+    public Map<IdPrefix, String> getIds(Object[] args, MethodSignature signature) {
```
jdamerow, line 128: does this need to be public?

vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
```diff
@@ -48,7 +63,83 @@ public void setExhibition(JoinPoint jp) {
...
+        Map<IdPrefix, String> ids = getIds(args, signature);
+        String spaceId = ids.getOrDefault(IdPrefix.SPACEID, "");
+        String moduleId = ids.getOrDefault(IdPrefix.MODULEID, "");
+        return redirectRequest(jp, spaceId, moduleId, indexOfModel, exhibition);
+    }
+    
+    
+    
+    // Based on exhibition mode, get the redirect page or pass control to controller.  
+    public Object redirectRequest(ProceedingJoinPoint jp, String spaceId, String moduleId, int modelIndex, Exhibition exhibition) throws Throwable{
```
jdamerow, line 93: does this need to be public?

vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
```diff
@@ -48,7 +63,83 @@ public void setExhibition(JoinPoint jp) {
...
+        }
+        Map<IdPrefix, String> ids = getIds(args, signature);
+        String spaceId = ids.getOrDefault(IdPrefix.SPACEID, "");
+        String moduleId = ids.getOrDefault(IdPrefix.MODULEID, "");
+        return redirectRequest(jp, spaceId, moduleId, indexOfModel, exhibition);
+    }
+    
+    
+    
+    // Based on exhibition mode, get the redirect page or pass control to controller.
```
jdamerow, line 92: should be proper javadoc

vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
```diff
@@ -48,7 +63,83 @@ public void setExhibition(JoinPoint jp) {
...
+            return jp.proceed();
+        }
+        return jp.proceed();
+    }
+    
+
+
+
+
+    // Method to get the space or module id's from the request body.
```
jdamerow, line 127: should be proper javadoc

jdamerow:
Two more things:
- the offiline/maintenance page should have an image plus the message (not in an alert box_, please have  @adhvikmadhav create one.
- when the user is logged in he first sees a pop up and then the alert message. Only the alert message is enough.
image

vspace/src/main/webapp/WEB-INF/views/exhibition/maintenance.jsp
```diff
@@ -0,0 +1,16 @@
...
+<script>
+$( document ).ready(function() {
+
+});
+</script>
+<div class="" style="text-align: center;">
+<h5>${modeValue}</h5>
+</div>
+<div style="text-align: center;">
+<img alt="Not FOund" src="../../resources/maintenance/img/maintenance.png" />
```
jdamerow, line 15: should be "Found" not "FOund"

jdamerow: and src should be a `c:value` tag (something like ```<c:url value="/resources...```

jdamerow:
Jenkins, test this please.

jdamerow:
- when I go to module in maintenance mode, I get a broken image (might be fixed by my above comment)
image

- when the user is logged in and in the staff area, there should be something that shows the exhibition is offline, maybe an icon and a message on top of the menu (below "Virtual Spaces 2.0"):
image

jdamerow:
Second comment is not addressed

PR #109, opened 2020-04-03
Story/vspc 48

prjadhav14:


vspace/src/main/java/edu/asu/diging/vspace/web/ExhibitionSequencesController.java
```diff
@@ -0,0 +1,48 @@
...
+		IModule module = moduleManager.getModule(moduleId);
+		model.addAttribute("module", module);
+		List<ISequence> list=moduleManager.getModuleSequences(moduleId);
+		boolean flag=true;
+		for(ISequence sequence:list) {
+			if(sequence.getId().equals(sequenceId)) {
+				flag=false;
+				break;
+			}
+		}
```
jdamerow, line 36: use the stream and filter API of collections here

jdamerow: it's probalby even better to add a method to the sequence repo that gets a sequence by id and module.

vspace/src/main/java/edu/asu/diging/vspace/web/ExhibitionSequencesController.java
```diff
@@ -0,0 +1,48 @@
...
+
+	@Autowired
+	private IModuleManager moduleManager;
+
+	@Autowired
+	private ISequenceManager sequenceManager;
+
+	@RequestMapping(value = "/exhibit/module/{moduleId}/sequence/{sequenceId}")
+	public String sequence(Model model, @PathVariable("sequenceId") String sequenceId, @PathVariable("moduleId") String moduleId) {
+		IModule module = moduleManager.getModule(moduleId);
```
jdamerow, line 27: what if the module doesn't exist?

vspace/src/main/java/edu/asu/diging/vspace/web/ExhibitionSlideController.java
```diff
@@ -0,0 +1,73 @@
...
+	public String slide(Model model, @PathVariable("slideId") String slideId,@PathVariable("moduleId") String moduleId, @PathVariable("sequenceId") String sequenceId) {
+		IModule module = moduleManager.getModule(moduleId);
+		model.addAttribute("module", module);
+		model.addAttribute("startSequenceId", module.getStartSequence().getId());
+		model.addAttribute("firstSlide", module.getStartSequence().getSlides().get(0).getId());
+		model.addAttribute("slides", sequenceManager.getSequence(sequenceId).getSlides());
+		model.addAttribute("currentSequenceId", sequenceId);
+		model.addAttribute("currentSlide", slideId);
+		model.addAttribute("currentSlideCon", sildeManager.getSlide(slideId));
+		model.addAttribute("startSequenceId", module.getStartSequence().getId());
```
jdamerow, line 40: what if module or start sequence doesn't exist? or if slide is not part of the module?

prjadhav14: These controllers are not part of story 48 and all the above comments are addressed in story 58. I will try to delete these files from this story.

PR #110, opened 2020-04-06
Story/vspc-48_1

prjadhav14:


jdamerow:
Jenkins, test this please.

jdamerow:
- why is the menu on the left side called "Modules"? 
- the menu should be hidden by default
- let's keep the the round circles behind the icons
- generally, the layout should look more like the mockups (e.g. rounded edges of the images, no top menu, etc)
- we want to keep the exhibition title though, so let's ask Adhvik for help again.
- the image needs to be centered
- if the title of a space is too long, there is a line break that hides part of the title

PR #115, opened 2020-04-24
Story/vspc 48 1

prjadhav14:


jdamerow:
Jenkins, test this please.

jdamerow:
please make it look like the mockup (the round backgrounds for the icons, no top bar, round corners, etc)

jdamerow:
Jenkins, test this please.

jdamerow:
- same as with modules, get rid of header.
- the red labels should be removed. please talk to adhvik on how add link labels (we still want to keep the option of adding a label to a link)
image
- external link should have a new icon as well
image
- space heading should be left-aligned with picture and the modules icon in the right upper corner should be evenly spaced from top and right side
image
- the module list that open when clicking on the right upper icon should be smaller, corners not so round, and should be further away form the top (i think it looked a lot different in the mockup)
image
- the space description was in a pop up I believe in the mockup not at the bottom below the image
- the space menu on the left should have a smaller close icon and less whitespace at the top
image

jdamerow:
- if there is no space description there shouldn't be an "i" icon and description box.
- the box for the description is too long (it shouldn't be higher than the background image):
image
- the label for the links should be in a color that fits with the rest of the layout 
- the spaces menu on the right shouldn't be longer than the background image
- the menu icon for the spaces menu should be smaller
- there is something odd going on with the footer when the image is very hight:
image
- if there are no modules on a page, there shouldn't be a module list icon in the right upper corner

jdamerow:
Jenkins, test his please.

jdamerow:
Jenkins, test this please.

jdamerow:
- the space description shouldn't break words in the middle (it should break at whitespace)
- the links to a module are build differently. The icon has the space id in it, the top right corner has no space id in it.
- the space menu icon (top left corner) should be smaller
- the labels for the icons need maybe a half-transparent background or something like that. Right now, they can be very hard to read (see for example https://diging-dev.asu.edu/vspace-dev/exhibit/space/SPA1). Check with @adhvikmadhav how to make that best work.

jdamerow:
please resolve conflicts

PR #119, opened 2020-05-29
Story/vspc 72

prjadhav14:


vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SpaceManagerTest.java
```diff
@@ -108,22 +114,60 @@ public void test_storeSpace_success() throws FileStorageException {
...
+        Space space=new Space();
+        space.setId(spaceId);
+        space.setSpaceStatus(SpaceStatus.UNPUBLISHED);
+        List<Space> spaceList=new ArrayList<>();
+        spaceList.add(space);
+        Mockito.when(spaceRepo.findAllBySpaceStatus(SpaceStatus.UNPUBLISHED)).thenReturn(spaceList);
+        List<ISpace> spaceWithPublishedStatus= managerToTest.getSpacesWithStatus(SpaceStatus.UNPUBLISHED);
+        String actualSpaceIdWithUnPubStatus = spaceWithPublishedStatus.get(0).getId();
+        Assert.assertEquals(spaceId, actualSpaceIdWithUnPubStatus);
+    }
```
jdamerow, line 171: what about the case when there is no space with the given status?

jdamerow:
Jenkins, test this please.

jdamerow:
- there's something wrong with the layout of the dropdown:
image
it also doesn't show the current selection
- and the access denied page doesn't have to look pretty (there is an extra ticket for that), but at least center it in the middle of the page or something ;)
image

PR #125, opened 2020-06-08
[Story/VSPC-91]

prjadhav14:


jdamerow:
Jenkins, test this please.

jdamerow:
That just puts the footer way too far down (I have to scroll to see it) and probably even would it keep there. You probably want to use absolute positioning with a set distance from the bottom or something like that.

jdamerow:
Jenkins, test this please.

jdamerow:
problem isn't solved
image

PR #126, opened 2020-06-15
Story/vspc 92

prjadhav14:
This PR contains changes only to make module links editable. As of now, please ignore the console log statements in the JSP file.


vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditModuleLinkController.java
```diff
@@ -0,0 +1,75 @@
...
+@Controller
+public class EditModuleLinkController {
+
+    @Autowired
+    private ISpaceManager spaceManager;
+    
+    @Autowired
+    private ILinkManager linkManager;
+
+    @RequestMapping(value = "/staff/space/{id}/editmodulelink", method = RequestMethod.POST)
```
jdamerow, line 33: this url looks like the id is the space id. should be `/staff/space/link/module/{id}` or something like that

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditModuleLinkController.java
```diff
@@ -0,0 +1,75 @@
...
+        try {
+            display = linkManager.editModuleLink(title, source, new Float(x), new Float(y),
+                    new Integer(rotation), linkedModuleId, moduleLinkLabel, type, moduleLinkIdValueEdit, moduleLinkDisplayId);
+        } catch (SpaceDoesNotExistException e) {
+            ObjectMapper mapper = new ObjectMapper();
+            ObjectNode node = mapper.createObjectNode();
+            node.put("errorMessage", "space could not be found.");
+            return new ResponseEntity<>(mapper.writeValueAsString(node), HttpStatus.INTERNAL_SERVER_ERROR);
+        }
+        display.setRotation(new Integer(rotation));
```
jdamerow, line 63: why is the rotation set separately?

vspace/src/main/webapp/WEB-INF/views/staff/spaces/space.jsp
```diff
@@ -153,9 +165,11 @@ $( document ).ready(function() {
 	});
 	
 	 $("#addModuleLinkButton").click(function(e) {
+	     console.log("In add Module Link");
 		$("#createModuleLinkAlert").hide();
 		$("#bgImage").off("click");
 		$("#bgImage").on("click", function(e){
+		    console.log("On click of image");
```
jdamerow, line 172: remove

vspace/src/main/webapp/WEB-INF/views/staff/spaces/space.jsp
```diff
@@ -153,9 +165,11 @@ $( document ).ready(function() {
 	});
 	
 	 $("#addModuleLinkButton").click(function(e) {
+	     console.log("In add Module Link");
```
jdamerow, line 168: remove

vspace/src/main/webapp/WEB-INF/views/staff/spaces/space.jsp
jdamerow: and everywhere else too

vspace/src/main/java/edu/asu/diging/vspace/core/factory/impl/ModuleLinkFactory.java
```diff
@@ -24,4 +24,13 @@ public IModuleLink createModuleLink(String title, ISpace space) {
...
     }
+    
+    @Override
+    public IModuleLink editModuleLink(String title, ISpace space, String linkId) {
+        IModuleLink link = new ModuleLink();
+        link.setName(title);
+        link.setSpace(space);
+        link.setId(linkId);
+        return link;
+    }
```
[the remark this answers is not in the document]
jdamerow: ?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/LinkManager.java
```diff
@@ -302,7 +304,79 @@ public IModuleLinkDisplay createModuleLink(String title, ISpace source, float po
         moduleLinkDisplayRepo.save((ModuleLinkDisplay) display);
         return display;
     }
+    
+    @Override
+    public IModuleLinkDisplay editModuleLink(String title, ISpace source, float positionX, float positionY,
+            int rotation, String linkedModuleId, String moduleLinkLabel, String linkId, String moduleLinkDisplayId)
+            throws SpaceDoesNotExistException {
+        
+        source = spaceManager.getSpace(source.getId());
```
jdamerow, line 313: just pass the id, if you don't need the whole space

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/LinkManager.java
```diff
@@ -302,7 +304,79 @@ public IModuleLinkDisplay createModuleLink(String title, ISpace source, float po
...
+        if (source == null) {
+            throw new SpaceDoesNotExistException();
+        }
+        Optional<ModuleLink> linkOptional = moduleLinkRepo.findById(linkId);
+        IModuleLink link = linkOptional.get();
```
jdamerow, line 318: you should check if the link is present first or you'll get an exception if the link id doesn't exist

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/LinkManager.java
```diff
@@ -302,7 +304,79 @@ public IModuleLinkDisplay createModuleLink(String title, ISpace source, float po
...
+        link.setName(title);
+        IModule module = new Module();
+        module.setId(linkedModuleId);
+        link.setModule(module);
```
jdamerow, line 322: that seems odd, why don't you get the actual module it links to and set it here?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/LinkManager.java
```diff
@@ -302,7 +304,79 @@ public IModuleLinkDisplay createModuleLink(String title, ISpace source, float po
...
+        moduleLinkRepo.save((ModuleLink) link);
+        Optional<ModuleLinkDisplay> moduleLinkOptional = moduleLinkDisplayRepo.findById(moduleLinkDisplayId);
+        IModuleLinkDisplay display = moduleLinkOptional.get();
```
jdamerow, line 325: same as above and also, I would first get all objects from the db and set all the values and then save them all. Otherwise, it's a) harder to follow what is changed and saved (it's a lot easier to just find the block of code that saves objects to see which objects are being saved) and also if for example the display object doesn't exist the link object should be changed and there is no db transaction that needs to be rolled back.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/LinkManager.java
```diff
@@ -302,7 +304,79 @@ public IModuleLinkDisplay createModuleLink(String title, ISpace source, float po
...
+        IExternalLink link = linkOptional.get();
+        link.setName(title);
+        link.setExternalLink(externalLink);
+        externalLinkRepo.save((ExternalLink) link);
+        Optional<ExternalLinkDisplay> externalLinkOptional = externalLinkDisplayRepo.findById(externalLinkDisplayId);
+        IExternalLinkDisplay display = externalLinkOptional.get();
+        display.setPositionX(positionX);
+        display.setPositionY(positionY);
+        externalLinkDisplayRepo.save((ExternalLinkDisplay) display);
+        return display;
```
jdamerow, line 376: looks like this the same code three times. there should be one method that does all the common stuff and the individual methods call this one method (this will be easier to do with the above described changes).

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditSpaceLinkController.java
```diff
@@ -0,0 +1,79 @@
...
+
+import edu.asu.diging.vspace.core.exception.ImageCouldNotBeStoredException;
+import edu.asu.diging.vspace.core.exception.SpaceDoesNotExistException;
+import edu.asu.diging.vspace.core.model.ISpace;
+import edu.asu.diging.vspace.core.model.display.ISpaceLinkDisplay;
+import edu.asu.diging.vspace.core.services.ILinkManager;
+import edu.asu.diging.vspace.core.services.ISpaceManager;
+
+@Controller
+public class EditSpaceLinkController {
```
jdamerow, line 26: there is a lot of duplicate code in the three controllers. I would create a superclass and pull up the common code.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/LinkManager.java
prjadhav14: display is an instance of different objects in each of the functions.

jdamerow: but they inherit from the same superclass, don't they?

prjadhav14: Yes, converted into single function,

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditSpaceLinkController.java
```diff
@@ -0,0 +1,47 @@
...
+            @RequestParam("spaceLinkIdValueEdit") String spaceLinkIdValueEdit, @RequestParam("spaceLinkDisplayId") String spaceLinkDisplayId)
+                    throws NumberFormatException, SpaceDoesNotExistException, LinkDoesNotExistsException, IOException {
+
+        
+        ResponseEntity<String> validation = checkIfSpaceExists(spaceManager, id, x, y);
+        if(validation!=null) {
+            return validation;
+        }
+        ISpaceLinkDisplay display;
+        display = linkManager.editSpaceLink(title, id, new Float(x), new Float(y),
```
jdamerow, line 41: merge with line 40

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditSpaceLinks.java
```diff
@@ -0,0 +1,50 @@
...
+        if (x == null || x.trim().isEmpty() || y == null || y.trim().isEmpty()) {
+            ObjectMapper mapper = new ObjectMapper();
+            ObjectNode node = mapper.createObjectNode();
+            node.put("errorMessage", "No link coordinates specified.");
+            return new ResponseEntity<String>(mapper.writeValueAsString(node), HttpStatus.BAD_REQUEST);
+        }
+        return null;
+    }
+    
+    public ResponseEntity<String> imageCouldNotBeStored() throws IOException{
```
jdamerow, line 30: is that being used? also, it doesn't look like the image can be changed currently?

prjadhav14: Removed the block of code that is not used.
No, the image on space is not editable. Should I work on making image type and alert type links editable?

jdamerow: all properties of links need to be editable

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/LinkManager.java
```diff
@@ -306,8 +308,127 @@ public IModuleLinkDisplay createModuleLink(String title, ISpace source, float po
         return display;
     }
 
+    @Override
+    public IModuleLinkDisplay editModuleLink(String title, String id, float positionX, float positionY,
```
jdamerow, line 312: I would refactor here a bit. To avoid this class getting too big, create subclasses for each Link type and use the template pattern to create the methods:
- the superclass has save/edit/delete/etc methods that call abstract methods (implemented in the child classes) that do the specifics.
- e.g. superclass calls `spaceValidation`, `linksValidation` etc and then calls abstract methods that set link properties, etc.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditSpaceLinks.java
```diff
@@ -0,0 +1,43 @@
...
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+import edu.asu.diging.vspace.core.model.ISpace;
+import edu.asu.diging.vspace.core.services.ISpaceManager;
+
+public class EditSpaceLinks {
```
jdamerow, line 14: bad class name, it's the superclass for controllers, right? so it should be called controller.

vspace/src/main/webapp/WEB-INF/views/staff/spaces/space.jsp
```diff
@@ -125,11 +127,18 @@ $( document ).ready(function() {
...
-    
+	$("#editModuleLinkInfo").draggable();
+	$("#editSpaceLinkInfo").draggable();
+	$("#editExternalLinkInfo").draggable();
     
 	// store where a user clicked on an image
 	var storeX;
 	var storeY;
+	/* var storeXEdit;
+	var storeYEdit; */
```
jdamerow, line 138: remove if not needed

vspace/src/main/webapp/WEB-INF/views/staff/spaces/space.jsp
```diff
@@ -285,6 +294,109 @@ $( document ).ready(function() {
...
 		});
 	});
+	$("#editModuleLinkBtn").click(function(e) {
+	    e.preventDefault();
+	    var linkId = $("#moduleLinkIdEdit").val();
+	    if(storeX == undefined || storeY == undefined){
+	        $("#errorMsg").text("Please click on the image to specify where the new link should be located.");
+	        $('#errorAlert').show();
+	        return;
+	   	}
```
jdamerow, line 304: indentation

vspace/src/main/webapp/WEB-INF/views/staff/spaces/space.jsp
```diff
@@ -679,46 +874,147 @@ $( document ).ready(function() {
...
+	}
+	
+	function editExternalLinkInfo() {
+		var info = {};
+		info["x"] = storeX;
+		info["y"] = storeY;
+		info["type"] = $("#extTypeEdit").val();
+		info["externalLinkLabel"] = $("#externalLinkLabelEdit").val();
+		info["externalLinkURL"] = $("#externalLinkURLEdit").val();
+	    return info;
```
jdamerow, line 906: indentation

vspace/src/main/webapp/WEB-INF/views/staff/spaces/space.jsp
```diff
@@ -679,46 +874,147 @@ $( document ).ready(function() {
...
+		$("#spaceLinkInfoLabel").text(spaceLinkName);
+		$("#spaceLinkDisplayId").val(displayLinkId);
+		$("#spaceLinkInfoLabelEdit").text(spaceLinkName);
+		$("#spaceLinkLabelEdit").val(spaceLinkName);
+		$("#spaceLinkId").val(spaceLinkId);
+		$("#spaceLinkIdValueEdit").val(spaceLinkId);
+		$("#spaceLinkRotationEdit").val(rotation);
+		
+		$('#typeSpaceEdit option[value="'+linkType+'"]').attr("selected", "selected");
+		$('#spaceLinkIdEdit option[value="'+selectedSpaceId+'"]').attr("selected", "selected");
```
jdamerow, line 925: indentation

vspace/src/main/webapp/WEB-INF/views/staff/spaces/space.jsp
```diff
@@ -679,46 +874,147 @@ $( document ).ready(function() {
...
         $('img[data-link-id="' + spaceLinkId + '"]').css("border", "solid 1px #c1bb88");
+        
+        $("#bgImage").on("click", function(e){
+			e.preventDefault();
+			var posX = $(this).position().left;
+			var posY = $(this).position().top;    
+			storeX = e.pageX - $(this).offset().left;
+			storeY = e.pageY - $(this).offset().top;
+			showSpaceLinkEdit(editSpaceLinkInfo());
+		});
```
jdamerow, line 941: indentation

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditSpaceLinksController.java
```diff
@@ -0,0 +1,43 @@
...
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+import edu.asu.diging.vspace.core.model.ISpace;
+import edu.asu.diging.vspace.core.services.ISpaceManager;
+
+public class EditSpaceLinksController {
+
+    public ResponseEntity<String> checkIfSpaceExists(ISpaceManager spaceManager, String id, String x, String y) throws IOException{
```
jdamerow, line 16: these methods should not be public unless they are indented to be used outside of the inheriting classes (which they are not, are they?).

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/LinkManager.java
```diff
@@ -10,76 +9,25 @@
...
+import edu.asu.diging.vspace.core.model.display.ILinkDisplay;
 import edu.asu.diging.vspace.core.model.impl.VSImage;
-import edu.asu.diging.vspace.core.services.ILinkManager;
-import edu.asu.diging.vspace.core.services.IModuleManager;
 import edu.asu.diging.vspace.core.services.ISpaceManager;
 
-@Transactional
-@Service
-public class LinkManager implements ILinkManager {
+public abstract class LinkManager{
```
jdamerow, line 26: where are the subclasses? they should be in this pr too, shouldn't they?

vspace/src/main/java/edu/asu/diging/vspace/core/services/IExternalLinkManager.java
```diff
@@ -0,0 +1,26 @@
...
+public interface IExternalLinkManager {
+    public IExternalLinkDisplay createLink(String title, ISpace source, float positionX, float positionY, int rotation,
+            String externalLink, String externalLabel, DisplayType displayType, byte[] linkImage, String imageFilename)
+                    throws SpaceDoesNotExistException, ImageCouldNotBeStoredException, SpaceDoesNotExistException;
+
+    public List<ExternalLinkDisplay> getLinkDisplays(String spaceId);
+
+    public void deleteLink(String linkId);
+
+    public IExternalLinkDisplay editLink(String title, String id, float positionX, float positionY, int rotation,
```
jdamerow, line 22: let's call this 'updateLink', usually "update" is standard not "edit"

vspace/src/main/java/edu/asu/diging/vspace/core/services/IModuleLinkManager.java
```diff
@@ -0,0 +1,27 @@
...
+
+    public IModuleLinkDisplay createLink(String title, ISpace source, float positionX, float positionY, int rotation,
+            String linkedModuleId, String moduleLinkLabel, DisplayType displayType, byte[] linkImage, String imageFilename)
+                    throws SpaceDoesNotExistException, ImageCouldNotBeStoredException, SpaceDoesNotExistException;
+
+    public List<ModuleLinkDisplay> getLinkDisplays(String spaceId);
+
+    public void deleteLink(String linkId);
+
+    public IModuleLinkDisplay editLink(String title, String id, float positionX, float positionY, int rotation,
```
jdamerow, line 23: as above

vspace/src/main/java/edu/asu/diging/vspace/core/services/ISpaceLinkManager.java
```diff
@@ -0,0 +1,23 @@
...
+import edu.asu.diging.vspace.core.model.display.DisplayType;
+import edu.asu.diging.vspace.core.model.display.ISpaceLinkDisplay;
+import edu.asu.diging.vspace.core.model.display.impl.SpaceLinkDisplay;
+
+public interface ISpaceLinkManager {
+    public ISpaceLinkDisplay createLink(String title, ISpace source, float positionX, float positionY, int rotation,
+            String linkedSpaceId, String spaceLinkLabel, DisplayType displayType, byte[] linkImage, String imageFilename)
+                    throws SpaceDoesNotExistException, ImageCouldNotBeStoredException, SpaceDoesNotExistException;
+    public List<SpaceLinkDisplay> getLinkDisplays(String spaceId);
+    public ISpaceLinkDisplay editLink(String title, String id, float positionX, float positionY, int rotation,
```
jdamerow, line 18: same as above

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ExternalLinkManager.java
```diff
@@ -0,0 +1,150 @@
...
+    private ImageRepository imageRepo;
+
+    @Autowired
+    private IStorageEngine storage;
+
+    @Override
+    public IExternalLinkDisplay createLink(String title, ISpace source, float positionX, float positionY, int rotation,
+            String externalLink, String externalLabel, DisplayType displayType, byte[] linkImage, String imageFilename)
+                    throws SpaceDoesNotExistException, ImageCouldNotBeStoredException, SpaceDoesNotExistException {
+        source = spaceManager.getSpace(source.getId());
```
jdamerow, line 68: Why don't you pass the id if you just need that?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ModuleLinkManager.java
```diff
@@ -0,0 +1,125 @@
...
+import edu.asu.diging.vspace.core.model.display.IModuleLinkDisplay;
+import edu.asu.diging.vspace.core.model.display.impl.ModuleLinkDisplay;
+import edu.asu.diging.vspace.core.model.impl.ModuleLink;
+import edu.asu.diging.vspace.core.services.IModuleLinkManager;
+import edu.asu.diging.vspace.core.services.IModuleManager;
+import edu.asu.diging.vspace.core.services.ISpaceManager;
+
+@Transactional
+@Service
+public class ModuleLinkManager extends LinkManager implements IModuleLinkManager{
```
jdamerow, line 33: I think the template pattern would work well here:
- LinkManager has delete/create/update/...Link methods that call abstract methods which are implemented in the subclasses and do the link specific stuff.
- create a link superclass/interface that has the common stuff (owning space)
- you can probably even use generics to make this more elegant, but maybe start with the other stuff first.

prjadhav14: I did not understand methods calling an abstract method part in the first bullet point. Currently, LinkManager contains abstract methods whose implementation is written in child classes like ExternalLinkManager and its interfaces are used to track implementation of which class is to be invoked while creating or deleting links. Can we please discuss this on zoom?

jdamerow: Yes, we can discuss this after my meeting. But basically you would have something like this (pseudo code):
```
class LinkManager:
      method createLink():
             // get space to the link belongs to
             
             // call abstract method to get link target
             target = getTarget()
             
             // call abstract method to create link
             link = createLinkObject()
             ...

class ModuleLinkManager:
        method getTarget():
               // gets and returns the link target (for module links the linked module)
```

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/ExternalLinkValue.java
```diff
@@ -0,0 +1,33 @@
+package edu.asu.diging.vspace.core.model.impl;
+
+public class ExternalLinkValue extends VSpaceElement {
+
+    private String value;
+
+    public ExternalLinkValue() {
+
+    }
```
jdamerow, line 9: remove empty line and make this one line

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/ModuleLink.java
```diff
@@ -12,65 +12,76 @@
...
+     * @see edu.asu.diging.vspace.core.model.impl.IModuleLink#setModule(edu.asu.diging.vspace.core.model.IModule)
+     */
+    @Override
+    public void setModule(IModule module) {
+        this.module = module;
+    }
+
+    @Override
+    public IVSpaceElement getTarget() {
+        return this.module;
```
jdamerow, line 80: this should return IModule

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/SpaceLink.java
```diff
@@ -75,4 +76,21 @@ public ISpace getTargetSpace() {
...
+    public ISpace getSpace() {
+        return sourceSpace;
+    }
+    @Override
+    public void setSpace(ISpace space) {
+        this.sourceSpace = space;
+        
+    }
+    @Override
+    public IVSpaceElement getTarget() {
```
jdamerow, line 89: this should return ispace

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ExternalLinkManager.java
```diff
@@ -0,0 +1,118 @@
...
+        IExternalLink link = externalLinkFactory.createExternalLink(title, source, ((ExternalLinkValue)target).getValue());
+        link.setName(linkLabel);
+        externalLinkRepo.save((ExternalLink) link);
+        return link;
+    }
+
+    @Override
+    protected IVSpaceElement getTarget(String externalLink) {
+        ExternalLinkValue value = new ExternalLinkValue(externalLink);
+        return value;
```
jdamerow, line 80: merge with line 79

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ExternalLinkManager.java
```diff
@@ -0,0 +1,118 @@
...
+    @Autowired
+    private IExternalLinkDisplayFactory externalLinkDisplayFactory;
+
+    @Override
+    public List<ILinkDisplay> getLinkDisplays(String spaceId) {
+        return new ArrayList<>(externalLinkDisplayRepo.findExternalLinkDisplaysForSpace(spaceId));
+    }
+
+    @Override
+    public void deleteLink(String linkId) {
```
jdamerow, line 54: my suspicion is that this should be part of linkmanager too and just call abstract methods.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/LinkManager.java
```diff
@@ -89,182 +36,73 @@
...
-        // we need this to fully load the space
-        source = spaceManager.getSpace(source.getId());
-        if (source == null) {
-            throw new SpaceDoesNotExistException();
-        }
+        IVSpaceElement target = getTarget(linkedId);
+        ILink link = (ILink) createLinkObject(title, id, target, linkLabel); 
+        ILinkDisplay displayLink = createLinkDisplay(link);
+        setDisplayProperties(displayLink, positionX, positionY, rotation, displayType, linkImage, imageFilename);
+        return saveLinkAndDisplay(link, displayLink);
```
jdamerow, line 47: if you add an abstract method `getLinkRepo` and `getDisplayRepo`, then you don't need saveLinkAndDisplay and you might be able to get rid of other methods as well.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditModuleLinkController.java
```diff
@@ -0,0 +1,49 @@
...
+            @RequestParam("type") String displayType)
+                    throws NumberFormatException, SpaceDoesNotExistException, LinkDoesNotExistsException, IOException, ImageCouldNotBeStoredException {
+
+        ResponseEntity<String> validation = checkIfSpaceExists(spaceManager, id, x, y);
+        if(validation!=null) {
+            return validation;
+        }
+        DisplayType type = displayType.isEmpty() ? null : DisplayType.valueOf(displayType);
+        IModuleLinkDisplay display;
+        display = (IModuleLinkDisplay) moduleLinkManager.updateLink(title, id, new Float(x), new Float(y),
```
jdamerow, line 44: merge with 43

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ExternalLinkManager.java
prjadhav14: we don't have much common code in the delete method which can be coded in LinkManager superclass. It would be just one abstract delete method that can be defined in sub-classes and invoked through parent delete in LinkManager.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/LinkManager.java
prjadhav14: here we are creating and saving a new link and display link. getLinkRepo and getDisplayRepo are used in update methods. getting rid of saveLinkAndDisplay from create method though.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ExternalLinkManager.java
jdamerow: they all the the same thing though, you can probably write it like this in the superclass:
```
ILink link = getLinkById(linkId)
// ... return if null

removeFromLinkList(link.getSpace()) // abstract method
deleteLinkDisplay() // abstract method
deleteLink() // abstract method
```
this way the subclasses just have to implement the abstract methods and don't worry about the order. If we make changes that require the delete method to change order or do something additional, all we have to do is to change the one method in the link manager.

there is probably an even more elegant way using generics, but this should work.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ExternalLinkManager.java
```diff
@@ -0,0 +1,123 @@
...
+        IExternalLinkDisplay display = externalLinkOptional.get();
+        return display;
+    }
+
+    @Override
+    protected ILink getLink(String externalLinkIDValueEdit) throws LinkDoesNotExistsException {
+        Optional<ExternalLink> linkOptional = externalLinkRepo.findById(externalLinkIDValueEdit);
+        linksValidation(linkOptional);
+        IExternalLink link = linkOptional.get();
+        return link;
```
jdamerow, line 110: this should return null if the link doesn't exist

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ExternalLinkManager.java
```diff
@@ -0,0 +1,123 @@
...
+    @Override
+    protected ILinkDisplay getDisplayLink(String externalLinkDisplayId) throws LinkDoesNotExistsException {
+        Optional<ExternalLinkDisplay> externalLinkOptional = externalLinkDisplayRepo.findById(externalLinkDisplayId);
+        linksValidation(externalLinkOptional);
+        IExternalLinkDisplay display = externalLinkOptional.get();
+        return display;
+    }
+
+    @Override
+    protected ILink getLink(String externalLinkIDValueEdit) throws LinkDoesNotExistsException {
```
jdamerow, line 106: why is the parameter called externalLinkIDValue*Edit*?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/LinkManager.java
```diff
@@ -89,182 +36,72 @@
...
-        display.setPositionY(positionY);
-        display.setName(title);
-        display.setType(displayType != null ? displayType : DisplayType.ARROW);
+    public void linksValidation(Optional link) throws LinkDoesNotExistsException {
+        if(!link.isPresent()) {
+            throw new LinkDoesNotExistsException("Link Does Not Exists");
+        }
+    }
 
+    public void setDisplayProperties(ILinkDisplay linkDisplay,float positionX,float positionY,int rotation, DisplayType displayType, byte[] linkImage, String imageFilename) throws ImageCouldNotBeStoredException {
```
jdamerow, line 95: is that used anywhere outside of the manager classes? if not, it should not be public.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/LinkManager.java
```diff
@@ -89,182 +36,72 @@
...
-        moduleLinkRepo.delete((ModuleLink) link);
-    }
-    
-    @Override
-    public void deleteExternalLink(String linkId) {
-        Optional<ExternalLink> linkOptional = externalLinkRepo.findById(linkId);
-        if (!linkOptional.isPresent()) {
-            return;
-        }
+    protected abstract void setTarget(ILink link, IVSpaceElement target);
```
jdamerow, line 72: thinking about this, i would actually use generics in this class as well. this way this method could be
```
setTarget(L link, T target)
```
where L extends ILink and T extends IVSpaceElement. You could avoid a lot of casting in the subclasses.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ExternalLinkManager.java
prjadhav14: throws exception from linkValidation method if link is not present.

jdamerow: then either the method shouldn't throw an exception (it can just return null which would be fine), or you need to handle the exception here. but all other `get...` methods return null if element is not found, the same should be true here.

vspace/src/main/java/edu/asu/diging/vspace/core/model/ILLink.java
```diff
@@ -0,0 +1,5 @@
+package edu.asu.diging.vspace.core.model;
+
+public interface ILLink<T> extends ILink<IVSpaceElement>{
```
jdamerow, line 3: what is this interface for?

vspace/src/main/java/edu/asu/diging/vspace/core/model/ILink.java
```diff
@@ -0,0 +1,12 @@
+package edu.asu.diging.vspace.core.model;
+
+public interface ILink<T extends IVSpaceElement> extends IVSpaceElement{
+
+    ISpace getSpace();
+
+    void setSpace(ISpace space);
+
+    IVSpaceElement getTarget();
```
jdamerow, line 9: this should return T

vspace/src/main/java/edu/asu/diging/vspace/core/model/ITVSpaceElement.java
```diff
@@ -0,0 +1,5 @@
+package edu.asu.diging.vspace.core.model;
+
+public interface ITVSpaceElement<T> extends IVSpaceElement{
```
jdamerow, line 3: what is this interface for?

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/ModuleLink.java
```diff
@@ -12,65 +12,81 @@
...
+    public IModule getTarget() {
+        return this.module;
+    }
+
+    //    @Override
+    //    public void setTarget(IModule target) {
+    //        this.module = target;
+    //    }
+    @Override
+    public void setTarget(IVSpaceElement target) {
```
jdamerow, line 88: this should get a imodule

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/SpaceLink.java
```diff
@@ -13,66 +13,90 @@
...
+
+    }
+    @Override
+    public ISpace getTarget() {
+        return targetSpace;
+    }
+    //    @Override
+    //    public void setTarget(ISpace target) {
+    //        this.targetSpace = target;
+    //    }
```
jdamerow, line 95: ?

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/SpaceLink.java
```diff
@@ -13,66 +13,90 @@
...
+    @Override
+    public void setTarget(IVSpaceElement target) {
```
jdamerow, line 97: should get ISpace

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ExternalLinkManager.java
jdamerow: ?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ExternalLinkManager.java
jdamerow: mrege with line 114

vspace/src/main/java/edu/asu/diging/vspace/core/services/ILinkManager.java
```diff
@@ -3,34 +3,23 @@
...
+import edu.asu.diging.vspace.core.exception.LinkDoesNotExistsException;
 import edu.asu.diging.vspace.core.exception.SpaceDoesNotExistException;
-import edu.asu.diging.vspace.core.model.ISpace;
 import edu.asu.diging.vspace.core.model.display.DisplayType;
-import edu.asu.diging.vspace.core.model.display.IExternalLinkDisplay;
-import edu.asu.diging.vspace.core.model.display.IModuleLinkDisplay;
-import edu.asu.diging.vspace.core.model.display.ISpaceLinkDisplay;
+import edu.asu.diging.vspace.core.model.display.ILinkDisplay;
 
 public interface ILinkManager {
```
jdamerow, line 11: this should probably be parameterized

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/LinkManager.java
jdamerow: everywhere you use IVSpaceElement, it should use T

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ExternalLinkManager.java
```diff
@@ -0,0 +1,124 @@
...
+            return null;
+        }
+        return linkOptional.get();
+    }
+
+    @Override
+    protected ILinkDisplay saveDisplayLinkRepo(ILink link, float positionX, float positionY, int rotation,
+            DisplayType displayType, byte[] linkImage, String imageFilename) throws ImageCouldNotBeStoredException {
+        ILinkDisplay displayLink = externalLinkDisplayFactory.createExternalLinkDisplay((IExternalLink)link);
+        setDisplayProperties(displayLink, positionX, positionY, rotation, displayType, linkImage, imageFilename);
```
jdamerow, line 103: since that's the same for all links (it is, right?) it should be called in the superclass.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceLinkManager.java
```diff
@@ -0,0 +1,125 @@
...
+
+    @Override
+    protected ILink createLinkObject(String title, String id, IVSpaceElement target, String spaceLinkLabel) {
+
+        ISpace source = spaceManager.getSpace(id);
+        ISpaceLink link = spaceLinkFactory.createSpaceLink(title, source);
+        link.setTarget((ISpace)target);
+        link.setName(spaceLinkLabel);
+        spaceLinkRepo.save((SpaceLink) link);
+        return (ILink) link;
```
jdamerow, line 62: everything that's called in each subclass should be called in the superclass

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ExternalLinkManager.java
prjadhav14: Yes, but this needs to be called after the creation of DisplayLink and before saving link and DisplayLink. Also, to acieve this I would need to create two abstract methods in superclass. We did this to reduce number of methods in superclass.

jdamerow: but this way you copy/paste the same code into all the subclasses just with slight modifications; that's more error prone that having more abstract methods that are pretty straight-forward to implement.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceLinkManager.java
```diff
@@ -0,0 +1,115 @@
...
+    @Override
+    protected void setTarget(ISpaceLink link, ISpace target) {
+        link.setTarget(target);
+    }
+
+    @Override
+    protected ILinkDisplay updateLinkAndDisplay(ISpaceLink link, ILinkDisplay displayLink) {
+        spaceLinkRepo.save((SpaceLink) link);
+        spaceLinkDisplayRepo.save((SpaceLinkDisplay) displayLink);
+        return displayLink;
```
jdamerow, line 92: merge with line 91

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/LinkManager.java
```diff
@@ -89,182 +38,86 @@
...
-    @Override
-    public ISpaceLinkDisplay createSpaceLink(String title, ISpace source, float positionX, float positionY,
-            int rotation, String linkedSpaceId, String spaceLinkLabel, DisplayType displayType, byte[] linkImage,
-            String imageFilename) throws ImageCouldNotBeStoredException, SpaceDoesNotExistException {
-        // we need this to fully load the space
-        source = spaceManager.getSpace(source.getId());
-        if (source == null) {
-            throw new SpaceDoesNotExistException();
-        }
+    public ILinkDisplay createLink(String title, String id, float positionX, float positionY,
```
jdamerow, line 41: since you already have two generic types, might add this one as well.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ModuleLinkManager.java
```diff
@@ -0,0 +1,118 @@
...
+        link.setModule(target);
+        link.setName(linkLabel);
+        return link;
+    }
+
+    @Override
+    protected ILinkDisplay updateLinkAndDisplay(IModuleLink link, ILinkDisplay displayLink) {
+        moduleLinkRepo.save((ModuleLink) link);
+        moduleLinkDisplayRepo.save((ModuleLinkDisplay) displayLink);
+        return displayLink;
```
jdamerow, line 71: merge with line 70

vspace/src/main/webapp/WEB-INF/views/staff/spaces/space.jsp
jdamerow: ?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/LinkManager.java
```diff
@@ -1,85 +1,32 @@
...
+import edu.asu.diging.vspace.core.model.display.ILinkDisplay;
 import edu.asu.diging.vspace.core.model.impl.VSImage;
 import edu.asu.diging.vspace.core.services.ILinkManager;
-import edu.asu.diging.vspace.core.services.IModuleManager;
 import edu.asu.diging.vspace.core.services.ISpaceManager;
 
 @Transactional
-@Service
-public class LinkManager implements ILinkManager {
+public abstract class LinkManager<L,T, U> implements ILinkManager<L,T,U>{
```
jdamerow, line 25: 2 things: 
- I don't think you need L,T, U in the class definition. should be enough in the interface I believe.
- if you make `L extends ILink` you don't need to cast L e.g. in line 45

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/LinkManager.java
```diff
@@ -89,182 +36,78 @@
...
-     */
-    @Override
-    public IExternalLinkDisplay createExternalLink(String title, ISpace source, float positionX, float positionY,
-            String externalLink, DisplayType displayType, byte[] linkImage, String imageFilename)
-            throws ImageCouldNotBeStoredException, SpaceDoesNotExistException {
-        // we need this to fully load the space
-        source = spaceManager.getSpace(source.getId());
+    protected abstract U createDisplayLink(L link);
+
+    protected void spaceValidation(String id) throws SpaceDoesNotExistException{
```
jdamerow, line 94: methods should be verbs, so e.g. `validateSpace`

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ModuleLinkManager.java
jdamerow: with line 66

vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddExternalLinkController.java
```diff
@@ -20,38 +20,38 @@
...
         String filename = null;
         if (file != null) {
             linkImage = file.getBytes();
             filename = file.getOriginalFilename();
         }
         DisplayType type = displayType.isEmpty() ? null : DisplayType.valueOf(displayType);
-        
-        IExternalLinkDisplay display = linkManager.createExternalLink(title, space, new Float(x), new Float(y), externalLink, type, linkImage, filename);
+
+        IExternalLinkDisplay display = (IExternalLinkDisplay) externalLinkManager.createLink(title, id, new Float(x), new Float(y), 0, externalLink, title, type, linkImage, filename);
```
jdamerow, line 54: there should be no need for casting here anymore

vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddModuleLinkController.java
```diff
@@ -52,8 +54,8 @@
         DisplayType type = displayType.isEmpty() ? null : DisplayType.valueOf(displayType);
         IModuleLinkDisplay display;
         try {
-            display = linkManager.createModuleLink(title, source, new Float(x), new Float(y),
-                    new Integer(rotation), linkedModuleId, moduleLinkLabel, type);
+            display = (IModuleLinkDisplay) moduleLinkManager.createLink(title, id, new Float(x), new Float(y),
```
jdamerow, line 57: or here

vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddSpaceLinkController.java
```diff
@@ -61,7 +61,7 @@
         DisplayType type = displayType.isEmpty() ? null : DisplayType.valueOf(displayType);
         ISpaceLinkDisplay display;
         try {
-            display = linkManager.createSpaceLink(title, source, new Float(x), new Float(y),
+            display = (ISpaceLinkDisplay) spaceLinkManager.createLink(title, id, new Float(x), new Float(y),
```
jdamerow, line 64: or here

vspace/src/main/webapp/WEB-INF/views/staff/spaces/space.jsp
```diff
@@ -294,6 +300,109 @@ $( document ).ready(function() {
...
+	            var linkData = JSON.parse(data);
+	            $("#bgImage").off("click");
+	            externalLinkInfo["id"] = linkData["id"];
+	            showExternalLinkEdit(editExternalLinkInfo, true);
+	            $("#editExternalLinkInfo").hide();
+	            $("#errorMsg").text("");
+	            $('#errorAlert').hide();
+	       	}
+	 	});
+	});
```
jdamerow, line 405: this looks like it's the same code for all three link types (with just different html elements, classes ,etc). common code should be in methods so you don't repeat yourself.

vspace/src/main/webapp/WEB-INF/views/staff/spaces/space.jsp
```diff
@@ -548,6 +675,47 @@ $( document ).ready(function() {
...
+		var posY = $("#bgImage").position().top;
+		externalLink["x"]=storeX;
+		externalLink["y"]=storeY;
+		$('.elabel-'+selectedExternalLinkId).text(externalLink["externalLinkLabel"]);
+		$('.externalLink-'+selectedExternalLinkId).css({ 'left': externalLink["x"] + posX});
+		$('.externalLink-'+selectedExternalLinkId).css({ 'top': externalLink["y"] + posY});
+		$('.elabel-'+selectedExternalLinkId).css({ 'left': externalLink["x"] + posX - 10});
+		$('.elabel-'+selectedExternalLinkId).css({ 'top': externalLink["y"] + posY + 16});
+		feather.replace();
+	}
```
jdamerow, line 717: same here

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SpaceLinkManagerTest.java
```diff
@@ -0,0 +1,67 @@
...
+        ISpaceLink spaceLinkActual = managerToTest.getLink(newSpaceLink.getId());
+        Assert.assertEquals(mockSpaceLink.get().getId(), spaceLinkActual.getId());
+    }
+
+    @Test
+    public void test_getLink_idNotExists() throws Exception {
+        Mockito.when(spaceLinkRepo.findById(Mockito.anyString())).thenReturn(Optional.empty());
+        Assert.assertNull(managerToTest.getLink(Mockito.anyString()));
+    }
+}
```
jdamerow, line 67: what about delete and update link methods?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/LinkManager.java
```diff
@@ -1,85 +1,32 @@
...
+import edu.asu.diging.vspace.core.model.display.ILinkDisplay;
 import edu.asu.diging.vspace.core.model.impl.VSImage;
 import edu.asu.diging.vspace.core.services.ILinkManager;
-import edu.asu.diging.vspace.core.services.IModuleManager;
 import edu.asu.diging.vspace.core.services.ISpaceManager;
 
 @Transactional
-@Service
-public class LinkManager implements ILinkManager {
+public abstract class LinkManager<L extends ILink,T extends IVSpaceElement, U extends ILinkDisplay> implements ILinkManager{
```
jdamerow, line 25: the interface should have the same extends constraints

jdamerow: and then you would probably parameterize the interface here too. does Eclipse not give you a warning?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/LinkManager.java
```diff
@@ -89,182 +36,78 @@
...
-            imageRepo.save((VSImage) image);
-
-            display.setImage(image);
-        }
-
-        spaceLinkDisplayRepo.save((SpaceLinkDisplay) display);
-        return display;
+    public void deleteLink(String linkId){
+        L link = getLink(linkId);
+        removeFromLinkList(((ILink) link).getSpace(),link);
```
jdamerow, line 71: do you really need to cast here?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ModuleLinkManager.java
```diff
@@ -0,0 +1,107 @@
...
+import edu.asu.diging.vspace.core.model.ISpace;
+import edu.asu.diging.vspace.core.model.display.impl.ModuleLinkDisplay;
+import edu.asu.diging.vspace.core.model.impl.ModuleLink;
+import edu.asu.diging.vspace.core.services.IModuleLinkManager;
+import edu.asu.diging.vspace.core.services.IModuleManager;
+import edu.asu.diging.vspace.core.services.ISpaceManager;
+
+@Transactional
+@Service
+public class ModuleLinkManager extends LinkManager<IModuleLink,IModule,ModuleLinkDisplay> implements IModuleLinkManager{
```
jdamerow, line 26: same as above

vspace/src/main/java/edu/asu/diging/vspace/core/services/IModuleLinkManager.java
```diff
@@ -0,0 +1,5 @@
+package edu.asu.diging.vspace.core.services;
+
+public interface IModuleLinkManager<IModuleLink,IModule,IModuleLinkDisplay> extends ILinkManager<IModuleLink,IModule,IModuleLinkDisplay>{
```
jdamerow, line 3: I don't think you need the parameters for the IModuleLinkManager (does Eclipse not show you a warning?)>

vspace/src/main/java/edu/asu/diging/vspace/core/services/ISpaceLinkManager.java
```diff
@@ -0,0 +1,5 @@
+package edu.asu.diging.vspace.core.services;
+
+public interface ISpaceLinkManager<ISpaceLink,ISpace,SpaceLinkDisplay> extends ILinkManager<ISpaceLink,ISpace,SpaceLinkDisplay>{
```
jdamerow, line 3: same here

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ModuleLinkManager.java
```diff
@@ -0,0 +1,107 @@
...
+    @Override
+    protected IModule getTarget(String linkedModuleId) {
+        return moduleManager.getModule(linkedModuleId);
+    }
+
+    @Override
+    protected IModuleLink createLinkObject(String title, String id) {
+        ISpace source = spaceManager.getSpace(id);
+        IModuleLink link = moduleLinkFactory.createModuleLink(title, source);
+        return link;
```
jdamerow, line 60: merge with line 59

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ModuleLinkManagerTest.java
```diff
@@ -0,0 +1,121 @@
...
+        IModuleLink moduleLink = new ModuleLink();
+        moduleLink.setId("MOL001");
+        moduleLink.setSpace(space);
+        moduleLinkDisplay.setName("TestModule");
+        moduleLinkDisplay.setPositionX(10);
+        moduleLinkDisplay.setPositionY(30);
+        moduleLinkDisplay.setRotation(20);
+        moduleLinkDisplay.setType(DisplayType.MODULE);
+        moduleLinkDisplay.setLink(moduleLink);
+        Mockito.when(moduleLinkDisplayRepo.save(moduleLinkDisplay)).thenReturn(moduleLinkDisplay);;
```
jdamerow, line 86: remove second `;`

jdamerow:
Jenkins, test this please.

jdamerow:
When you edit a link, change the label, click "Edit Space " (this should be "Edit Link" btw), and then click on the link again to edit it, the old label is still in it. So if you change something else and click the "edit" button again, you change the label back to the previous value.

Also, let's put the delete link button part in the same modal with the edit form.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/LinkManager.java
```diff
@@ -1,85 +1,33 @@
...
+import edu.asu.diging.vspace.core.model.display.impl.LinkDisplay;
 import edu.asu.diging.vspace.core.model.impl.VSImage;
 import edu.asu.diging.vspace.core.services.ILinkManager;
-import edu.asu.diging.vspace.core.services.IModuleManager;
 import edu.asu.diging.vspace.core.services.ISpaceManager;
 
 @Transactional
-@Service
-public class LinkManager implements ILinkManager {
+public abstract class LinkManager<L extends ILink<T>,T extends IVSpaceElement, U extends LinkDisplay> implements ILinkManager<L, T, U>{
```
jdamerow, line 26: wouldn't that be `U extends ILinkDisplay`?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/LinkManager.java
```diff
@@ -89,182 +37,80 @@
...
+    public U createLink(String title, String id, float positionX, float positionY,
+            int rotation, String linkedId, String linkLabel, DisplayType displayType, byte[] linkImage,
+            String imageFilename) throws SpaceDoesNotExistException,ImageCouldNotBeStoredException, SpaceDoesNotExistException{
+
+        L link = createLinkObject(title, id);
+        T target = getTarget(linkedId);
+        link.setName(linkLabel);
+        link.setTarget(target);
+        U displayLink = createDisplayLink(link);
+        setDisplayProperties((ILinkDisplay) displayLink, positionX, positionY, rotation, displayType, linkImage, imageFilename);
```
jdamerow, line 50: and then you don't need to cast here

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/LinkManager.java
```diff
@@ -89,182 +37,80 @@
...
+            DisplayType displayType, byte[] linkImage, String imageFilename) throws SpaceDoesNotExistException, LinkDoesNotExistsException, ImageCouldNotBeStoredException{
+
+        validateSpace(id);
+
+        L link =  getLink(linkId);
+        T target = getTarget(linkedId);
+        link.setName(title);
+        link.setTarget(target);
+        U displayLink = getDisplayLink(linkDisplayId);
+        setDisplayProperties((ILinkDisplay) displayLink,positionX,positionY,rotation, displayType, linkImage, imageFilename);
```
jdamerow, line 67: or here

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ModuleLinkManager.java
```diff
@@ -0,0 +1,106 @@
...
+    protected IModuleLink getLink(String moduleLinkID){
+        Optional<ModuleLink> moduleLink = moduleLinkRepo.findById(moduleLinkID);
+        if(moduleLink.isPresent()) {
+            return moduleLink.get();
+        }
+        return null;
+    }
+
+    @Override 
+    protected ModuleLinkDisplay createDisplayLink(IModuleLink link) {
```
jdamerow, line 87: since the generics should extend the interface, you would return the interface here and wouldn't have to cast in line 88

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceLinkManager.java
```diff
@@ -0,0 +1,105 @@
...
+        return spaceLinkRepo.save((SpaceLink) link);
+    }
+
+    @Override
+    protected ISpace getTarget(String linkedSpaceId) {
+        return spaceManager.getSpace(linkedSpaceId);
+    }
+
+    @Override
+    protected SpaceLinkDisplay getDisplayLink(String spaceLinkDisplayId){
```
jdamerow, line 61: same here

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceLinkManager.java
```diff
@@ -0,0 +1,105 @@
...
+
+    @Override
+    protected SpaceLinkDisplay updateLinkAndDisplay(ISpaceLink link, SpaceLinkDisplay displayLink) {
+        spaceLinkRepo.save((SpaceLink) link);
+        return spaceLinkDisplayRepo.save(displayLink);
+    }
+
+    @Override
+    protected SpaceLinkDisplay createDisplayLink(ISpaceLink link){
+        return (SpaceLinkDisplay) spaceLinkDisplayFactory.createSpaceLinkDisplay(link);
```
jdamerow, line 87: same here

vspace/src/main/webapp/WEB-INF/views/staff/spaces/space.jsp
jdamerow: ?

vspace/src/main/webapp/WEB-INF/views/staff/spaces/space.jsp
jdamerow: ?

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SpaceLinkManagerTest.java
jdamerow: update, delete link test methods are still missing

jdamerow:
Jenkins, test this please.

jdamerow:
the page shouldn't reload after a link has been edited or deleted.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SpaceLinkManagerTest.java
jdamerow: ?

jdamerow:
Jenkins, test this please.

jdamerow:
on loading of home, staging throws:
```
java.lang.IllegalArgumentException: Can not set boolean field edu.asu.diging.vspace.core.model.impl.VSpaceElement.showUnpublishedLinks to null value
	sun.reflect.UnsafeFieldAccessorImpl.throwSetIllegalArgumentException(UnsafeFieldAccessorImpl.java:167)
	sun.reflect.UnsafeFieldAccessorImpl.throwSetIllegalArgumentException(UnsafeFieldAccessorImpl.java:171)
	sun.reflect.UnsafeBooleanFieldAccessorImpl.set(UnsafeBooleanFieldAccessorImpl.java:80)
	java.lang.reflect.Field.set(Field.java:764)
	org.hibernate.property.access.spi.SetterFieldImpl.set(SetterFieldImpl.java:41)
	org.hibernate.tuple.entity.AbstractEntityTuplizer.setPropertyValues(AbstractEntityTuplizer.java:649)
	org.hibernate.tuple.entity.PojoEntityTuplizer.setPropertyValues(PojoEntityTuplizer.java:205)
	org.hibernate.persister.entity.AbstractEntityPersister.setPropertyValues(AbstractEntityPersister.java:4899)
	org.hibernate.engine.internal.TwoPhaseLoad.doInitializeEntity(TwoPhaseLoad.java:190)
	org.hibernate.engine.internal.TwoPhaseLoad.initializeEntity(TwoPhaseLoad.java:129)
```

prjadhav14:
> on loading of home, staging throws:
> 
> ```
> java.lang.IllegalArgumentException: Can not set boolean field edu.asu.diging.vspace.core.model.impl.VSpaceElement.showUnpublishedLinks to null value
> 	sun.reflect.UnsafeFieldAccessorImpl.throwSetIllegalArgumentException(UnsafeFieldAccessorImpl.java:167)
> 	sun.reflect.UnsafeFieldAccessorImpl.throwSetIllegalArgumentException(UnsafeFieldAccessorImpl.java:171)
> 	sun.reflect.UnsafeBooleanFieldAccessorImpl.set(UnsafeBooleanFieldAccessorImpl.java:80)
> 	java.lang.reflect.Field.set(Field.java:764)
> 	org.hibernate.property.access.spi.SetterFieldImpl.set(SetterFieldImpl.java:41)
> 	org.hibernate.tuple.entity.AbstractEntityTuplizer.setPropertyValues(AbstractEntityTuplizer.java:649)
> 	org.hibernate.tuple.entity.PojoEntityTuplizer.setPropertyValues(PojoEntityTuplizer.java:205)
> 	org.hibernate.persister.entity.AbstractEntityPersister.setPropertyValues(AbstractEntityPersister.java:4899)
> 	org.hibernate.engine.internal.TwoPhaseLoad.doInitializeEntity(TwoPhaseLoad.java:190)
> 	org.hibernate.engine.internal.TwoPhaseLoad.initializeEntity(TwoPhaseLoad.java:129)
> ```

Issue fixed under story VSPC-113 and merged into this branch

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ExternalLinkManagerTest.java
```diff
@@ -0,0 +1,240 @@
...
+        externalLinkDisplay.setExternalLink(externalLink);
+        externalLinkDisplayList.add(externalLinkDisplay);
+        List<IExternalLinkDisplay> externalLinkObj = externalLinkDisplayList;
+        Mockito.when(externalLinkDisplayRepo.findExternalLinkDisplaysForSpace(space.getId())).thenReturn(externalLinkObj);
+        managerToTest.deleteLinkDisplayRepo(externalLink);
+        Mockito.verify(externalLinkDisplayRepo).deleteByExternalLink(externalLink);
+    }
+
+    @Test
+    public void test_deleteLinkRepo_linkPresent(){
```
jdamerow, line 154: your test methods should test the public methods. the protected methods don't really need tests

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ExternalLinkManagerTest.java
```diff
@@ -0,0 +1,240 @@
...
+        Mockito.when(spaceManager.getSpace(spaceId1)).thenReturn(space);
+        ExternalLink newExternalLink = new ExternalLink(); 
+        newExternalLink.setId("externalLink1");
+        Optional<ExternalLink> mockExternalLink = Optional.of((ExternalLink)externalLink);
+        Mockito.when(externalLinkRepo.findById(externalLink.getId())).thenReturn(mockExternalLink);
+
+        Optional<ExternalLinkDisplay> mockExternalLinkDisplay = Optional.of((ExternalLinkDisplay)externalLinkDisplay);
+        Mockito.when(externalLinkDisplayRepo.findById(externalLinkDisplay.getId())).thenReturn(mockExternalLinkDisplay);
+
+        Mockito.when(managerToTest.updateLinkAndDisplay(externalLink, externalLinkDisplay)).thenReturn(externalLinkDisplay);
```
jdamerow, line 192: whenever you write a mock for `managerToTest` you're probably doing something wrong.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ExternalLinkManagerTest.java
```diff
@@ -0,0 +1,240 @@
...
+        externalLink.setSpace(space);
+        externalLinkDisplay.setExternalLink(externalLink);
+        externalLinkDisplayList.add(externalLinkDisplay);
+        externalLinkDisplay = new ExternalLinkDisplay();
+        externalLink = new ExternalLink();
+        externalLink.setId("EXL002");
+        externalLink.setSpace(space);
+        externalLinkDisplay.setExternalLink(externalLink);
+        externalLinkDisplayList.add(externalLinkDisplay);
+        List<IExternalLinkDisplay> externalLinkObj = externalLinkDisplayList;
```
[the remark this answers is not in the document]
prjadhav14: we want to remove the link from the list so that it does not appear on the screen after deletion. Here I am mocking that behavior by adding the data to list allowing method to remove the link from the given space.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ExternalLinkManagerTest.java
```diff
@@ -0,0 +1,177 @@
...
+        Mockito.when(externalLinkRepo.save((ExternalLink) externalLink)).thenReturn((ExternalLink)externalLink);
+        Mockito.when(externalLinkDisplayRepo.save((ExternalLinkDisplay)externalLinkDisplay)).thenReturn((ExternalLinkDisplay)externalLinkDisplay);
+
+        IExternalLinkDisplay actualUpdatedLink = managerToTest.updateLink("TestExternalNew", spaceId1, 20, 40, 0, "www.google.com", "TestExternalNew", "EXL001", "EXLD001", DisplayType.ARROW, null, null);
+        Assert.assertEquals(externalLinkDisplay.getId(), actualUpdatedLink.getId());
+        Assert.assertEquals(externalLinkDisplay.getName(), actualUpdatedLink.getName());
+        Assert.assertEquals(new Double(externalLinkDisplay.getPositionX()), new Double(actualUpdatedLink.getPositionX()));
+        Assert.assertEquals(new Double(externalLinkDisplay.getPositionY()), new Double(actualUpdatedLink.getPositionY()));
+        Assert.assertEquals(externalLinkDisplay.getExternalLink().getId(), actualUpdatedLink.getExternalLink().getId());
+        Assert.assertEquals(externalLinkDisplay.getType(), actualUpdatedLink.getType());
```
jdamerow, line 137: what about the actual link address (stackoverflow.com vs google)?

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ModuleLinkManagerTest.java
```diff
@@ -0,0 +1,198 @@
...
+
+        IModuleLinkDisplay savedModuleLinkDisplay1 = managerToTest.createLink("New Module Link", spaceId1, 10, 30, 40, "MOD001", "New Module Link", DisplayType.ARROW, null, null);
+        Assert.assertEquals(moduleDisplayLink.getId(), savedModuleLinkDisplay1.getId());
+        Assert.assertEquals(moduleDisplayLink.getName(), savedModuleLinkDisplay1.getName());
+        Assert.assertEquals(new Double(moduleDisplayLink.getPositionX()), new Double(savedModuleLinkDisplay1.getPositionX()));
+        Assert.assertEquals(new Double(moduleDisplayLink.getPositionY()), new Double(savedModuleLinkDisplay1.getPositionY()));
+        Assert.assertEquals(moduleDisplayLink.getRotation(), savedModuleLinkDisplay1.getRotation());
+        Assert.assertEquals(moduleDisplayLink.getLink().getId(), savedModuleLinkDisplay1.getLink().getId());
+        Assert.assertEquals(moduleDisplayLink.getType(), savedModuleLinkDisplay1.getType());
+        Mockito.verify(moduleLinkDisplayRepo).save((ModuleLinkDisplay)moduleDisplayLink);
```
jdamerow, line 101: what about the module the link points?

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SpaceLinkManagerTest.java
```diff
@@ -0,0 +1,201 @@
...
+        Mockito.when(spaceLinkDisplayRepo.findById(spaceLinkDisplay.getId())).thenReturn(mockSpaceLinkDisplay);
+
+        managerToTest.deleteLink("SPL001");
+
+        Mockito.verify(spaceLinkDisplayRepo).deleteByLink(spaceLink);
+        Mockito.verify(spaceLinkRepo).delete((SpaceLink)spaceLink);
+    }
+
+    @Test
+    public void test_updateDisplayLink_success() throws SpaceDoesNotExistException, LinkDoesNotExistsException, ImageCouldNotBeStoredException {
```
jdamerow, line 142: this tests updateLink, not updateDisplayLink, right?

jdamerow:
Clicking "Edit Link" (let's name this just "Save"), throws a 500:
```
JaversException SNAPSHOT_STATE_VIOLATION: snapshots are immutable
	org.javers.core.metamodel.object.CdoSnapshotStateBuilder.withPropertyValue(CdoSnapshotStateBuilder.java:31)
	org.javers.core.json.typeadapter.commit.CdoSnapshotStateDeserializer.deserialize(CdoSnapshotStateDeserializer.java:39)
	org.javers.core.json.typeadapter.commit.CdoSnapshotTypeAdapter.deserializeSnapshotState(CdoSnapshotTypeAdapter.java:71)
	org.javers.core.json.typeadapter.commit.CdoSnapshotTypeAdapter.fromJson(CdoSnapshotTypeAdapter.java:57)
	org.javers.core.json.typeadapter.commit.CdoSnapshotTypeAdapter.fromJson(CdoSnapshotTypeAdapter.java:21)
	org.javers.core.json.JsonConverterBuilder.lambda$registerJsonTypeAdapterForType$4(JsonConverterBuilder.java:166)
	com.google.gson.internal.bind.TreeTypeAdapter.read(TreeTypeAdapter.java:69)
	com.google.gson.Gson.fromJson(Gson.java:888)
	com.google.gson.Gson.fromJson(Gson.java:953)
	com.google.gson.Gson.fromJson(Gson.java:926)
```

vspace/src/main/webapp/WEB-INF/views/staff/spaces/space.jsp
```diff
@@ -104,19 +155,35 @@ $( document ).ready(function() {
...
+		inputType.setAttribute("id", "externalLinkType-${link.externalLink.id}");
+		inputType.setAttribute("value", "${link.type}");
+		var inputExternalLinkURL = document.createElement("input");
+		inputExternalLinkURL.setAttribute("type", "hidden");
+		inputExternalLinkURL.setAttribute("id", "externalLinkTarget-${link.externalLink.id}");
+		inputExternalLinkURL.setAttribute("value", "${link.externalLink.externalLink}");
+		
+		$("#space").append(inputLabel);
+		$("#space").append(inputType);
+		$("#space").append(inputExternalLinkURL);
```
jdamerow, line 183: indentation

prjadhav14:
reopening to replicate 500 on save.

prjadhav14:
Reopening for the new deployment to debug 500 issue on space link edit

prjadhav14:
Reopening to deploy newer version of Javers

prjadhav14:
Reopening to test the latest 5.11.1 Javers

jdamerow:
Jenkins, test this please.

PR #129, opened 2020-06-18
Story/VSPC-94

prjadhav14:


jdamerow:
Jenkins, test this please.

jdamerow:
I'm just realizeing we don't have the backend work done yet, to be able to show branching points. Let's implement that first (I'll create a ticket) and then come back to this branch. One comment though: the menu that opens shows a tiny bit of the menu icon which makes it look odd:
image

vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSlideController.java
```diff
@@ -80,20 +81,25 @@ public String slide(Model model, @PathVariable("slideId") String slideId, @PathV
         String nextSlideId = "";
         String prevSlideId = "";
         int slideIndex = 0;
-        slideIndex = sequenceSlides.indexOf(sildeManager.getSlide(slideId));
+        ISlide currentSlide = sildeManager.getSlide(slideId);
+        slideIndex = sequenceSlides.indexOf(currentSlide);
```
jdamerow, line 85: merge with line 82

vspace/src/main/webapp/WEB-INF/views/exhibition/module.jsp
```diff
@@ -28,87 +28,158 @@ body {
     function closeLangNav() {
         document.getElementById("myLangSidenav").style.width = "0";
     }
+    function openNav(){
+            document.getElementById("mySidenav").style.width = "170px";
+        	document.getElementById("mySidenav").style.height = "700px";
```
jdamerow, line 33: indentation

jdamerow:
Jenkins, test this please.

jdamerow:
A few things:
- the buttons on the left need titles that are shown when hovering over them and they need to be closer together
image
- remove the border around the list component
image
- both buttons seem to bring me back to the beginning of the module
image

prjadhav14:
> A few things:
> 
>     * the buttons on the left need titles that are shown when hovering over them and they need to be closer together
>       image
> 
>     * remove the border around the list component
>       image
> 
>     * both buttons seem to bring me back to the beginning of the module
>       image

For last point both buttons showing same links -> The links is created using previous choiceID from which user has navigated to current branching choice. Now if, the slide is a fresh load there is no previous choice to create link and hence link to existing choice is provided.

jdamerow:
I have a module with the start sequence A-B-C, C being a branching point branching to a sequence D-E. Both buttons brought me back to A.

jdamerow:
But one button should bring me back to C.

prjadhav14:
> I have a module with the start sequence A-B-C, C being a branching point branching to a sequence D-E. Both buttons brought me back to A.

Right, when you navigate to D-E first button will take you back to A and second button to C, and when you are on C technically both previous and start sequence are same i.e A.

vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSlideController.java
```diff
@@ -79,21 +88,52 @@ public String slide(Model model, @PathVariable("slideId") String slideId, @PathV
...
+            String[] elementValues = seondLastElement.split(",");
+            model.addAttribute("previousChoiceSequence", elementValues[0]);
+            model.addAttribute("previousChoiceSlide", elementValues[1]);
+            choiceHistory.addToSequenceSlideHistory(seondLastElement);
+            choiceHistory.addToSequenceSlideHistory(lastElement);
+        }
+
+        if(back) {
+            choiceHistory.removeLastElementFromSequenceSlideHistory();
+        }
```
jdamerow, line 136: this all seems very complicate. wouldn't you just need this:
- in sequence controller:
   - pass parameter with branching point id that led to the sequence (if no branching point id is passed, check if sequence is start sequence, if not, find first branching point that links to the sequence)
  - push sequence/slide object onto stack
- in slide controller:
  - if there is a sequence/slide object on the stack, use that to generate links (only peek, no pop)
  - if back parameter exists and is true, pop last sequence/slide if slide is current branching point
- in branching point jsp:
  - add BP id to choice links
  - add back parameter to back to last BP button

vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSequencesController.java
```diff
@@ -54,12 +60,20 @@ public String sequence(Model model, @PathVariable("sequenceId") String sequenceI
...
-        
+
         List<ISlide> slides = sequenceManager.getSequence(sequenceId).getSlides();
         if (slides.size() == 0) {
             throw new SlidesInSequenceNotFoundException();
         }
+
+        if(choiceId==null) {
+            choiceId=sequenceId;
+            branchingPointId = sequenceManager.getSequence(sequenceId).getSlides().get(0).getId();
```
jdamerow, line 71: you should check if there are slides in the sequence first or you risk a null pointer or out of bounds exception

vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSequencesController.java
```diff
@@ -54,12 +60,20 @@ public String sequence(Model model, @PathVariable("sequenceId") String sequenceI
...
+        }
+        if(choice){
```
jdamerow, line 73: bad variable name

vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSlideController.java
```diff
@@ -79,21 +86,33 @@ public String slide(Model model, @PathVariable("slideId") String slideId, @PathV
...
         model.addAttribute("slides", sequenceSlides);
         model.addAttribute("currentSequenceId", sequenceId);
         model.addAttribute("nextSlide", nextSlideId);
         model.addAttribute("prevSlide", prevSlideId);
-        model.addAttribute("currentSlideCon", sildeManager.getSlide(slideId));
+        model.addAttribute("currentSlideCon", currentSlide);
+        if(currentSlide instanceof BranchingPoint) {
+            model.addAttribute("choices", ((BranchingPoint)currentSlide).getChoices());
+        }
+        if(back) {
```
jdamerow, line 108: you only want to do this if the current slide is a branching point, otherwise it is not possible that the user clicked the back to branching point button.

vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSlideController.java
```diff
@@ -79,21 +86,33 @@ public String slide(Model model, @PathVariable("slideId") String slideId, @PathV
...
+            choiceHistory.removeLastElementFromSequenceSlideHistory();
+        }
+        if(choiceHistory.getFromSequenceSlideHistory().size()>0) {
+            model.addAttribute("showBackToPreviousChoice", true);
+            model.addAttribute("previousChoiceId", choiceHistory.peekTopChoiceId());
```
jdamerow, line 113: this is the previous squence id, right? not the choice id?

vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSequencesController.java
jdamerow: what is this for?

vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSequencesController.java
```diff
@@ -54,12 +60,20 @@ public String sequence(Model model, @PathVariable("sequenceId") String sequenceI
...
         }
-        
+
         List<ISlide> slides = sequenceManager.getSequence(sequenceId).getSlides();
         if (slides.size() == 0) {
             throw new SlidesInSequenceNotFoundException();
         }
+
+        if(choiceId==null) {
+            choiceId=sequenceId;
```
jdamerow, line 70: so choiceId is actually sequence id? bad variable name

vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSequencesController.java
jdamerow, line 73: just check if there is a branching point id

vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSequencesController.java
prjadhav14: This check is implemented one step above this block and exception is getting thrown, I will use the same list variable to get the first slide in the sequence.

vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSequencesController.java
prjadhav14: If the user clicks on the choice from branching point slide we are checking the choice flag and pushing the current choice id and branching point id to the stack.

vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSequencesController.java
jdamerow: ah, right.

vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSequencesController.java
jdamerow: this

vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSlideController.java
prjadhav14: The user will be navigating to the next sequence from choice(choiceID is nothing but ID linked to sequenceID). previouschoice made more sense to me as we say we are linking choices to branching points and using those to navigate to the linked sequences from branching points. your thoughts?

vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSequencesController.java
prjadhav14: Yes, but we want to store currentSequenceId and we lose that info when we click on choice from branching point. currentSequenceID, sequenceID would be confusing so I thought of using choiceID. Should I call that currentSequenceID?

prjadhav14: ignore, removed this block of code.

vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSequencesController.java
prjadhav14: removing this flag, not required :-)

vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSlideController.java
prjadhav14: current slide can be a branching point if the user clicks on choice and the first slide of that choice is a branching point?

jdamerow: what I mean is:
- this controller is called when the new slide page is rendered, right? 
- if back is true, then the user came from another sequence and clicked the 'back to bp' button
- the back to bp button will always bring the user back the the previous bp
- so if the current slide is not a bp, then there was an error somewhere, because that can't be.

vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSlideController.java
jdamerow: previous choice id sounds like you're storing the id of the click branch point choice, but that's not what you're storing right? the id will literally hold the id of the previous sequence, right? so that's what I would call that.

vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSequencesController.java
```diff
@@ -54,12 +61,15 @@ public String sequence(Model model, @PathVariable("sequenceId") String sequenceI
...
             throw new SequenceNotFoundException(sequenceId);
         }
-        
+
         List<ISlide> slides = sequenceManager.getSequence(sequenceId).getSlides();
         if (slides.size() == 0) {
             throw new SlidesInSequenceNotFoundException();
         }
+        if(branchingPointId!=null){
+            choiceHistory.addToSequenceSlideHistory(choiceId,branchingPointId);
```
jdamerow, line 70: same as below, choice id is the previous sequence id, right?

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/ChoicesHistory.java
```diff
@@ -0,0 +1,51 @@
...
+import org.springframework.context.annotation.Scope;
+import org.springframework.context.annotation.ScopedProxyMode;
+import org.springframework.stereotype.Component;
+
+@Component
+@Scope(value="session", proxyMode = ScopedProxyMode.TARGET_CLASS)
+public class ChoicesHistory{
+
+    class ChoicesHistoryData{
+        String choiceId;
```
jdamerow, line 14: sequence id, rigth?

jdamerow: should be adjusted through out this file

vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSequencesController.java
```diff
@@ -54,12 +61,15 @@ public String sequence(Model model, @PathVariable("sequenceId") String sequenceI
...
         }
-        
+
         List<ISlide> slides = sequenceManager.getSequence(sequenceId).getSlides();
         if (slides.size() == 0) {
             throw new SlidesInSequenceNotFoundException();
         }
+        if(branchingPointId!=null){
+            choiceHistory.addToSequenceSlideHistory(previousSequenceId,branchingPointId);
+        }
```
jdamerow, line 71: I would put this logic into the slide controller. otherwise you assume that the sequence controller will always be called first and then redirect but you can't guarantee that. if we change the implementation to directly link to the branching point, the history function will fail.

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/ChoicesHistory.java
```diff
@@ -0,0 +1,51 @@
...
+
+    }
+
+    private Stack<ChoicesHistoryData> sequenceSlideHistory = new Stack<>();
+
+    public void addToSequenceSlideHistory(String choiceId, String branchingPointId) {
+        sequenceSlideHistory.push(new ChoicesHistoryData(choiceId, branchingPointId));
+    }
+
+    public boolean backNavigationExists() {
```
jdamerow, line 35: just call that "hasHistory". much easier to understand. also, methods that return booleans typically start with has or is.

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/ChoicesHistory.java
```diff
@@ -0,0 +1,51 @@
...
+        }
+        public String getBranchingPointId() {
+            return branchingPointId;
+        }
+
+    }
+
+    private Stack<ChoicesHistoryData> sequenceSlideHistory = new Stack<>();
+
+    public void addToSequenceSlideHistory(String choiceId, String branchingPointId) {
```
jdamerow, line 31: just call that `addToHistory`. there is no other add method that adds anything.

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/ChoicesHistory.java
```diff
@@ -0,0 +1,51 @@
...
+
+    public void removeLastElementFromSequenceSlideHistory() {
+        sequenceSlideHistory.pop();
+    }
+
+    public String peekTopBranchingPointId() {
+        return sequenceSlideHistory.peek().getBranchingPointId();
+    }
+
+    public String peekTopChoiceId() {
```
jdamerow, line 47: peek implies it's the top element (no need for 'Top'). same in the above method.

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/ChoicesHistory.java
```diff
@@ -0,0 +1,51 @@
...
+
+    public void addToSequenceSlideHistory(String choiceId, String branchingPointId) {
+        sequenceSlideHistory.push(new ChoicesHistoryData(choiceId, branchingPointId));
+    }
+
+    public boolean backNavigationExists() {
+        return !sequenceSlideHistory.isEmpty();
+    }
+
+    public void removeLastElementFromSequenceSlideHistory() {
```
jdamerow, line 39: just call this `pop` or `popFromHistory` or something. This is unnecessarily wordy.

vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSlideController.java
```diff
@@ -79,21 +85,33 @@ public String slide(Model model, @PathVariable("slideId") String slideId, @PathV
...
         model.addAttribute("slides", sequenceSlides);
         model.addAttribute("currentSequenceId", sequenceId);
         model.addAttribute("nextSlide", nextSlideId);
         model.addAttribute("prevSlide", prevSlideId);
-        model.addAttribute("currentSlideCon", sildeManager.getSlide(slideId));
+        model.addAttribute("currentSlideCon", currentSlide);
+        if(currentSlide instanceof BranchingPoint) {
+            model.addAttribute("choices", ((BranchingPoint)currentSlide).getChoices());
+            if(back) {
+                choiceHistory.removeLastElementFromSequenceSlideHistory();
```
jdamerow, line 107: you might want to make sure that branching point and sequence are the same as the current slide ones. otherwise, if someone just attached the back parameter to a random branching point or it gets accidentally passed to it, this would cause the history to get out of sync.

vspace/src/main/webapp/WEB-INF/views/exhibition/module.jsp
```diff
@@ -28,87 +28,147 @@ body {
...
+                    test="${contents['class'].simpleName == 'ChoiceBlock'}">
+                    <div id="${contents.id}"
+                        class="textDiv"
+                        style="margin: 1%;">
+                        <c:if test="${contents.showsAll eq true}">
+                            <div class="list-group">
+                                <c:forEach items="${choices}"
+                                    var="choice">
+                                    <a
+                                        href="<c:url value="/exhibit/${spaceId}/module/${module.id}/sequence/${choice.sequence.id}/slide/${choice.sequence.slides[0].id}?previousSequenceId=${currentSequenceId}&branchingPoint=${slideId}" />"
```
jdamerow, line 154: i would still link the sequence here as you had it before, just pass the parameters through to the slide redirect; if you do it this way, you will have to deal with empty sequences here instead of in the controller (where you should deal with it).

vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSlideController.java
```diff
@@ -79,21 +88,36 @@ public String slide(Model model, @PathVariable("slideId") String slideId, @PathV
...
         model.addAttribute("sequence",sequenceExist);
         model.addAttribute("slides", sequenceSlides);
         model.addAttribute("currentSequenceId", sequenceId);
         model.addAttribute("nextSlide", nextSlideId);
         model.addAttribute("prevSlide", prevSlideId);
-        model.addAttribute("currentSlideCon", sildeManager.getSlide(slideId));
+        model.addAttribute("currentSlideCon", currentSlide);
+        if(currentSlide instanceof BranchingPoint) {
+            model.addAttribute("choices", ((BranchingPoint)currentSlide).getChoices());
+            if(back && sequenceHistory.peekSequenceId().equalsIgnoreCase(sequenceId)) {
```
jdamerow, line 109: I think I would either test for both (sequence id and bp id) or just the bp id, but this case will fail if there are multiple bps in a sequence (and if the parameters are accidentally passed).

jdamerow:
as discussed on slack. afterwards it's probably good to go!

vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSequencesController.java
```diff
@@ -63,6 +63,11 @@ public String sequence(Model model, @PathVariable("sequenceId") String sequenceI
             throw new SlidesInSequenceNotFoundException();
         }
         String firstSlideId = slides.get(0).getId();
-        return "redirect:/exhibit/{spaceId}/module/" + moduleId + "/sequence/" + sequenceId + "/slide/" + firstSlideId +"?branchingPoint="+branchingPointId+"&previousSequenceId="+previousSequenceId;
+        
+        if(branchingPointId==null) {
+            return "redirect:/exhibit/{spaceId}/module/" + moduleId + "/sequence/" + sequenceId + "/slide/" + firstSlideId;
+        }else {
+            return "redirect:/exhibit/{spaceId}/module/" + moduleId + "/sequence/" + sequenceId + "/slide/" + firstSlideId +"?branchingPoint="+branchingPointId+"&previousSequenceId="+previousSequenceId;
+        }
```
jdamerow, line 71: how about 
```
...?branchingPoint="+(branchingPointId != null ? branchingPointId : "") + ...
```
or even better use `String.format`

prjadhav14: If we do this, the null check will again fail in the slide controller because this will lead parameter to take "" value in slide controller.

prjadhav14: Let me check how it works if i pass null to slideController for null in sequenceController

vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSequencesController.java
```diff
@@ -54,12 +57,15 @@ public String sequence(Model model, @PathVariable("sequenceId") String sequenceI
...
-        
+
         List<ISlide> slides = sequenceManager.getSequence(sequenceId).getSlides();
         if (slides.size() == 0) {
             throw new SlidesInSequenceNotFoundException();
         }
         String firstSlideId = slides.get(0).getId();
-        return "redirect:/exhibit/{spaceId}/module/" + moduleId + "/sequence/" + sequenceId + "/slide/" + firstSlideId;
+
+        return String.format("redirect:/exhibit/{spaceId}/module/" + moduleId + "/sequence/" + sequenceId + "/slide/" + firstSlideId + "?branchingPoint=%s&previousSequenceId=%s",
```
jdamerow, line 67: the point of using string.format is to get rid of all these concatenations; all the variables should be input to the format function (replace them with %s).

vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSlideController.java
```diff
@@ -79,21 +88,36 @@ public String slide(Model model, @PathVariable("slideId") String slideId, @PathV
...
         model.addAttribute("prevSlide", prevSlideId);
-        model.addAttribute("currentSlideCon", sildeManager.getSlide(slideId));
+        model.addAttribute("currentSlideCon", currentSlide);
+        if(currentSlide instanceof BranchingPoint) {
+            model.addAttribute("choices", ((BranchingPoint)currentSlide).getChoices());
+            if(back && sequenceHistory.peekBranchingPointId().equalsIgnoreCase(slideId)) {
+                sequenceHistory.popFromHistory();
+            }
+        }
+        if(branchingPointId!="" && branchingPointId!=null){
```
jdamerow, line 113: this is not how you check if branchingPointId is empty string; also first check for null then for empty string or you get a nullpointer if it's null.

PR #135, opened 2020-07-15
Story/vspc 49

Sagar-k93:


jdamerow:
Jenkins, test this please.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/SpaceManagerTest.java
```diff
@@ -168,13 +170,13 @@ public void test_deleteSpaceById_whenSpaceHasLinks() {
...
     public void test_deleteSpaceById_whenLinksToSpace() {  
         Space space = new Space();
-        spaceLink = new SpaceLink();
         space.setId(spaceId1);
         spaceLink.setId(spaceLinkId1);
         spaceLink.setTargetSpace(space);
         Mockito.when(spaceLinkRepo.getLinkedFromSpaces(spaceId1)).thenReturn(Arrays.asList(spaceLink));
         managerToTest.deleteSpaceById(spaceId1);
         Mockito.verify(spaceLinkRepo).deleteBySourceSpaceId(spaceId1);
+        Mockito.verify(spaceLink).setTargetSpace(null);
```
jdamerow, line 179: I don't think this will work since space link is not mocked.

Sagar-k93: I set it to mocked and checked.

jdamerow: no need to mock it just assert targetspace is null for that object

jdamerow:
Jenkins, test this please.

jdamerow:
Jenkins, test this please.

vspace/src/main/webapp/WEB-INF/views/staff/spaces/spacelist.jsp
```diff
@@ -13,14 +20,54 @@
...
+                    Are you sure you want to delete ?
+                </p>
+                <div id="warningMessage">
+                <div class="text-danger">
+                Other spaces have links to this space!
+                </div>
+                </div>
+                <div id="exhibitionMessage">
+                <div class="text-danger">
+                This space is the exhibition start space! This will set exhibition to null.
```
jdamerow, line 55: I think I made a comment about this warning message in the previous pull request. Also, I've changed the font size for the messages here, but it also needs to be changed on the space.jsp page.

jdamerow:
There is a bug somewhere:
- i created a new space
- I linked from another space to the new space
- when I clicked the delete button on the space list view it did not show me a warning (which it should in this case becuase there was an incoming link); I did not delete the space.
- I added a link from the new space to the other space. I went to the space list view and when I click on the delete button, nothing happens.

maybe it has soemthing to do with the fact that I created space B, linked from space A to space B and then created another link from space B to space A (so there is a loop).

jdamerow:
there is still a bug. I do not get the message that the space is linked form another space (same space as before A links to B, B links to A, and I'm trying to remove B).

jdamerow:
Jenkins, test this please.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceManager.java
```diff
@@ -1,207 +1,256 @@
...
+                exhibitRepo.save(exhibition);
+            }
+            // When space has no other links attached to it
+            spaceDisplayRepo.deleteBySpaceId(id);
+            spaceRepo.deleteById(id);
+        }
+    }
+
+    @Override
+    public List<SpaceLink> getLinkedSpace(String id) {
```
jdamerow, line 246: should be plural since it returns a list

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceManager.java
```diff
@@ -1,207 +1,256 @@
...
+
+        return spaceLinkRepo.getLinkedSpaces(id);
+    }
+    
+    @Override
+    public List<SpaceLink> getLinkedFromSpace(String id) {
```
jdamerow, line 252: same here

PR #141, opened 2020-07-30
[VSPC-102] Fixed access denied message when logged in.

Sagar-k93:
When a space is unpublished, a broken image is shown on the public page and the access denied message, and the yellow warning box, but no message. When logged in, it should show the space and the warning but not the access denied message.


jdamerow:
Jenkins, test this please.

PR #143, opened 2020-08-03
Story/VSPC 107

prjadhav14:
These changes will show the previous Branching Point name when the user hovers over Back to the previous Branching Choice button on public slide screens.

### DIFF
--- a/vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSlideController.java
+++ b/vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSlideController.java
@@ -119,7 +119,9 @@ public class ExhibitionSlideController {
         if(sequenceHistory.hasHistory()) {
             model.addAttribute("showBackToPreviousChoice", true);
             model.addAttribute("previousSequenceId", sequenceHistory.peekSequenceId());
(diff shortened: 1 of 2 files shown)


vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSlideController.java
```diff
@@ -119,7 +119,9 @@ public String slide(Model model, @PathVariable("slideId") String slideId, @PathV
         if(sequenceHistory.hasHistory()) {
             model.addAttribute("showBackToPreviousChoice", true);
             model.addAttribute("previousSequenceId", sequenceHistory.peekSequenceId());
-            model.addAttribute("previousBranchingPointId", sequenceHistory.peekBranchingPointId());
+            String previousBranchPointId = sequenceHistory.peekBranchingPointId();
+            model.addAttribute("previousBranchingPointId", previousBranchPointId);
+            model.addAttribute("previousBranchingPointName", ((BranchingPoint)slideManager.getSlide(previousBranchPointId)).getName());
```
jdamerow, line 124: just put the branching point in the model and then use it in the jsp otherwise if we later want to add more properties of the branching point to the page, we'll always have to to change two files the controller and the jsp (instead of just the jsp).

jdamerow:
Jenkins, test this please.

PR #144, opened 2020-08-03
Story/VSPC-104

prjadhav14:
Fixed null pointer issue while adding slide by removing default selected blank value for the Slide Type selection criterion.


jdamerow:
Jenkins, test this please.

PR #150, opened 2020-08-07
story/VSPC 105

prjadhav14:
When the user is logged in as staff and navigates to the Slide screen through the Module screen. The slide configuration screen should show the list of sequences it belongs to. Also, update sequences list when user cklicks on Slide and adding show all link to display all sequences in the active module.


jdamerow:
Jenkins, test this please.

jdamerow:
Well.. that needs some layouting... ;) but let's do that with Adhvik's new layout. However, the module overview that lists all slides needs to show that info too though.

jdamerow:
the slide cards on the module page need to show the sequences they are in. It could be just an icon that indicate a slide is in a sequence and then when you click on the icon it shows a popover with the list of sequences, but the user shouldn't have to navigate away from the page. Also, the branching point icon seems misplaced now:
image

jdamerow:
Ok, let's leave it like that for now and then see what Adhvik comes up with :)

PR #151, opened 2020-08-07
Story/vspc 95

Sagar-k93:
Only the slide content should scroll. The header and links on the sides should stay where they are so that the user can scroll to the end of the slide and just click the next or prev button without having to scroll up again.


jdamerow:
Jenkins, test this please.

jdamerow:
- the heading is cut off a litte at the top:
image
- the "back to previous branching point" icon moves:
image
image

jdamerow:
The next button is in the wrong place now (top right corner)
image

jdamerow:
one more change since you're already working on this. would you please align the prev/next buttons so that they are at the same height? right now, the next button is further down the the prev button.
image

jdamerow:
Jenkins, test this please.

jdamerow:
there is still something off, the module name is now cut off a little at the top and the popover of back to start is below the next button:
image
image

jdamerow:
Make it so, Jenkins.

jdamerow:
there is still something off on small screens:
image

jdamerow:
Make it so, Jenkins

jdamerow:
Make it so, Jenkins.

jdamerow:
there are still issues on small screens
image

jdamerow:
- the font seems to be a different now for modules and spaces
- on a phone, it looks like this:
   - the two buttons on the left should not overlap
image
   - there needs to be a little bit more space between the back/forward links and the text (just a few pixel is enough):
image

PR #152, opened 2020-08-11
Story/vspc 112

Sagar-k93:
The list of spaces should indicate if a space doesn't have any incoming links. There should be an icon which shows a message that the space does not have any incoming links.

### DIFF
--- a/vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/Space.java
+++ b/vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/Space.java
@@ -9,6 +9,7 @@ import javax.persistence.GeneratedValue;
 import javax.persistence.Id;
 import javax.persistence.OneToMany;
 import javax.persistence.OneToOne;
+import javax.persistence.Transient;
 
 import org.hibernate.annotations.GenericGenerator;
 import org.hibernate.annotations.NotFound;
@@ -53,6 +54,9 @@ public class Space extends VSpaceElement implements ISpace {
     @OneToOne(targetEntity=VSImage.class)
     @NotFound(action=NotFoundAction.IGNORE)
     private IVSImage image;
+    
+    @Transient
+    private Boolean hasIncomingLinks;
 
     /* (non-Javadoc)
      * @see edu.asu.diging.vspace.core.model.impl.ISpacee#getId()
@@ -155,4 +159,13 @@ public class Space extends VSpaceElement implements ISpace {
         }
         this.showUnpublishedLinks = showUnpublishedLinks;
     }
+
+    public Boolean getHasIncomingLinks() {
+        return hasIncomingLinks;
+    }
+
+    public void setHasIncomingLinks(Boolean hasIncomingLinks) {
+        this.hasIncomingLinks = hasIncomingLinks;
+    }
+    
 }
--- a/vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/Space.java
+++ b/vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/Space.java
@@ -56,7 +56,7 @@ public class Space extends VSpaceElement implements ISpace {
     private IVSImage image;
     
     @Transient
-    private Boolean hasIncomingLinks;
+    private Boolean incomingLinks;
 
     /* (non-Javadoc)
      * @see edu.asu.diging.vspace.core.model.impl.ISpacee#getId()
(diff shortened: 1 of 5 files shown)


vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/Space.java
```diff
@@ -53,6 +54,9 @@
     @OneToOne(targetEntity=VSImage.class)
     @NotFound(action=NotFoundAction.IGNORE)
     private IVSImage image;
+    
+    @Transient
+    private Boolean hasIncomingLinks;
```
jdamerow, line 59: java naming conventions for booleans is that the getter would be "is..." or sometimes "has..." or "can...". This won't be possible with this variable name. In other words, this needs to be renamed along with the getter/setter ;)

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ListSpacesController.java
```diff
@@ -1,32 +1,45 @@
...
     public String listSpaces(Model model) {
-
-        model.addAttribute("spaces", spaceRepo.findAll());
+        
+        Iterable<Space> allSpaces = spaceRepo.findAll();
+        Iterator<Space> iterator = allSpaces.iterator();
+        while(iterator.hasNext()) {
+            Space space = iterator.next();
+            space.setHasIncomingLinks( (spaceLinkRepo.getLinkedFromSpaces(space.getId())).size() > 0 ? true : false );
+        }
```
jdamerow, line 36: business logic that needs to happen in manager class

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceManager.java
```diff
@@ -253,4 +254,14 @@ public void deleteSpaceById(String id) {
 
         return spaceLinkRepo.getLinkedFromSpaces(id);
     }
+    
+    @Override
+    public Iterable<Space> getAllSpacesWithIncomingLinkInfo(Iterable<Space> spaceList) {
+        Iterator<Space> iterator = spaceList.iterator();
+        while(iterator.hasNext()) {
+            Space space = iterator.next();
+            space.setIncomingLinks( (spaceLinkRepo.getLinkedFromSpaces(space.getId())).size() > 0 ? true : false );
```
jdamerow, line 263: merge with line 262

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceManager.java
```diff
@@ -253,4 +254,14 @@ public void deleteSpaceById(String id) {
 
         return spaceLinkRepo.getLinkedFromSpaces(id);
     }
+    
+    @Override
+    public Iterable<Space> getAllSpacesWithIncomingLinkInfo(Iterable<Space> spaceList) {
```
jdamerow, line 259: either change the name or don't give it a list of spaces, but right now you say "getAllSpaces" but also give it a list of spaces. What is it all spaces or just the passed ones?

jdamerow:
Make it so, Jenkins.

jdamerow:
The icon is shown for the spaces that do have incoming links, not for the ones that don't.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceManager.java
Sagar-k93: Changed the name. THe spaceList was a list of all spaces. Changed that too allSpaces as well.

vspace/src/main/java/edu/asu/diging/vspace/core/services/ISpaceManager.java
```diff
@@ -28,5 +29,7 @@
     void deleteSpaceById(String id) throws SpaceDoesNotExistException;
     
     List<SpaceLink> getIncomingLinks(String id);
+    
+    Iterable<Space> addIncomingLinkInfoToAllSpaces(Iterable<Space> allSpaces);
```
jdamerow, line 33: still the same issue, is it all spaces or the passed spaces (which don't have to be all of them)?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceManager.java
```diff
@@ -253,4 +254,14 @@ public void deleteSpaceById(String id) {
 
         return spaceLinkRepo.getLinkedFromSpaces(id);
     }
+    
+    @Override
+    public Iterable<Space> addIncomingLinkInfoToAllSpaces(Iterable<Space> allSpaces) {
```
jdamerow, line 259: calling your list "allSpaces" is confusing, why would it have to be all?

jdamerow:
Make it so, Jenkins.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceManager.java
Sagar-k93: It is all spaces, and for each space that we display in the space list page ( which is all the spaces ), I add the boolean for if they have  incoming links or not.

vspace/src/main/java/edu/asu/diging/vspace/core/services/ISpaceManager.java
Sagar-k93: I am passing all the spaces to this method. That's why I named it allSpaces.

jdamerow:
Make it so, Jenkins.

PR #154, opened 2020-08-14
[VSPC-83] Changed css for the slide blocks on module page.

Sagar-k93:
On modules page, if the description of a slide is too long it'll overflow. Changed to show scroll and data is contained within the block.


jdamerow:
Make it so, Jenkins.

jdamerow:
if there isn't much text, it should use up so much space:
image

jdamerow:
Make it so, Jenkins.

PR #157, opened 2020-08-27
story/VSPC 87

prjadhav14:
This development will add a virtual-space specific 404 image if the space configured for the exhibition is an unpublished space.


vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSpaceController.java
```diff
@@ -70,6 +70,7 @@ public String space(@PathVariable("id") String id, Model model) {
             model.addAttribute("alertType", "danger");
             model.addAttribute("messageType","invalidSpace");
             model.addAttribute("message", "Access Denied.");
+            model.addAttribute("isExhibition", true);
```
jdamerow, line 73: Instead of this (why is it called `isExhibition` by the way?), I would forward to a 404 page.

prjadhav14: ok, the page should contain just image or header and footer as well?

jdamerow: the user needs to be able to get back to the actual exhibition, so the header should be there. and the footer should be on every page.

prjadhav14: calling this isExhibition because this property will be only set for exhibition screen.

jdamerow: this should be unnecessary once you forward to a dedicated page

vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSpaceController.java
```diff
@@ -66,10 +66,7 @@ public String space(@PathVariable("id") String id, Model model) {
             model.addAttribute("externalLinkList", externalLinkManager.getLinkDisplays(id));  
         }
         else {
-            model.addAttribute("showAlert", true);
-            model.addAttribute("alertType", "danger");
-            model.addAttribute("messageType","invalidSpace");
-            model.addAttribute("message", "Access Denied.");
+            return "badrequest";
```
jdamerow, line 69: I would redirect to `/404` or `/403`. There will be other places where we wanna show that and having a dedicated page for that makes sense. Plus, Spring security tries to redirect to `/403` I believe and right now that just generates a 404 exception, so it would be nice to have anyways.

jdamerow:
Make it so, Jenkins.

jdamerow:
Let's make the image a little smaller and centered. Right now it doesn't fit on a laptop screen. Also, there should be a message somewhere saying "The page you were looking for could not be found."

jdamerow:
image

vspace/src/main/java/edu/asu/diging/vspace/web/HomeController.java
```diff
@@ -34,9 +34,14 @@ public String home(Model model) {
...
     }
-    
+
     @RequestMapping(value = "/403") 
     public String accessDenied() {
-    	return "accessDenied";
+        return "accessDenied";
+    }
+
+    @RequestMapping(value="/exhibit/404")
```
jdamerow, line 43: last change, let's give this two urls, the current one and just `/404`.

jdamerow (changes requested):
after that should be good to go

vspace/src/main/java/edu/asu/diging/vspace/web/HomeController.java
```diff
@@ -34,9 +34,19 @@ public String home(Model model) {
...
+    public String error404(){
+        return "badrequest";
+    }
+
+    @RequestMapping(value="/404")
```
jdamerow, line 48: value also takes a list, just adjust the requestmapping in line 43.

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
`/404` needs authentication which kind of defeats the purpose ;)

jdamerow:
Make it so, Jenkins.

PR #166, opened 2020-11-23
[VSPC-139]

adhvikmadhav:


jdamerow:
Make it so, Jenkins.

jdamerow:
I really like the pop-up version, a few notes:
- let's make the right panel a lot less spacey, so the image can be bigger
image
(smaller buttons, less white space, smaller input fields, etc)
- the top part needs to be a lot less spacey (less white space between elements, smaller input fields/drop downs, etc)
image
- let's try to find a different solution for the incoming and outgoing links, they should be visiable alongside the image to be useful, or maybe above the image in an expandable box or something.
image
- and let's put the delete button on the bottom right corner

jdamerow:
Make it so, Jenkins.

jdamerow:
- please make the right side less spacey and less wide, so there is more space for the image (this is true for all pop ups)
image
- there are more links on the pop up than on the image on the space page
image
image
- after "save" has been click to add a link, let's show a spinner until the save went through and the pop up closes (right now it feels like it doesn't work)
- let's make the text smaller:
image
- let's move that all up a bit
image
- let's make these buttons a little smaller, and maybe more like regular buttons or like links
image
- and let's move the image/buttons up a little (reduce whitespace)
- not sure what to do with the incoming/outgoing links; the list should be shown by default (could be collapsible) so on the top would not be a good position (also they need to show the icons they should before (if they are published or not)). maybe to the left of the image in a collapsible panel that collapses to the left?
- delete button should be a little smaller (more like the other buttons)

jdamerow:
please make pr to 120 branch, so I see what has actually been changed

PR #167, opened 2020-12-02
Story/vspc 55

prjadhav14:
This ticket is about filtering the image list using the selected image category. This is achieved using jQuery and a hidden column in the table containing image category values.


jdamerow:
This only filters the currently visible data in the table, doesn't it? Not all data? So if I have 5 pages it will only show me the results for a specific category that are on that one page, not all? It should filter all data (make a call to the server).

prjadhav14:
> This only filters the currently visible data in the table, doesn't it? Not all data? So if I have 5 pages it will only show me the results for a specific category that are on that one page, not all? It should filter all data (make a call to the server).

Okay, is pagination expected if we give a server call and get more than 10 items on the list?

jdamerow:
yes, you shouldn't have to change much, except the method that is called to retrieve the data (but with the same page, etc.)

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ListImagesController.java
```diff
@@ -65,4 +73,45 @@ public String listSpaces(@PathVariable(required = false) String page,
...
         return "staff/images/imagelist";
     }
+    
+    @RequestMapping("/staff/images/list/tag/{filter}/{page}")
+    public String listSpacesByTag(@PathVariable(required = false) String page,
+            @PathVariable(required = false) String filter,
+            @RequestParam(value = "sort", required = false) String sortedBy,
+            @RequestParam(value = "order", required = false) String order, Model model) {
+        int pageNo;
+        page = StringUtils.isEmpty(page) ? "1" : page;
```
jdamerow, line 83: switch with line 82 (declare variables as close to where they are needed as possible)

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ListImagesController.java
```diff
@@ -65,4 +73,45 @@ public String listSpaces(@PathVariable(required = false) String page,
...
+        try {
+            pageNo = imageService.validatePageNumber(Integer.parseInt(page));
+        } catch (NumberFormatException numberFormatException){
+            pageNo = 1;
```
jdamerow, line 87: log a warning here

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ListImagesController.java
```diff
@@ -65,4 +73,45 @@ public String listSpaces(@PathVariable(required = false) String page,
...
+        }
+        model.addAttribute("totalPages", imageService.getTotalPages());
+        model.addAttribute("currentPageNumber", pageNo);
+        model.addAttribute("totalImageCount", imageService.getTotalImageCount());
+        List<IVSImage> imagesList = imageService.getImages(pageNo, sortedBy, order);
```
jdamerow, line 92: you should only get the images from the db that have a certain category (your total image count is wrong here too, isn't it?). Also, this should probably happen in the `listSpaces` method and pass the filter as query parameters. A lot of duplicate code here.

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
```diff
@@ -116,8 +127,15 @@ public ImageData getImageDimensions(IVSImage image, int width, int height) {
...
+        return sortingParameters;
+    }
+
+    @Override
+    public List<IVSImage> getImagesByCategory(int pageNo, ImageCategory category, String sortedBy, String order) {
+        Sort sortingParameters = getSortingParameters(sortedBy, order);
+        pageNo = validatePageNumber(pageNo);
         Pageable sortByRequestedField = PageRequest.of(pageNo - 1, pageSize, sortingParameters);
-        Page<VSImage> images = imageRepo.findAll(sortByRequestedField);
+        Page<IVSImage> images = imageRepo.findByCategoriesLike(sortByRequestedField, category);
```
jdamerow, line 138: Is there a reason that you do a like query instead of the regular query?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
```diff
@@ -135,6 +153,22 @@ public long getTotalPages() {
...
     }
 
+    /**
+     * Method to return the total image count
+     * 
+     * @return total count of images in DB
+     */
+    @Override
+    public long getTotalImageCountForSelectedFilter(ImageCategory category) {
+        return imageRepo.findByCategoriesLike(category).size();
```
jdamerow, line 163: you can do a `countBy` query and let the db do the counting

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
```diff
@@ -135,6 +153,22 @@ public long getTotalPages() {
...
+    }
+
+    @Override
+    public long getTotalPagesForSelectedFilter(ImageCategory category) {
+        int count = imageRepo.findByCategoriesLike(category).size();
```
jdamerow, line 168: same as above

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
```diff
@@ -135,6 +153,22 @@ public long getTotalPages() {
...
+        return (count % pageSize==0) ? count / pageSize:(count / pageSize) + 1;
```
jdamerow, line 169: let's change the spacing so it's clear what belongs together
`coutn/pageSize : (count/pageSize)+1;`

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ListImagesController.java
```diff
@@ -30,12 +30,14 @@
     private ISpaceManager spaceManager;
 
     @RequestMapping("/staff/images/list")
-    public String listSpacesWithoutNum(Model model) {
-        return "redirect:/staff/images/list/1";
+    public String listSpacesWithoutNum(Model model, @RequestParam(value = "imageCat", required = false) String imageCategory) {
+
+        return String.format("redirect:/staff/images/list/1?imageCat=%s",(imageCategory != null ? imageCategory : "Show All"));
```
jdamerow, line 35: let's use `all` instead of `Show all` since it's a parameter

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ListImagesController.java
```diff
@@ -48,7 +50,19 @@ public String listSpaces(@PathVariable(required = false) String page,
         model.addAttribute("totalPages", imageService.getTotalPages());
         model.addAttribute("currentPageNumber", pageNo);
         model.addAttribute("totalImageCount", imageService.getTotalImageCount());
-        List<IVSImage> images = imageService.getImages(pageNo, sortedBy, order);
+        List<IVSImage> images;
+        if(imageCategory.equals("Show All")) {
```
jdamerow, line 54: should not be hard-coded; especially since it's used before put in enum or constant in an interface

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ListImagesController.java
```diff
@@ -48,7 +50,19 @@ public String listSpaces(@PathVariable(required = false) String page,
...
+            model.addAttribute("totalPages", imageService.getTotalPages());
+            model.addAttribute("currentPageNumber", pageNo);
+            model.addAttribute("totalImageCount", imageService.getTotalImageCount());
+            images = imageService.getImages(pageNo, sortedBy, order);
+        }
+        else {
```
jdamerow, line 60: code style

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ListImagesController.java
```diff
@@ -48,7 +50,19 @@ public String listSpaces(@PathVariable(required = false) String page,
...
+            model.addAttribute("totalPages", imageService.getTotalPagesForSelectedFilter(ImageCategory.valueOf(imageCategory)));
```
jdamerow, line 61: since you're passing the category here into the method `getTotalPages` is descriptive enough as a method name

jdamerow: javadoc should do the rest of the explaining

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ListImagesController.java
```diff
@@ -48,7 +50,19 @@ public String listSpaces(@PathVariable(required = false) String page,
...
+            model.addAttribute("currentPageNumber", pageNo);
+            model.addAttribute("totalImageCount", imageService.getTotalImageCountForSelectedFilter(ImageCategory.valueOf(imageCategory)));
```
jdamerow, line 63: same as above

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ListImagesController.java
```diff
@@ -48,7 +50,19 @@ public String listSpaces(@PathVariable(required = false) String page,
...
+            images = imageService.getImagesByCategory(pageNo, ImageCategory.valueOf(imageCategory), sortedBy, order);
```
jdamerow, line 64: same as above

vspace/src/main/webapp/WEB-INF/views/staff/images/imagelist.html
```diff
@@ -123,11 +140,17 @@ <h2>Images</h2>
                         href="#">Created By</a></th>
                     <th scope="col" sort-prop="creationDate"><a
                         href="#">Created Date</a></th>
-                    <th scope="col" sort-prop="tag">Tag</th>
+                    <th scope="col" sort-prop="tag">Tag<select id="filterTag" name="tagFilter"
+                                class="form-control form-control-sm">
+                    <option>Show All</option>
+                    <option th:each="catTag:${imageCategories}"
+                                    th:value="${catTag}" th:selected="${#strings.equals(catTag,imageCategory)}">[[${@environment.getProperty('image_category_'
+                                    + catTag)}]]</option>
```
jdamerow, line 148: let's format that better
```
<option ...>
   [[...]]
</option>
```

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
```diff
@@ -116,8 +127,15 @@ public ImageData getImageDimensions(IVSImage image, int width, int height) {
         else {
             sortingParameters = sortingParameters.descending();
         }
+        return sortingParameters;
+    }
+
+    @Override
+    public List<IVSImage> getImages(int pageNo, ImageCategory category, String sortedBy, String order) {
```
jdamerow, line 134: is this method any different from the above `getImages`? it's only the find line, isn't it? Should be one method.

jdamerow (changes requested):

PR #168, opened 2020-12-09
Story/vspc 115

Sagar-k93:
It should be possible to include videos in slides. We need another content block but for videos. For now, let's have the video stored somewhere else (e.g. Youtube or another hosting service) and include them via URL.


jdamerow:
Please reopen to vspc-120 branch, so I actually see the changes.

PR #172, opened 2020-12-17
Story/vspc 142

ashmichheda:


jdamerow:
Make it so, Jenkins.

jdamerow:
The title of the exhibition should be shown, not of the current space. also, the current change seems to break the module pages.

vspace/src/main/webapp/WEB-INF/views/layouts/navbar.html
```diff
@@ -6,7 +6,8 @@
 	<body>
 		<div th:fragment="navbar">
 			<nav class="nav-bar" style="height:48px">
-		      <a class="navbar-brand" href="#"><img th:src="@{/resources/maintenance/img/logo-11.png}" alt="Virtual Spaces logo" width="100px"></a>
+		      <a class="navbar-brand" href="#"><img th:src="@{/resources/maintenance/img/logo-11.png}" alt="Virtual Spaces logo" width="100px">
+		      <span th:if="${exhibition != null}">[[${exhibition.title}]]</span></a>
```
jdamerow, line 10: I believe you can just do `[[${exhibition?.title}]]` and loose the if.

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

PR #173, opened 2020-12-18
Story/vspc 148

prjadhav14:
Ui fixes with respect to white space between space items in space lists and on the dashboard.


jdamerow:
Make it so, Jenkins.

jdamerow:
There is something off at the end:
image
Also, there is still quite a bit of white space that can be reduced (e.g. between boxes, within boxes top/bottom).

jdamerow:
Make it so, Jenkins.

jdamerow:
Whitespace needs to be reduced, icon needs to be centered in button
image

jdamerow:
Make it so, Jenkins.

jdamerow:
on dashboard, the break points don't seem to be right (spaces run over the white background):
image

on the spaces list, the break point should either earlier or the button should be align with the end of the fifth button. (actually both buttons (add and grid) should be aligned.
image

PR #174, opened 2020-12-18
Story/vspc 149

Sagar-k93:
Make the side bar less wide. There is a lot of whitespace that can be cut, make the top bar less high (cut whitespace), add hover titles to the icons in the top bar so the user knows what they do, the logout icon seems to have grey background, there are question marks in the bottom of the side bar, that needs to be fixed.


jdamerow:
Make it so, Jenkins.

jdamerow:
THere is a lot of space to the right of the menu now, and the menu entries don't seem to be centered:
image

Still a lot of whitespace at the top of the page content

PR #175, opened 2021-01-04
Story/vspc 146

prjadhav14:
Staff module page redesign to increase the number of grid items in a row.


jdamerow:
Make it so, Jenkins.

jdamerow:
Let's make the sequences a list. Maybe show the last modification date/time. also, there is still a lot of whitespace on the page (at the top, between the elements, etc), make all the buttons smaller, get rid of the outlined button (make them filled).

jdamerow:
Make it so, Jenkins.

jdamerow:
- the button should be smaller and the text should have equally spacing on all sites
image
- the sequence should be smaller and less spacious
image

jdamerow:
Make it so, Jenkins.

jdamerow:
- All buttons should be the same size (font/padding);
- new slide button: text isn't centered
- start sequence drop down needs to be wider so it's easier to red
- let's remove the negative margin, since the overall layout needs to be changed to reduce the padding of the page and then the negative margin will just mess it up (leave it with the default margins for now)

PR #177, opened 2021-01-04
Story/vspc 144

prjadhav14:
Modules list page redesign to fill up the used space.


jdamerow:
Make it so, Jenkins.

jdamerow:
Let's make the list layout the default one and sort it by newest first. The other layout somehow makes it harder to see what's going on.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ListModulesController.java
```diff
@@ -15,7 +15,7 @@
 
     @RequestMapping("/staff/module/list")
     public String listSpaces(Model model) {
-        model.addAttribute("vspacMmodules", moduleRepo.findAll());
+        model.addAttribute("vspacMmodules", moduleRepo.findAllByOrderByCreationDateDesc());
```
jdamerow, line 18: since you're at this, method should be renamed since it's not list modules.

jdamerow (changes requested):

jdamerow:
And let's reduce some of the whitespace
image
And make the add button a little smaller.

jdamerow:
Make it so, Jenkins.

jdamerow:
let's reduce the space between the grid button and the table a little more and make the white space around the grid icon a little less. then should be good to go.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ListModulesController.java
jdamerow: also, please fix spelling in attribute name

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

jdamerow:
👍 alright, let's reopen this to develop once 120 is merged.

PR #178, opened 2021-01-08
Story/vspc 140

Sagar-k93:
The search should search all space title and descriptions, all module title and descriptions and all slide texts. From the result page it should be possible to go directly to the found object.


jdamerow:
Please reopen PR to VSPC-120 branch so only your actually changes are visible.

PR #180, opened 2021-01-20
changed CSS properties to place bootstrap menus wrt to screens

ashmichheda:


vspace/src/main/webapp/WEB-INF/views/exhibition/space.html
```diff
@@ -20,71 +20,84 @@
 
 .Info_cz_Class:hover .tooltiptext {
     visibility: visible;
-}
-
+} */
```
jdamerow, line 23: remove comments if not needed

vspace/src/main/webapp/WEB-INF/views/exhibition/space.html
```diff
@@ -20,71 +20,84 @@
...
+            if(window.matchMedia(maxWidthSmallScreens).matches){
+                $("#mySidenav").css("width", "170px");
+                $("#mySidenav").css("height", "260px");
+                $("#mySidenav").css("left", "200px");
+		        $("#mySidenav").css("right", "200px");
+            }
+			/*
+			 Condition for big screens
+			*/
+			else if (window.matchMedia(maxWidth).matches){
```
jdamerow, line 97: indentation

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
It's actually almost less usable now. The links are partially not displayed on my phone. The menu should open above the image like the bootstrap menus do, big enough to select them with a finger.
image

ashmichheda:
I went through an example of bootstrap menu. By default, the menu appears extended when the screen is maximized and they collapse when the screen is minimized.
But in our application, the menu is already collapsed when maximized and when the screen is minimized we still want the menu to be collapsed. 
Only on small screens, it should show above the image. I am not sure.

I am going through the below example.

[Boostrap NavBar](https://getbootstrap.com/docs/3.4/examples/navbar/)

Is this the right one to refer as an example?

ashmichheda:
Should the menu be along with the sign out and dashboard icons for small screens?

PR #181, opened 2021-01-25
Story/vspc 89

ashmichheda:


vspace/src/main/webapp/WEB-INF/views/staff/spaces/spacelist.html
```diff
@@ -182,4 +368,5 @@ <h5 class="modal-title" id="deleteModalTitle">
         </div>
     </div>
 </body>
+>>>>>>> branch 'develop' of https://github.com/diging/virtual-spaces-2.0.git
```
jdamerow, line 371: there is a conflict in this file that needs to be resolved.

jdamerow (changes requested):

jdamerow (changes requested):
there is an empty file being added (slide.jsp), please review your pr before you open it.

jdamerow:
Make it so, Jenkins.

jdamerow:
Layout of added part is still different than the rest

jdamerow:
Make it so, Jenkins.

PR #188, opened 2021-02-05
Story/vspc 55

prjadhav14:
This ticket is about filtering the image list using the selected image category. Opening a new PR with comparison to develop branch after working on review comments given on draft PR.


vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
```diff
@@ -84,6 +85,20 @@ public ImageData getImageDimensions(IVSImage image, int width, int height) {
...
 
+    private Sort getSortingParameters(String sortedBy, String order) {
+        Sort sortingParameters = Sort.by(SortByField.CREATION_DATE.getValue()).descending();
+        if(sortedBy!=null && SortByField.getAllValues().contains(sortedBy)) {
+            sortingParameters = Sort.by(sortedBy);
+        }
+        if(order!=null && order.equalsIgnoreCase(Sort.Direction.ASC.toString())) {
+            sortingParameters = sortingParameters.ascending();
+        }
+        else {
```
jdamerow, line 96: code style

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
```diff
@@ -92,69 +107,48 @@ public ImageData getImageDimensions(IVSImage image, int width, int height) {
...
-     * 
-     * @param pageNo page provided by calling method
-     * @return 1 if pageNo less than 1 and lastPage if pageNo greater than
-     *         totalPages.
-     */
     @Override
-    public int validatePageNumber(int pageNo) {
-        long totalPages = getTotalPages();
+    public int validatePageNumber(int pageNo, String category) {
+        long totalPages = getTotalPages(category);
```
jdamerow, line 151: Is there a reason you deleted all javadoc?  it definitely needs some.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
```diff
@@ -92,69 +107,48 @@ public ImageData getImageDimensions(IVSImage image, int width, int height) {
...
-        if(sortedBy!=null && SortByField.getAllValues().contains(sortedBy)) {
-            sortingParameters = Sort.by(sortedBy);
-        }
-        if(order!=null && order.equalsIgnoreCase(Sort.Direction.ASC.toString())) {
-            sortingParameters = sortingParameters.ascending();
-        }
-        else {
-            sortingParameters = sortingParameters.descending();
-        }
+    public List<IVSImage> getImages(int pageNo, String category, String sortedBy, String order) {
```
jdamerow, line 115: can't you just use `null` to get all categories? It would be more intuitive than a string that in theory could be anything.

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
prjadhav14: In that case, URL query parameter value for imageCat would also be shown as null?

jdamerow: these are two separate things. You can pass an empty query parameter but pass null into the service class.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
```diff
@@ -84,77 +84,103 @@ public ImageData getImageDimensions(IVSImage image, int width, int height) {
...
+    
     /**
      * Method to return the requested images
      * 
      * @param pageNo. if pageNo<1, 1st page is returned, if pageNo>total pages,last
      *                page is returned
      * @return list of images in the requested pageNo
      */
+    @Override
+    public List<IVSImage> getImages(int pageNo, String category) {
```
jdamerow, line 108: category needs to be added to javadoc

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
```diff
@@ -84,77 +84,103 @@ public ImageData getImageDimensions(IVSImage image, int width, int height) {
...
-        else {
-            sortingParameters = sortingParameters.descending();
-        }
+    public List<IVSImage> getImages(int pageNo, String category, String sortedBy, String order) {
+        Sort sortingParameters = getSortingParameters(sortedBy, order);
+        pageNo = validatePageNumber(pageNo, category);
         Pageable sortByRequestedField = PageRequest.of(pageNo - 1, pageSize, sortingParameters);
-        Page<VSImage> images = imageRepo.findAll(sortByRequestedField);
+        Page<VSImage> images;
+        if(category==null || category.equals("null") || category.isEmpty()) {
```
jdamerow, line 132: category should never equal "null". That should be taken care of before this method is called.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
```diff
@@ -84,77 +84,103 @@ public ImageData getImageDimensions(IVSImage image, int width, int height) {
...
      * 
-     * @return totalPages required to display all images in DB
+     * @return total count of images in DB
      */
+
     @Override
-    public long getTotalPages() {
-        return (imageRepo.count() % pageSize==0) ? imageRepo.count() / pageSize:(imageRepo.count() / pageSize) + 1;
+    public long getTotalImageCount(String category) {
+        if(category==null || category.equals("null") || category.isEmpty()) {
```
jdamerow, line 152: same as above

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
```diff
@@ -84,77 +84,103 @@ public ImageData getImageDimensions(IVSImage image, int width, int height) {
...
+            return imageRepo.count();
+        }
+        return imageRepo.countByCategories(ImageCategory.valueOf(category));
```
jdamerow, line 155: also, just noticing, I would pass in an ImageCategory and making the conversion from string to ImageCategory before calling this class.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
```diff
@@ -84,77 +84,103 @@ public ImageData getImageDimensions(IVSImage image, int width, int height) {
...
      * 
-     * @return total count of images in DB
+     * @return totalPages required to display all images in DB
      */
+
     @Override
-    public long getTotalImageCount() {
-        return imageRepo.count();
+    public long getTotalPages(String category) {
+        if(category==null || category.equals("null") || category.isEmpty()) {
```
jdamerow, line 166: same as before

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
jdamerow: I know this method was already here, but does it need to be public? Is it used anywhere outside of this class?

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
Also, imageCat should not be passing "null" to the controller. It should just be empty string.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
prjadhav14: In that case, we would need to add the "All" constant to handle all images without any filter applied scenario.

jdamerow: or just null?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
jdamerow: ?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ListImagesController.java
```diff
@@ -30,25 +30,32 @@
...
     @RequestMapping("/staff/images/list/{page}")
     public String listSpaces(@PathVariable(required = false) String page,
+            @RequestParam(value = "imageCat", required = false) String imageCategory,
             @RequestParam(value = "sort", required = false) String sortedBy,
             @RequestParam(value = "order", required = false) String order, Model model) {
         int pageNo;
         page = StringUtils.isEmpty(page) ? "1" : page;
+        ImageCategory category = null;
+        if(!imageCategory.isEmpty()) {
+            category = ImageCategory.valueOf(imageCategory);
```
jdamerow, line 47: I believe this line can throw a `IllegalArgumentException`. Since this is unvalidated input from the front end you should handle this here.

jdamerow (changes requested):

jdamerow:
When filtering, it doesn't keep the previous sort order.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
jdamerow: This an open question.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ListImagesController.java
```diff
@@ -30,25 +31,38 @@
...
-            @RequestParam(value = "order", required = false) String order, Model model) {
+            @RequestParam(value = "order", required = false) String order, Model model, RedirectAttributes attributes) {
         int pageNo;
         page = StringUtils.isEmpty(page) ? "1" : page;
+        ImageCategory category = null;
+        if(!imageCategory.isEmpty()) {
+            try {
+                category = ImageCategory.valueOf(imageCategory);
+            }catch(IllegalArgumentException e) {
+                category=null;
```
jdamerow, line 52: I would log a warning here.

vspace/src/main/webapp/WEB-INF/views/staff/images/imagelist.html
```diff
@@ -76,15 +77,29 @@
...
         }else{
             window.location.href = window.location.pathname+
-            '?sort='+ headerSortPropName+'&order=asc';
+            '?imageCat=[(${imageCategory})]&sort='+ headerSortPropName+'&order=asc';
         }
     }
     });
+   
+   $("#filterTag").on("change", function(e) {
+       //e.preventDefault();
```
jdamerow, line 91: Why is this commented out?

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
prjadhav14: The method is also used in Controller and hence cannot be made private.

jdamerow:
Make it so, Jenkins.

PR #189, opened 2021-02-08
Story/vspc 93

ashmichheda:


jdamerow:
please reread the PR guidelines.

vspace/src/main/java/edu/asu/diging/vspace/core/data/ModuleRepository.java
```diff
@@ -1,19 +1,19 @@
...
+
+@Repository
+@JaversSpringDataAuditable
+public interface ModuleRepository extends PagingAndSortingRepository<Module, String> {
+
+    List<Module> findTop5ByOrderByCreationDateDesc();
+    
+    List<Module> findAllByOrderByCreationDateDesc();
+
+}
```
jdamerow, line 19: what has changed in this file?

vspace/src/main/webapp/WEB-INF/views/layouts/navbar.html
```diff
@@ -1,46 +1,46 @@
...
+		          	<form sec:authorize="isAuthenticated()" th:action="@{/logout}" method="POST">
+		         	 	<input type="hidden" name="${_csrf.parameterName}" value="${_csrf.token}"/>
+		  				<button class="btn nav-link" type="submit" title="Logout"><i class="fas fa-sign-out-alt"></i></button>
+		         	</form>
+		         	<a class="btn nav-link" sec:authorize="isAnonymous()" th:href="@{/login}"><i class="fas fa-sign-in-alt"></i></a>
+		        </div>
+		      </div>
+		    </nav>
+	    </div>
+	</body>
```
jdamerow, line 45: what has actually been changed in this file?

jdamerow (changes requested):
Have you looked at the PR after you've opened it? according to this, you've changed a lot of these files completely, which I don't believe you did?

ashmichheda:
When I merged code from Develop to my branch. I saw that these files were updated. So I have just changed the UNIX file encoding to all these files.

ashmichheda:
Am I supposed to do anything else as well?

jdamerow:
DashboardController does not seem to have unix file endings. Any file that shows up as changed although you didn't change it, make sure the change is necessary, and make a comment why it has been changed.

PR #201, opened 2021-04-19
Story/vspc 154

ashmichheda:

   vspc 154
...Put description here...
  
None
... describe here ...


vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSlideController.java
```diff
@@ -67,7 +67,7 @@ public String slide(Model model, @PathVariable("slideId") String slideId, @PathV
         if (module.getStartSequence() == null) {
             model.addAttribute("showAlert", true);
             model.addAttribute("message", "Sorry, module has not been configured yet.");
-            return "module";
```
jdamerow, line 70: what happened to that page? It should still exist just in a different location?

jdamerow: Then the module page needs to be fixed. I'm pretty certain that it was the same page before for the not configured message and the module. We don't want to have to change two pages if we make layout changes, if we can just change one instead.

jdamerow:
Make it so, Jenkins.

jdamerow:
There needs to be a back button so the user can get back to the space they came from.

jdamerow:
Make it so, Jenkins.

PR #213, opened 2021-05-21
[VSPC-73] done the implementation of graphical view of spaces

AvirupBiswas:

There should be some kind of diagram that shows all spaces (possibly as thumbnails of the selected background images) with links between connected spaces and icons for module that are connected to the spaces they are linked from
  
... describe here ...


jdamerow:
This is the ticket we talked about earlier, right? Please reopen when you're done with the changes.

Also, in the top part it says: "you can copy the ticket if it hasn't changed". Please do so.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,172 @@
...
+                List<ISpaceLinkDisplay> tempSpaceToSpacelinkList = new ArrayList<>();
+                tempSpaceToSpacelinkList.add(link);
+                AllSpaceToSpacelinksListMap.put(link.getLink().getSourceSpace().getId(), tempSpaceToSpacelinkList);
+            } else {
+                List<ISpaceLinkDisplay> tempSpaceToSpacelinkList = AllSpaceToSpacelinksListMap
+                        .get(link.getLink().getSourceSpace().getId());
+                tempSpaceToSpacelinkList.add(link);
+                AllSpaceToSpacelinksListMap.put(link.getLink().getSourceSpace().getId(), tempSpaceToSpacelinkList);
+            }
+        });
```
jdamerow, line 90: see above

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,172 @@
...
+
+        List<SpaceLinkDisplay> AllSpaceLinkForGivenOrNullSpaceStatusList = spaceLinkDisplayRepository
+                .findAllSpaceLinksForGivenOrNullSpaceStatus(SpaceStatus.PUBLISHED);
+        Map<String, List<ISpaceLinkDisplay>> AllSpaceLinkForGivenOrNullSpaceStatustMap = new HashMap<>();
```
jdamerow, line 94: bad variable name; should be shorter (and no capital first letter!)

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,172 @@
...
+                AllSpaceLinkForGivenOrNullSpaceStatustMap.put(link.getLink().getSourceSpace().getId(),
+                        tempSpaceToSpacelinkList);
+            }
+        });
+        if (spaceList != null && !spaceList.isEmpty()) {
+            spaceList.forEach(s -> {
+                ObjectNode node = mapper.createObjectNode();
+                node.put("name", s.getName());
+                node.put("link", request.getContextPath() + "/staff/space/" + s.getId());
+                node.put("img", request.getContextPath() + "/api/image/" + s.getImage().getId());
```
jdamerow, line 115: these paths should not be hard-coded; put them in a constant in the controller they come from and use the constant here (and in the the requestmapping annotation).

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,172 @@
...
+                node.put("isModule", false);
+                List<String> spaceAllLink = new ArrayList<>();
+                List<IModuleLinkDisplay> spaceToModulelinksList = AllSpaceToModulelinksListMap.get(s.getId());
```
jdamerow, line 118: initialize where needed

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,172 @@
...
+            }
+        });
+        if (spaceList != null && !spaceList.isEmpty()) {
+            spaceList.forEach(s -> {
+                ObjectNode node = mapper.createObjectNode();
+                node.put("name", s.getName());
+                node.put("link", request.getContextPath() + "/staff/space/" + s.getId());
+                node.put("img", request.getContextPath() + "/api/image/" + s.getImage().getId());
+                node.put("isModule", false);
+                List<String> spaceAllLink = new ArrayList<>();
```
jdamerow, line 117: initialize where needed

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,172 @@
...
+                List<IModuleLinkDisplay> spaceToModulelinksList = AllSpaceToModulelinksListMap.get(s.getId());
+                List<ISpaceLinkDisplay> spaceLinks = null;
+                if (s.isShowUnpublishedLinks()) {
+                    spaceLinks = AllSpaceToSpacelinksListMap.get(s.getId());
+                } else {
+                    spaceLinks = AllSpaceLinkForGivenOrNullSpaceStatustMap.get(s.getId());
+                }
```
jdamerow, line 124: on the staff page all links should be shown; the spaces that are unpublished should have a little icon like in the space list page.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,172 @@
...
+                List<ISpaceLinkDisplay> spaceToSpaceLinksList = null;
+                if (spaceLinks != null) {
+                    spaceToSpaceLinksList = spaceLinks.stream().filter(spaceLinkDisplayObj -> !spaceLinkDisplayObj
+                            .getLink().getTargetSpace().isHideIncomingLinks()).collect(Collectors.toList());
```
jdamerow, line 128: same here, the graph should show how all spaces and modules are connected and have icons to indicate if a space hides incoming links.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,172 @@
...
+            } else {
+                List<ISpaceLinkDisplay> tempSpaceToSpacelinkList = AllSpaceLinkForGivenOrNullSpaceStatustMap
+                        .get(link.getLink().getSourceSpace().getId());
+                tempSpaceToSpacelinkList.add(link);
+                AllSpaceLinkForGivenOrNullSpaceStatustMap.put(link.getLink().getSourceSpace().getId(),
+                        tempSpaceToSpacelinkList);
+            }
+        });
+        if (spaceList != null && !spaceList.isEmpty()) {
+            spaceList.forEach(s -> {
```
jdamerow, line 111: since this is a huge for loop, make it a proper for loop and give `s` a better name; this is really confusing.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,172 @@
...
+                spaceLinkMap.put(s.getId(), spaceAllLink);
+                allNodeList.add(s.getId());
+                nodeArray.add(node);
+            });
+        }
+        if (moduleList != null && !moduleList.isEmpty()) {
+            moduleList.forEach(m -> {
+                ObjectNode node = mapper.createObjectNode();
+                node.put("name", m.getName());
+                node.put("link", request.getContextPath() + "/staff/module/" + m.getId());
```
jdamerow, line 151: see above

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,172 @@
...
+        spaceLinkMap.forEach((k, v) -> {
+            if (!v.isEmpty()) {
+                v.forEach(e -> {
+                    ObjectNode node = mapper.createObjectNode();
+                    node.put("target", allNodeList.indexOf(e));
+                    node.put("source", allNodeList.indexOf(k));
+                    linkArray.add(node);
+                });
+            }
+        });
```
jdamerow, line 168: at least one of these loops should be a proper for loop; this nesting of `forEach` does not help with readability; it has the opposite effect. and give the variables proper names;

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,172 @@
...
+        model.addAttribute("overViewLink", mapper.writeValueAsString(linkArray));
```
jdamerow, line 169: "overview" is one word not two

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,172 @@
...
+        return "staff/spaces/graph";
+    }
```
jdamerow, line 171: this method needs to be broken up into smaller methods

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/core/data/SpaceRepository.java
```diff
@@ -18,5 +18,7 @@
     List<Space> findAllBySpaceStatus(SpaceStatus spaceStatus);
 
     List<Space> findAllByImageId(String imageId);
+    
+    List<Space> findAllByOrderByCreationDateAsc();
```
jdamerow, line 22: is this used anywhere?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,236 @@
...
+    private SpaceLinkDisplayRepository spaceLinkDisplayRepository;
+
+    @Autowired
+    private ModuleLinkDisplayRepository moduleLinkDisplayRepository;
+
+    @RequestMapping("/staff/overview")
+    public String spaceOverview(HttpServletRequest request, Model model) throws JsonProcessingException {
+        Iterable<Space> allSpaceList = spaceRepo.findAll();
+        List<Space> spaceList = new ArrayList<>();
+        allSpaceList.forEach(space -> spaceList.add(space));
```
jdamerow, line 57: why don't you use `allSpaceList`?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,236 @@
...
+        Iterable<Module> allModuleList = moduleRepo.findAll();
+        List<Module> moduleList = new ArrayList<>();
+        allModuleList.forEach(module -> moduleList.add(module));
```
jdamerow, line 60: see above

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,236 @@
...
+        ObjectMapper mapper = new ObjectMapper();
+        ArrayNode nodeArray = mapper.createArrayNode();
+        ArrayNode linkArray = mapper.createArrayNode();
```
jdamerow, line 63: declare/initialize variables where needed

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,236 @@
...
+        List<String> allNodeList = new ArrayList<>();
+        Map<String, List<String>> spaceLinkMap = new LinkedHashMap<>();
+
+        List<IModuleLinkDisplay> allSpaceToModulelinksList = moduleLinkDisplayRepository
```
jdamerow, line 67: needs proper camel case (`moduleLinks`)

vspace/src/main/java/edu/asu/diging/vspace/core/data/display/ModuleLinkDisplayRepository.java
```diff
@@ -19,4 +19,7 @@
     public List<IModuleLinkDisplay> findModuleLinkDisplaysForSpace(String spaceId);
 
     public void deleteByLink(IModuleLink link);
+    
+    @Query("SELECT d FROM ModuleLinkDisplay d")
+    public List<IModuleLinkDisplay> findModuleLinkDisplaysForAllSpace();
```
jdamerow, line 24: needs to be `allSpaces` (since all requires plural not singular)

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,236 @@
...
+                .findModuleLinkDisplaysForAllSpace();
+
+        Map<String, List<IModuleLinkDisplay>> allSpaceToModulelinksListMap = new HashMap<>();
```
jdamerow, line 70: see above

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,236 @@
...
+
+        getSpaceToAllOtherlink(allSpaceToModulelinksList, allSpaceToModulelinksListMap, true);
```
jdamerow, line 72: why not `spaceToModuleLinks` and `spaceToModuleLinksMap`

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,236 @@
...
+     * @param isModule
+     */
+    private <T> void getSpaceToAllOtherlink(List<T> spaceToAllOtherlinksList,
+            Map<String, List<T>> spaceToAllOtherlinksMap, boolean isModule) {
+        spaceToAllOtherlinksList.forEach(link -> {
+            String spaceId = isModule == true ? ((IModuleLinkDisplay) link).getLink().getSpace().getId()
+                    : ((ISpaceLinkDisplay) link).getLink().getSourceSpace().getId();
+            if (!spaceToAllOtherlinksMap.containsKey(spaceId)) {
+                List<T> tempSpaceToOtherlinkList = new ArrayList<>();
+                spaceToAllOtherlinksMap.put(spaceId, tempSpaceToOtherlinkList);
```
jdamerow, line 157: merge with line 156

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,236 @@
...
+            }
+            List<T> tempSpaceToOtherlinkList = spaceToAllOtherlinksMap.get(spaceId);
+            tempSpaceToOtherlinkList.add(link);
```
jdamerow, line 160: merge with line 159

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,236 @@
...
+
+    /**
+     * To map each space to all linked space or each space to all linked module
+     * 
+     * @param <T>
+     * @param spaceToAllOtherlinksList
+     * @param spaceToAllOtherlinksMap
+     * @param isModule
+     */
+    private <T> void getSpaceToAllOtherlink(List<T> spaceToAllOtherlinksList,
```
jdamerow, line 150: no generics needed here; just pass `List<ILinkDisplay>` (I'm fairly certain that exists)

jdamerow: also needs to be proper camelcase and plural `OtherLinks`

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,236 @@
...
+        Map<String, List<String>> spaceLinkMap = new LinkedHashMap<>();
+
+        List<IModuleLinkDisplay> allSpaceToModulelinksList = moduleLinkDisplayRepository
+                .findModuleLinkDisplaysForAllSpace();
+
+        Map<String, List<IModuleLinkDisplay>> allSpaceToModulelinksListMap = new HashMap<>();
+
+        getSpaceToAllOtherlink(allSpaceToModulelinksList, allSpaceToModulelinksListMap, true);
+
+        Iterable<SpaceLinkDisplay> spaceToSpacelinksList = spaceLinkDisplayRepository.findAll();
```
jdamerow, line 74: needs proper camel case

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,236 @@
...
+     * @param allSpaceToSpacelinksListMap
+     */
+    private void getSpaceLinkMap(HttpServletRequest request, List<Space> spaceList, ObjectMapper mapper,
+            ArrayNode nodeArray, List<String> allNodeList, Map<String, List<String>> spaceLinkMap,
+            Map<String, List<IModuleLinkDisplay>> allSpaceToModulelinksListMap,
+            Map<String, List<ISpaceLinkDisplay>> allSpaceToSpacelinksListMap) {
+        if (spaceList != null && !spaceList.isEmpty()) {
+            nodeFormation(request, spaceList, mapper, nodeArray, allNodeList, false);
+            for (Space space : spaceList) {
+                List<IModuleLinkDisplay> spaceToModulelinksList = allSpaceToModulelinksListMap.get(space.getId());
```
jdamerow, line 116: proper camel case

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,236 @@
...
+                List<ISpaceLinkDisplay> spaceLinks = null;
+                spaceLinks = allSpaceToSpacelinksListMap.get(space.getId());
```
jdamerow, line 118: merge with 117

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,236 @@
...
+                List<ISpaceLinkDisplay> spaceToSpaceLinksList = null;
+                if (spaceLinks != null) {
+                    spaceToSpaceLinksList = new ArrayList<>();
```
jdamerow, line 121: move to line 119

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,236 @@
...
+                    spaceToSpaceLinksList.addAll(spaceLinks);
+                }
+                List<String> spaceAllLink = new ArrayList<>();
+                if (spaceToModulelinksList != null && !spaceToModulelinksList.isEmpty()) {
+
+                    spaceToModulelinksList.forEach(link -> {
+                        spaceAllLink.add(link.getLink().getModule().getId());
+                    });
+                }
+                if (spaceToSpaceLinksList != null && !spaceToSpaceLinksList.isEmpty()) {
```
jdamerow, line 131: once `spaceToSpaceLinksList` has been initialized as suggested above, you can loose this if statement;

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,236 @@
...
+                List<IModuleLinkDisplay> spaceToModulelinksList = allSpaceToModulelinksListMap.get(space.getId());
+                List<ISpaceLinkDisplay> spaceLinks = null;
+                spaceLinks = allSpaceToSpacelinksListMap.get(space.getId());
+                List<ISpaceLinkDisplay> spaceToSpaceLinksList = null;
+                if (spaceLinks != null) {
+                    spaceToSpaceLinksList = new ArrayList<>();
+                    spaceToSpaceLinksList.addAll(spaceLinks);
+                }
+                List<String> spaceAllLink = new ArrayList<>();
+                if (spaceToModulelinksList != null && !spaceToModulelinksList.isEmpty()) {
```
jdamerow, line 125: you don't need the `isEmpyt` check; if the list is empty, the for loop will simply not be executed (since there are no objects in the list)

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,236 @@
...
+     * @param request
+     * @param nodeList
+     * @param mapper
+     * @param nodeArray
+     * @param allNodeList
+     * @param isModule
+     */
+    private <T> void nodeFormation(HttpServletRequest request, List<T> nodeList, ObjectMapper mapper,
+            ArrayNode nodeArray, List<String> allNodeList, boolean isModule) {
+        nodeList.forEach(nodeval -> {
```
jdamerow, line 177: make this a proper for loop (it's too complex for a `forEach`, there is no gain in using it.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,236 @@
...
+            VSpaceElement nodeObj = isModule == true ? (Module) nodeval : (Space) nodeval;
```
jdamerow, line 178: what's the point of this line? You are casting `nodeval` but you put it in a vspacelement variable.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,236 @@
...
+            ObjectNode node = mapper.createObjectNode();
+            node.put("name", nodeObj.getName());
+            StringBuilder linkPathBuilder = new StringBuilder();
+            linkPathBuilder.append(request.getContextPath());
```
jdamerow, line 182: pass the context pass into the method instead of the request object, since you don't need anything else from it.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,236 @@
...
+                }
+                if (((Space) nodeObj).isHideIncomingLinks()) {
+                    node.put("isHideIncomingLinks", true);
+                } else {
+                    node.put("isHideIncomingLinks", false);
+                }
+            }
+            nodeArray.add(node);
+            allNodeList.add(nodeObj.getId());
+        });
```
jdamerow, line 212: add some new lines to separate logical blocks visually

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,236 @@
...
+     * @param mapper
+     * @param linkArray
+     * @param allNodeList
+     * @param spaceLinkMap
+     */
+    private void sourceToTargetFormation(ObjectMapper mapper, ArrayNode linkArray, List<String> allNodeList,
+            Map<String, List<String>> spaceLinkMap) {
+        spaceLinkMap.forEach((k, v) -> {
+            if (!v.isEmpty()) {
+                v.forEach(e -> {
```
jdamerow, line 227: use regular for loops; there is no gain in nesting `forEach`, it's just harder to read (especially with varialbe names `v` and `k`.

jdamerow (changes requested):
- there are code factor issues to be adressed
- check all your variable names for proper camel case

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
AvirupBiswas: because VSpaceElement is the parent class of both Module and Space class. After ternary operator gets executed either it will Module Object or Space Object so it should be hold in a common variable.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
AvirupBiswas: After some modification of code the now is as follows:

Iterable<ModuleLinkDisplay> spaceToModuleLinks = moduleLinkDisplayRepository.findAll();
        Map<String, List<ModuleLinkDisplay>> spaceToModuleLinksMap = new HashMap<>();
        getSpaceToAllOtherlink(spaceToModuleLinks, spaceToModuleLinksMap, true);

Iterable<SpaceLinkDisplay> spaceToSpaceLinksList = spaceLinkDisplayRepository.findAll();
        Map<String, List<SpaceLinkDisplay>> spaceToSpaceLinksMap = new HashMap<>();
        getSpaceToAllOtherlink(spaceToSpaceLinksList, spaceToSpaceLinksMap, false);

as the findAll is returing the class not the iterface. Moreover getLink() method is not present in ILinkDisplay so I think generic is useful here.Please suggest.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
jdamerow: just pass VSpaceElement (which I believe should be IVspaceElement?) No casting needed.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
jdamerow: your manager should not return a list of classes but a list of interfaces; besides that though leave it for now and I'll take another look when all changes are made.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,246 @@
...
+                }
+                if (node != null) {
+                    nodeArray.add(node);
+                }
+                if (nodeObj.getId() != null) {
+                    allNodeList.add(nodeObj.getId());
+                }
+            }
+        }
+    }
```
jdamerow, line 224: to make all of this easier to read and maintain, I would create create two methods instead of this one: one that handles modules, the other that handles spaces; then I would take all the common parts and put those into shared methods, so that the modules and spaces methods just call other common methods but do what is specific to their classes, eg.:
```
// spaces method
for (Space space : spaces) {
   buildCommonNodeProperties(...) // also passing path to be used in link
   // set image path
   // set unpublished property
  // etc
}
```
similar for modules.

Does that make sense?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,246 @@
...
+
+    /**
+     * creating json for source node to all target nodes in the spaceoverview graph
+     * 
+     * @param mapper
+     * @param linkArray
+     * @param allNodeList
+     * @param spaceLinkMap
+     */
+    private void sourceToTargetFormation(ObjectMapper mapper, ArrayNode linkArray, List<String> allNodeList,
```
jdamerow, line 234: same as above, what does that mean? needs better name.

jdamerow (changes requested):
Please avoid to commit all these files that have only a line removed or something but no real changes (one of the checkboxes you checked, is actually mentioning exactly that).

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,246 @@
...
+     * @param allNodeList
+     * @param spaceLinkMap
+     * @param allSpaceToModuleLinksMap
+     * @param allSpaceToSpaceLinksMap
+     */
+    private void getSpaceLinkMap(String contextPath, Iterable<Space> allSpaceList, ObjectMapper mapper,
+            ArrayNode nodeArray, List<String> allNodeList, Map<String, List<String>> spaceLinkMap,
+            Map<String, List<ModuleLinkDisplay>> spaceToModuleLinksMap,
+            Map<String, List<SpaceLinkDisplay>> spaceToSpaceLinksMap) {
+        if (allSpaceList != null) {
```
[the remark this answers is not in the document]
AvirupBiswas: The whole for loop **for (Space space : allSpacesList)**  is there. So implemented check at this point.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,317 @@
...
+                }
+            }
+            if (spaceId != null) {
+                if (!spaceToAllOtherLinksMap.containsKey(spaceId)) {
+                    spaceToAllOtherLinksMap.put(spaceId, new ArrayList<>());
+                }
+                spaceToAllOtherLinksMap.get(spaceId).add(link);
+            }
+        }
+    }
```
jdamerow, line 181: break this into two methods: one for spacelinkdiplays one for modulelinkdisplays; you're not saving a lot of code by having a generic method, it just makes it more confusing.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,317 @@
...
+     * @param spaceToModuleLinksMap For getting list of links from one space to
+     *                              other modules
+     * @param spaceToSpaceLinksMap  For getting list of links from one space to
+     *                              other spaces
+     */
+    private void getSpaceLinkMap(String contextPath, Iterable<Space> allSpacesList, ObjectMapper mapper,
+            ArrayNode nodeArray, List<String> allNodesList, Map<String, List<String>> spaceLinkMap,
+            Map<String, List<ModuleLinkDisplay>> spaceToModuleLinksMap,
+            Map<String, List<SpaceLinkDisplay>> spaceToSpaceLinksMap) {
+        if (allSpacesList != null) {
```
jdamerow, line 107: this check should be where the method is being called

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,317 @@
...
+                    node.put("name", nodeObj.getName());
+                    String id = nodeObj.getId();
+
+                    if (!isModule) {
+                        createSpaceNode(linkPathBuilder, imagePathBuilder, isModule, (Space) nodeval, node, contextPath,
+                                id);
+                    } else {
+                        createModuleNode(linkPathBuilder, imagePathBuilder, isModule, node, contextPath, id);
+                    }
+                    if (node != null) {
```
jdamerow, line 215: how can node be null?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,317 @@
...
+                        nodeArray.add(node);
+                    }
+                    if (id != null) {
```
jdamerow, line 218: how can id be null?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,317 @@
...
+     * @param <T>
+     * @param contextPath This variable holds the contextpath of the application
+     * @param nodeList    List of space/module
+     * @param mapper      For creating ObjectNode
+     * @param nodeArray   Creating Array of ObjectNode for nodes in spaceoverview
+     *                    graph
+     * @param allNodeList List of all the nodes of spaceoverview graph
+     * @param isModule    boolean variable to indicate whether it is module or not
+     */
+    private <T> void createNodeForOverviewGraph(String contextPath, Iterable<T> nodeList, ObjectMapper mapper,
```
jdamerow, line 195: instead of the generics that don't really do anything for you here, just pass `Iterable<? extends IVSpaceElement>` and get rid of the `isModule` flag. Then iterate over `IVSpaceElement` and check with `instanceOf`.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,317 @@
...
+        linkPathBuilder.append(contextPath);
+        linkPathBuilder.append(SpaceController.STAFF_SPACE_PATH);
+        linkPathBuilder.append(spaceId);
+        node.put("link", linkPathBuilder.toString());
+        imagePathBuilder.setLength(0);
+        imagePathBuilder.append(contextPath);
+        imagePathBuilder.append(ImageApiController.API_IMAGE_PATH);
+        imagePathBuilder.append(space.getImage().getId());
+        node.put("img", imagePathBuilder.toString());
+        node.put("isModule", isModule);
```
jdamerow, line 255: if this method is to space nodes, why is there a flag `isModule` passed?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,317 @@
...
+     */
+    private void createModuleNode(StringBuilder linkPathBuilder, StringBuilder imagePathBuilder, boolean isModule,
+            ObjectNode node, String contextPath, String moduleId) {
+        linkPathBuilder.setLength(0);
+        linkPathBuilder.append(contextPath);
+        linkPathBuilder.append(ModuleController.STAFF_MODULE_PATH);
+        linkPathBuilder.append(moduleId);
+        node.put("link", linkPathBuilder.toString());
+        node.put("img", "");
+        node.put("isModule", isModule);
```
jdamerow, line 293: if this is for creating module nodes, why is there a flag `isModule` passed?

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
AvirupBiswas: That boolean value is necessary in the Json for both space and module. In case of space this flag false and in case of module this flag is true.
Based on the flag in the graph.html some actions are performed.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
AvirupBiswas: That boolean value is necessary in the Json for both space and module. In case of space this flag false and in case of module this flag is true.
Based on the flag in the graph.html some actions are performed.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
AvirupBiswas: This check is necessary as immediately after that there is a loop on it.

jdamerow:
closing as discussed on slack

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,321 @@
...
+    private ModuleRepository moduleRepo;
+
+    @Autowired
+    private SpaceLinkDisplayRepository spaceLinkDisplayRepository;
+
+    @Autowired
+    private ModuleLinkDisplayRepository moduleLinkDisplayRepository;
+
+    @RequestMapping("/staff/overview")
+    public String spaceOverview(HttpServletRequest request, Model model) throws JsonProcessingException {
```
jdamerow, line 54: this is all still very hard to follow, especially since you're mixing business logic (get a graph of all spaces and modules) and presentation (turn that graph into json). The business logic should actually go into a manager/service class (I would create a new one for that). This will also move the repositories in the business layer logic and it will make it easy to reuse for other purposes if needed. 
Your controller should then just turn the result into json. I would also recommend using ids for the edges as it should make the code that generates the json a lot easier (if you keep using the force layout, then this should help with that: http://bl.ocks.org/nitaku/7483341).

vspace/src/main/webapp/WEB-INF/views/staff/spaces/graph.html
```diff
@@ -0,0 +1,195 @@
...
+var linkJson = JSON.parse(strlinkVal);
+
+var w = window,
+d = document,
+e = d.documentElement,
+g = $('body')[0],
+x = w.innerWidth || e.clientWidth || g.clientWidth,
+y = w.innerHeight|| e.clientHeight|| g.clientHeight;
+
+var svg = d3.select("body").append("svg")
```
jdamerow, line 35: I'm sure there will be more than just the graph on the page eventually (if not in this ticket then in anotehr one), so I would use a dedicated div to append the svg to.

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
AvirupBiswas: I have to create link node json where I have to specify the source and target of each node .I don't have to write any extra code for edge creation. The link you have provided are using extra code to create link using id. We can discuss regarding this.
Note:I can see for each node in the node json there is a unique index starting from 0 gets appended to each json entry of the node json array.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceOverviewManager.java
```diff
@@ -0,0 +1,298 @@
...
+    @Autowired
+    private SpaceLinkDisplayRepository spaceLinkDisplayRepository;
+
+    @Autowired
+    private ModuleLinkDisplayRepository moduleLinkDisplayRepository;
+
+    /**
+     * To map each space to all linked space
+     */
+    @Override
```
jdamerow, line 51: needs proper javadoc

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceOverviewManager.java
```diff
@@ -0,0 +1,298 @@
...
+    public Map<String, List<ModuleLinkDisplay>> getSpaceToModuleLinks() {
+
+        Map<String, List<ModuleLinkDisplay>> spaceToModuleLinksMap = new HashMap<>();
+        Iterable<ModuleLinkDisplay> spaceToModuleLinksList = moduleLinkDisplayRepository.findAll();
+
+        for (ModuleLinkDisplay link : spaceToModuleLinksList) {
+            String spaceId = null;
+            if (link.getLink() != null && link.getLink().getSpace() != null
+                    && link.getLink().getSpace().getId() != null) {
```
jdamerow, line 60: `getId` should never return null, since all ModuleLinkDisplay are coming directly from the database hence all objects should have an id.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceOverviewManager.java
```diff
@@ -0,0 +1,298 @@
...
+
+                spaceId = link.getLink().getSpace().getId();
+            }
+            if (spaceId != null) {
+                if (!spaceToModuleLinksMap.containsKey(spaceId)) {
+                    spaceToModuleLinksMap.put(spaceId, new ArrayList<>());
+                }
+                spaceToModuleLinksMap.get(spaceId).add(link);
```
jdamerow, line 68: this if statement seems unnecessary; just put the code inside the previous if.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceOverviewManager.java
```diff
@@ -0,0 +1,298 @@
...
+            }
+        }
+        return spaceToModuleLinksMap;
+    }
+
+    /**
+     * To map each space to all linked space
+     */
+    @Override
```
jdamerow, line 77: needs proper javadoc

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceOverviewManager.java
```diff
@@ -0,0 +1,298 @@
...
+    public Map<String, List<SpaceLinkDisplay>> getSpaceToSpaceLinks() {
+
+        Map<String, List<SpaceLinkDisplay>> spaceToSpaceLinksMap = new HashMap<>();
+        Iterable<SpaceLinkDisplay> spaceToSpaceLinksList = spaceLinkDisplayRepository.findAll();
+        for (SpaceLinkDisplay link : spaceToSpaceLinksList) {
+            String spaceId = null;
+            if (link.getLink() != null && link.getLink().getSourceSpace() != null
+                    && link.getLink().getSourceSpace().getId() != null) {
```
jdamerow, line 85: see above

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceOverviewManager.java
```diff
@@ -0,0 +1,298 @@
...
+
+                spaceId = link.getLink().getSourceSpace().getId();
+            }
+            if (spaceId != null) {
+                if (!spaceToSpaceLinksMap.containsKey(spaceId)) {
+                    spaceToSpaceLinksMap.put(spaceId, new ArrayList<>());
+                }
+                spaceToSpaceLinksMap.get(spaceId).add(link);
```
jdamerow, line 93: see above

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceOverviewManager.java
```diff
@@ -0,0 +1,298 @@
...
+                List<SpaceLinkDisplay> spaceToSpaceLinksList = new ArrayList<>();
+                if (spaceLinks != null) {
+                    spaceToSpaceLinksList.addAll(spaceLinks);
+                }
+                List<String> spaceAllLink = new ArrayList<>();
+
+                if (spaceToModulelinksList != null) {
+                    spaceToModulelinksList.forEach(link -> {
+                        if (link.getLink() != null && link.getLink().getModule() != null
+                                && link.getLink().getModule().getId() != null) {
```
jdamerow, line 137: same as above, `getId` should never return null.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceOverviewManager.java
```diff
@@ -0,0 +1,298 @@
...
+
+        Iterable<Space> allSpacesList = spaceRepo.findAll();
+        if (allSpacesList != null) {
+            createSpaceNodeInOverviewGraph(contextPath, mapper, nodeArray);
+            for (Space space : allSpacesList) {
+                List<ModuleLinkDisplay> spaceToModulelinksList = spaceToModuleLinksMap.get(space.getId());
+                List<SpaceLinkDisplay> spaceLinks = spaceToSpaceLinksMap.get(space.getId());
+                List<SpaceLinkDisplay> spaceToSpaceLinksList = new ArrayList<>();
+                if (spaceLinks != null) {
+                    spaceToSpaceLinksList.addAll(spaceLinks);
```
jdamerow, line 130: why do you add the links into a new list here?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceOverviewManager.java
```diff
@@ -0,0 +1,298 @@
...
+                }
+                List<String> spaceAllLink = new ArrayList<>();
```
jdamerow, line 132: this list is called `spaceAllLink` but is a list of strings. what does it actually hold? name should reflect that.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceOverviewManager.java
```diff
@@ -0,0 +1,298 @@
...
+                        if (link.getLink() != null && link.getLink().getModule() != null
+                                && link.getLink().getModule().getId() != null) {
+                            spaceAllLink.add(link.getLink().getModule().getId());
+                        }
+                    });
+                }
+                if (spaceToSpaceLinksList != null) {
+                    spaceToSpaceLinksList.forEach(link -> {
+                        if (link.getLink() != null && link.getLink().getTarget() != null
+                                && link.getLink().getTarget().getId() != null) {
```
jdamerow, line 145: see above

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceOverviewManager.java
```diff
@@ -0,0 +1,298 @@
...
+                            spaceAllLink.add(link.getLink().getTarget().getId());
+                        }
+                    });
+                }
+                if (space.getId() != null) {
```
jdamerow, line 150: how can the id ever be null? it's coming from the database, there it must have na id.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceOverviewManager.java
```diff
@@ -0,0 +1,298 @@
...
+                    spaceLinkMap.put(space.getId(), spaceAllLink);
```
jdamerow, line 151: unless there is a good reason why `spaceLinkMap` is passed into the method,  it should be created inside this method and then returned.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceOverviewManager.java
```diff
@@ -0,0 +1,298 @@
...
+     * creating json for space and module node in the spaceoverview graph
+     * 
+     * @param contextPath This variable holds the contextpath of the application
+     * @param nodeList    List of space/module
+     * @param mapper      For creating ObjectNode
+     * @param nodeArray   Creating Array of ObjectNode for nodes in spaceoverview
+     *                    graph
+     */
+    @Override
+    public void createNodeForOverviewGraph(String contextPath, Iterable<? extends IVSpaceElement> nodeList,
```
jdamerow, line 167: all methods related to json should be in their own class and called after the space to space/module map has been created. what if we exchange the front end and won't need json anymore but a different format or we need the space to space/module map for something else? The way it is right now the code can't be reused for other purposes.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceOverviewManager.java
```diff
@@ -0,0 +1,298 @@
...
+
+        if (space.getSpaceStatus() != null && (space.getSpaceStatus().equals(SpaceStatus.UNPUBLISHED))) {
+            node.put("isUnpublished", true);
+        } else {
+            node.put("isUnpublished", false);
+        }
+        if (space.isHideIncomingLinks()) {
+            node.put("isHideIncomingLinks", true);
+        } else {
+            node.put("isHideIncomingLinks", false);
```
jdamerow, line 266: strings should be at least constants not hardcoded in the code

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,321 @@
...
+ * @author Avirup Biswas
+ *
+ */
+@Controller
+public class SpaceOverviewController {
+    @Autowired
+    private SpaceRepository spaceRepo;
+
+    @Autowired
+    private ModuleRepository moduleRepo;
```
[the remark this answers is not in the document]
jdamerow: ?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,62 @@
...
+
+        Map<String, List<String>> spaceLinkMap = new LinkedHashMap<>();
+
+        Map<String, List<ModuleLinkDisplay>> spaceToModuleLinksMap = spaceOverviewManager.getSpaceToModuleLinks();
+
+        Map<String, List<SpaceLinkDisplay>> spaceToSpaceLinksMap = spaceOverviewManager.getSpaceToSpaceLinks();
+
+        ObjectMapper mapper = new ObjectMapper();
+        ArrayNode nodeArray = mapper.createArrayNode();
+        ArrayNode linkArray = mapper.createArrayNode();
```
jdamerow, line 49: these objects should be created in the json helper class

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceOverviewManager.java
```diff
@@ -0,0 +1,119 @@
...
+     * 
+     * @param spaceLinkMap          For holding list of all links corresponding to
+     *                              each space
+     * @param spaceToModuleLinksMap For getting list of links from one space to
+     *                              other modules
+     * @param spaceToSpaceLinksMap  For getting list of links from one space to
+     *                              other spaces
+     */
+    @Override
+    public Map<String, List<String>> getSpaceLinkMap(Map<String, List<ModuleLinkDisplay>> spaceToModuleLinksMap,
```
jdamerow, line 89: what's the difference between this method and `getSpaceToSpaceLinks`? This needs to be clear from method name and javadoc.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceOverviewManager.java
```diff
@@ -0,0 +1,119 @@
...
+            Map<String, List<SpaceLinkDisplay>> spaceToSpaceLinksMap) {
+
+        Map<String, List<String>> spaceLinkMap = new LinkedHashMap<>();
+        Iterable<Space> allSpacesList = spaceRepo.findAll();
+        if (allSpacesList != null) {
+            for (Space space : allSpacesList) {
+                List<ModuleLinkDisplay> spaceToModulelinksList = spaceToModuleLinksMap.get(space.getId());
+                List<SpaceLinkDisplay> spaceToSpaceLinksList = spaceToSpaceLinksMap.get(space.getId());
+                List<String> listOfSpaceIdAndModuleId = new ArrayList<>();
```
jdamerow, line 98: since this is a list, it should be "Ids" not "Id"

vspace/src/main/java/edu/asu/diging/vspace/core/util/JsonHelper.java
```diff
@@ -0,0 +1,25 @@
...
+    {
+        String jsonResult="";
+        jsonResult=mapper.writeValueAsString(object);
+        return jsonResult;
+    }
+    
+    public static ObjectMapper getMapper() {
+        return mapper;
+    }
+}
```
jdamerow, line 25: none of these should be static methods or variables; create a component or service.

vspace/src/main/java/edu/asu/diging/vspace/core/util/JsonHelper.java
```diff
@@ -0,0 +1,25 @@
+package edu.asu.diging.vspace.core.util;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+public class JsonHelper {
```
jdamerow, line 6: javadoc

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,53 @@
...
+    private SpaceOverviewDataFormat spaceOverviewDataFormat;
+
+    @RequestMapping("/staff/overview")
+    public String spaceOverview(HttpServletRequest request, Model model) throws JsonProcessingException {
+
+        Map<String, List<ModuleLinkDisplay>> spaceToModuleLinksMap = spaceOverviewManager.getSpaceToModuleLinks();
+
+        Map<String, List<SpaceLinkDisplay>> spaceToSpaceLinksMap = spaceOverviewManager.getSpaceToSpaceLinks();
+
+        Map<String, List<String>> spaceLinkMap = spaceOverviewManager.getSpaceLinkMap(spaceToModuleLinksMap,
```
jdamerow, line 41: needs a better name since I don't think this maps spaces to links (but space ids to module/space ids?)

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceOverviewDataFormat.java
```diff
@@ -0,0 +1,221 @@
...
+        node.put(IS_MODULE, true);
+    }
+
+    /**
+     * @param contextPath   This variable holds the contextpath of the
+     *                      application
+     * @return Json corresponding to nodeArray
+     * @throws JsonProcessingException
+     */
+    public String createNodeForOverviewGraph(String contextPath) throws JsonProcessingException {
```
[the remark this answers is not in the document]
jdamerow: Also, why note just `createNodes`?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceOverviewDataFormat.java
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    /**
+     * creating json for edge between nodes in the spaceoverview graph
+     * 
+     * @param spaceLinkMap List of all links corresponding to each space
+     * @return Json corresponding to linkArray
+     * @throws JsonProcessingException
+     */
+    public String createLinkForOverviewGraph(Map<String, List<String>> spaceLinkMap) throws JsonProcessingException {
```
jdamerow, line 208: and `createEdges`?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,53 @@
...
+                spaceToSpaceLinksMap);
+
+        String nodeJson = spaceOverviewDataFormat.createNodeForOverviewGraph(request.getContextPath());
+
+        String linkJson = spaceOverviewDataFormat.createLinkForOverviewGraph(spaceLinkMap);
```
jdamerow, line 46: these are both several nodes and edges, right? so the name should reflect that.

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/core/services/ISpaceOverviewManager.java
```diff
@@ -0,0 +1,18 @@
...
+import edu.asu.diging.vspace.core.model.display.impl.ModuleLinkDisplay;
+import edu.asu.diging.vspace.core.model.display.impl.SpaceLinkDisplay;
+
+public interface ISpaceOverviewManager {
+
+    Map<String, List<ModuleLinkDisplay>> getSpaceToModuleLinks();
+
+    Map<String, List<SpaceLinkDisplay>> getSpaceToSpaceLinks();
+
+    Map<String, List<String>> getSpaceLinkedToSpacesAndModules(Map<String, List<ModuleLinkDisplay>> spaceToModuleLinksMap,
```
jdamerow, line 15: how about `getSpacesToSpacesAndModulesMap`?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceOverviewFormatter.java
```diff
@@ -0,0 +1,228 @@
...
+        }
+    }
+
+    /**
+     * This is a common method for both space and module
+     * 
+     * @param nodeval This variable store either space/module node
+     * @return After setting name variable of node it returns the node
+     */
+    private ObjectNode createNodeAndSetName(IVSpaceElement nodeval) {
```
jdamerow, line 110: what is "set name"?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceOverviewFormatter.java
```diff
@@ -0,0 +1,228 @@
...
+
+        ObjectMapper mapper = new ObjectMapper();
```
jdamerow, line 112: there shouldn't be any references to the json classes; that should be encapsulated in the jsonhelper.

jdamerow: or put everything in this class and get rid of the jsonhelper (what is it's responsibility here?).

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceOverviewManager.java
```diff
@@ -0,0 +1,123 @@
...
+                            listOfSpaceIdAndModuleIds.add(link.getLink().getTarget().getId());
+                        }
+                    });
+                }
+                spaceLinkMap.put(space.getId(), listOfSpaceIdAndModuleIds);
+            }
+        }
+        return spaceLinkMap;
+    }
+}
```
jdamerow, line 123: this class still seems to mix generating the data structure for the map and then also turning that into JSON. A class should only have one purpose (Single Responsibility Principle). So this class should generate the data structure and the jsonhelper should generate the json from it.

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceOverviewDataFormatter.java
```diff
@@ -0,0 +1,33 @@
...
+
+        String nodesJson = spaceOverviewJsonFormat.createNodes(contextPath);
+
+        String linksJson = spaceOverviewJsonFormat.createEdges(spacesToToSpacesAndModulesMap);
+
+        Map<String, String> jsonFormatMap = new HashMap<>();
+
+        jsonFormatMap.put("nodesJson", nodesJson);
+
+        jsonFormatMap.put("linksJson", linksJson);
```
jdamerow, line 29: blocks of. code that belong together should visibly be together (get rid of empty lines, e.g. line 22, 26, 28)

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceOverviewJsonFormat.java
```diff
@@ -0,0 +1,190 @@
...
+     *                         in staff side
+     * @param imagePathBuilder StringBuilder object to create the URL link for image
+     *                         in staff side
+     * @param space            passing the space object
+     * 
+     * @param contextPath      This variable holds the contextpath of the
+     *                         application
+     */
+    private SpaceOverview createSpaceNode(StringBuilder linkPathBuilder, StringBuilder imagePathBuilder, Space space,
+            String contextPath) {
```
jdamerow, line 92: why are the string builders being passed here? they don't seem to serve a purpose outside of this method.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceOverviewJsonFormat.java
```diff
@@ -0,0 +1,190 @@
...
+     * @param linkPathBuilder  StringBuilder object to create the URL link for
+     *                         module in staff side
+     * @param imagePathBuilder StringBuilder object to create the URL link for image
+     *                         in staff side
+     * @param module           module model object
+     * 
+     * @param contextPath      This variable holds the contextpath of the
+     *                         application
+     */
+    private SpaceOverview createModuleNode(StringBuilder linkPathBuilder, StringBuilder imagePathBuilder, Module module,
```
jdamerow, line 133: see above

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,56 @@
...
+
+        Map<String, String> jsonFormatMap = spaceOverviewDataFormatter.getJsonFormat(request.getContextPath(),
+                spacesToSpacesAndModulesMap);
+
+        String nodesJson = jsonFormatMap.get("nodesJson");
+
+        String linksJson = jsonFormatMap.get("linksJson");
+
+        model.addAttribute("overviewNode", nodesJson);
+        model.addAttribute("overviewLink", linksJson);
```
jdamerow, line 52: why don't you put the json as the map into the model directly?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/SpaceOverviewController.java
```diff
@@ -0,0 +1,56 @@
...
+
+        Map<String, List<String>> spacesToSpacesAndModulesMap = spaceOverviewManager
+                .getSpacesToSpacesAndModulesMap(spaceToModuleLinksMap, spaceToSpaceLinksMap);
+
+        Map<String, String> jsonFormatMap = spaceOverviewDataFormatter.getJsonFormat(request.getContextPath(),
+                spacesToSpacesAndModulesMap);
+
+        String nodesJson = jsonFormatMap.get("nodesJson");
+
+        String linksJson = jsonFormatMap.get("linksJson");
```
jdamerow, line 49: get rid of some of the empty lines so that lines of code that belong together are visibly together

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceOverviewDataFormatter.java
```diff
@@ -0,0 +1,30 @@
...
+import java.util.List;
+import java.util.Map;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+
+@Component
+public class SpaceOverviewDataFormatter {
```
jdamerow, line 13: javadoc

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceOverviewJsonFormat.java
```diff
@@ -0,0 +1,174 @@
...
+import edu.asu.diging.vspace.core.model.SpaceOverview;
+import edu.asu.diging.vspace.core.model.impl.Module;
+import edu.asu.diging.vspace.core.model.impl.Space;
+import edu.asu.diging.vspace.core.model.impl.SpaceStatus;
+import edu.asu.diging.vspace.web.api.ImageApiController;
+import edu.asu.diging.vspace.web.staff.ModuleController;
+import edu.asu.diging.vspace.web.staff.SpaceController;
+
+@Component
+public class SpaceOverviewJsonFormat {
```
jdamerow, line 22: javadoc

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

jdamerow:
What about the exception?

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

jdamerow:
Let's try a different layout that doesn't tend to create just one hairball when there are not enough links.  Also, it shouldn't move around, it should just stay fixed and the user can adjust as needed.

vspace/src/main/java/edu/asu/diging/vspace/core/model/SpaceOverview.java
```diff
@@ -0,0 +1,86 @@
+package edu.asu.diging.vspace.core.model;
```
jdamerow, line 1: this class is only used to pass data to the front end, isn't it? It shouldn't be in the core model package.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceOverviewManager.java
```diff
@@ -0,0 +1,131 @@
...
+import edu.asu.diging.vspace.core.data.SpaceRepository;
+import edu.asu.diging.vspace.core.data.display.ModuleLinkDisplayRepository;
+import edu.asu.diging.vspace.core.data.display.SpaceLinkDisplayRepository;
+import edu.asu.diging.vspace.core.model.display.impl.ModuleLinkDisplay;
+import edu.asu.diging.vspace.core.model.display.impl.SpaceLinkDisplay;
+import edu.asu.diging.vspace.core.model.impl.Space;
+import edu.asu.diging.vspace.core.services.ISpaceOverviewManager;
+
+@Service
+public class SpaceOverviewManager implements ISpaceOverviewManager {
```
jdamerow, line 23: needs test cases

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceOverviewJsonFormat.java
```diff
@@ -0,0 +1,166 @@
...
+        return spaceOverview;
+    }
+
+    /**
+     * @param contextPath This variable holds the contextpath of the application
+     * @param spacesToSpacesAndModulesMap
+     * @return JSON with all nodes
+     * @throws JsonProcessingException
+     */
+    public String createNodes(String contextPath, Map<String, List<String>> spacesToSpacesAndModulesMap)
```
jdamerow, line 146: needs test cases

jdamerow:
Make it so, Jenkins.

jdamerow:
for the diagram:
- the enlarged images are behind the small ones, should be the other way around
image
- the graph should only be a big as the window (there should be no scrolling needed) but the user should be able to zoom in.

tusanand:
Fixed alignment issue and issue with the image being cut at edges

jdamerow:
Make it so, Jenkins.

jdamerow:
actually, this one is the right one:
```
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /var/lib/jenkins/workspace/VSpace_deploy_to_review_on_pr/vspace/src/main/java/edu/asu/diging/vspace/config/PersistenceJPAConfig.java:[5,24] package javax.annotation does not exist
[ERROR] /var/lib/jenkins/workspace/VSpace_deploy_to_review_on_pr/vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SetupManager.java:[3,24] package javax.annotation does not exist
[ERROR] /var/lib/jenkins/workspace/VSpace_deploy_to_review_on_pr/vspace/src/main/java/edu/asu/diging/vspace/config/PersistenceJPAConfig.java:[39,6] cannot find symbol
  symbol:   class PostConstruct
  location: class edu.asu.diging.vspace.config.PersistenceJPAConfig
[ERROR] /var/lib/jenkins/workspace/VSpace_deploy_to_review_on_pr/vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SetupManager.java:[31,6] cannot find symbol
  symbol:   class PostConstruct
  location: class edu.asu.diging.vspace.core.services.impl.SetupManager
```
it needs an extra dependency

jdamerow:
Make it so, Jenkins.

PR #219, opened 2021-06-15
Story/vspc 150

pradeep-017:
# Guidelines for Pull Requests

If you haven't yet read our code review guidelines, please do so, You can find them [here](https://diging.atlassian.net/wiki/spaces/DIGING/pages/2256076801/Code+Review+Guidelines).

Please confirm the following by adding an x for each item (turn `[ ]` into `[x]`).

- [x] I have removed all code style changes that are not necessary (e.g. changing blanks across the whole file that don’t need to be changed, adding empty lines in parts other than your own code)
- [x] I am not making any changes to files that don’t have any effect (e.g. imports added that don’t need to be added)
- [x] I do not have any sysout statements in my code or commented out code that isn’t needed anymore
- [x] I am not reformatting any files in the wrong format or without cause. 
- [x] I am not changing file encoding or line endings to something else than UTF-8, LF
- [x] My pull request does not show an insane amount of files being changed although my ticket only requires a few files being changed
- [x] I have added Javadoc/documentation where appropriate
- [x] I have added test cases where appropriate
- [x] I have explained any part of my code/implementation decisions that is not be self-explanatory

## Please provide a brief description of your ticket 
Add Bibliography section to staff page

...Put description here...
  
## Anything else the reviewer needs to know?

... describe here ...

PR #228, opened 2021-07-06
Story/vspc 147

pradeep-017:

Completed changes for changes for Bibliographies and References. 
  
... describe here ...


vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ReferenceManager.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Override
+    public void updateReference(Reference reference) {
+        referenceRepo.save((Reference) reference);
+    }
+
+    @Override
+    public void deleteReferenceById(String referenceId, String BiblioId) {
+        if (referenceId == null) {
+            logger.error("Reference Id cannot be null.");
```
jdamerow, line 53: this should probably be just a warning not an error

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ReferenceManager.java
```diff
@@ -0,0 +1,76 @@
...
+        return null;
+    }
+
+    @Override
+    public void updateReference(Reference reference) {
+        referenceRepo.save((Reference) reference);
+    }
+
+    @Override
+    public void deleteReferenceById(String referenceId, String BiblioId) {
```
jdamerow, line 51: variable names are never capitalized in java (except constants)

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ReferenceManager.java
```diff
@@ -0,0 +1,76 @@
...
+        if (referenceId == null) {
+            logger.error("Reference Id cannot be null.");
+            return;
+        }
+
+        try {
+            referenceRepo.delete((Reference) getReference(referenceId));
```
jdamerow, line 58: does eclipse make you catch the exception here? I believe `IllegalArgumentException` is a runtime exception and is only thrown when id is null, which it should never be as you check it above.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ReferenceManager.java
```diff
@@ -0,0 +1,76 @@
...
+        } catch (IllegalArgumentException exception) {
+            logger.error("Unable to delete reference" + referenceId + ". ", exception);
+        }
+    }
+
+    @Override
+    public void deleteRefences(List<IReference> references, String BiblioId) {
```
jdamerow, line 65: see above

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ReferenceManager.java
```diff
@@ -0,0 +1,76 @@
...
+        for(IReference ref : references) {
+            deleteReferenceById(ref.getId(), BiblioId);
+        }
+        
+    }
+
+    @Override
+    public List<IReference> getReferencesForBiblio(String biblioId) {
+        return new ArrayList<>(referenceRepo.findReferencesForBiblio(biblioId));
```
jdamerow, line 74: is there a reason you don't just return the list that `findReferencesForBiblio` returns?

vspace/src/main/java/edu/asu/diging/vspace/referenceExpose/CitationStyleDefault.java
```diff
@@ -0,0 +1,7 @@
+package edu.asu.diging.vspace.referenceExpose;
+
+public class CitationStyleDefault implements ICitationStyle{
+    
+    String citationStyleName = "defaultZoteroCitation";
```
jdamerow, line 5: what is this for?

vspace/src/main/java/edu/asu/diging/vspace/referenceExpose/ICitationStyle.java
```diff
@@ -0,0 +1,7 @@
+package edu.asu.diging.vspace.referenceExpose;
+
+public interface ICitationStyle {
+    
+    String citationStyleName = null;
```
jdamerow, line 5: what is this for?

vspace/src/main/java/edu/asu/diging/vspace/referenceExpose/ZoteroDefaultMetaData.java
```diff
@@ -0,0 +1,118 @@
...
+    IReference reference;
+    
+    //These all values are fixed for default metadata 
+    String urlVersion = "url_ver=Z39.88-2004";
+    
+    String ctxVersion = "ctx_ver=Z39.88-2004";
+    
+    String rfrId = "info:sid/zotero.org:2";
+    
+    String rftValFmt = "info:ofi/fmt:kev:mtx:dissertation";
```
jdamerow, line 25: strings should be put in property file

vspace/src/main/java/edu/asu/diging/vspace/referenceExpose/ZoteroDefaultMetaData.java
```diff
@@ -0,0 +1,118 @@
+package edu.asu.diging.vspace.referenceExpose;
+
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
+
+import edu.asu.diging.vspace.core.model.IReference;
+
+public class ZoteroDefaultMetaData implements IReferenceMetadata {
```
jdamerow, line 9: It's a little confusing what all these classes do. This is all still pretty tightly coupled, I would change it like this:
- all instances of `IReferenceMetadata` should be services or components (maybe rename it to include the word formatter)
- ReferenceContext should be renamed to ReferencePresentationRegistry or something. It would get all beans of type IReferenceMetadata and put them in a map (identifier of formatter -> formatter instance)
- when we need a formatted metadata string, the identifier of the required formatter is used to retrieve the formatter via the registry

jdamerow: presentation of metadata should not be part of any domain objects and it shouldn't be necessary to touch existing classes just to add another formatter.

vspace/src/main/java/edu/asu/diging/vspace/referenceExpose/ZoteroDefaultMetaData.java
jdamerow: also, none of these need to be visible outside the class, do they?

vspace/src/main/java/edu/asu/diging/vspace/referenceExpose/ZoteroDefaultMetaData.java
```diff
@@ -0,0 +1,118 @@
...
+            if (url != null)
+                urlEncodedReferenceMetaData += ("&rft_id=") + URLEncoder.encode(url, StandardCharsets.UTF_8.toString());
+
+            if (volume != null)
+                urlEncodedReferenceMetaData += ("&rft.volume=")
+                        + URLEncoder.encode(volume, StandardCharsets.UTF_8.toString());
+
+            if (issue != null)
+                urlEncodedReferenceMetaData += ("&rft.issue=")
+                        + URLEncoder.encode(issue, StandardCharsets.UTF_8.toString());
```
jdamerow, line 88: please follow our code style

vspace/src/main/java/edu/asu/diging/vspace/referenceExpose/ZoteroDefaultMetaData.java
```diff
@@ -0,0 +1,118 @@
...
+                urlEncodedReferenceMetaData += ("&rft.editors=")
+                        + URLEncoder.encode(editors, StandardCharsets.UTF_8.toString());
+
+            if (note != null)
+                urlEncodedReferenceMetaData += ("&rft.note=") + URLEncoder.encode(note, StandardCharsets.UTF_8.toString());
+
+            urlEncodedReferenceMetaData += ("&rft.language=en");
+
+        } catch (UnsupportedEncodingException e) {
+            e.printStackTrace();
```
jdamerow, line 112: ?

vspace/src/main/java/edu/asu/diging/vspace/referenceExpose/ZoteroDefaultMetaData.java
```diff
@@ -0,0 +1,118 @@
...
+        this.reference = reference;
+    }
+
+    @Override
+    public String getReferenceMetadataStyle() {
+        return biblioMetadataStyle;
+    }
+
+    @Override
+    public String createReferenceMetadata(ICitationStyle citationStyle, IReference reference) {
```
jdamerow, line 38: what is `citationStyle` used for?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddBiblioBlockController.java
```diff
@@ -0,0 +1,34 @@
...
+import edu.asu.diging.vspace.core.model.impl.BiblioBlock;
+import edu.asu.diging.vspace.core.services.IContentBlockManager;
+
+@Controller
+public class AddBiblioBlockController {
+
+    @Autowired
+    private IContentBlockManager contentBlockManager;
+
+    @RequestMapping(value = "/staff/module/{moduleId}/slide/{id}/bibliography", method = RequestMethod.POST)
```
jdamerow, line 25: since you use `reference/add` below, I would use `references` here

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DeleteBiblioBlockController.java
```diff
@@ -0,0 +1,80 @@
...
+    @Autowired
+    private IReferenceManager referenceManager;
+
+    @RequestMapping(value = "/staff/module/{moduleId}/slide/{id}/biblio/{blockId}", method = RequestMethod.DELETE)
+    public ResponseEntity<String> deleteBiblioBlock(@PathVariable("moduleId") String moduleId, @PathVariable("id") String slideId, 
+            @PathVariable("blockId") String blockId) throws IOException {
+        
+        ISlide slide = slideManager.getSlide(slideId);
+        IModule module = moduleManager.getModule(moduleId);
+        // If the slide Id and module Id is not found, show message on screen.
```
jdamerow, line 48: unnecessary, this is obvious from the code

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DeleteBiblioBlockController.java
```diff
@@ -0,0 +1,80 @@
...
+            logger.warn("Module Id does not exist, bad request.");
+            return new ResponseEntity<String>(HttpStatus.BAD_REQUEST);
+        }
+        
+        List<IReference> refList = referenceManager.getReferencesForBiblio(blockId);
+        
+        if(refList.size()>0) {
+            try {
+                referenceManager.deleteRefences(refList, blockId);
+            } catch (Exception e) {
```
jdamerow, line 64: does this method really throw `Exception`? It shouldn't. It should be much more specific.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DeleteBiblioBlockController.java
```diff
@@ -0,0 +1,80 @@
...
+        if(module==null) {
+            logger.warn("Module Id does not exist, bad request.");
+            return new ResponseEntity<String>(HttpStatus.BAD_REQUEST);
+        }
+        
+        List<IReference> refList = referenceManager.getReferencesForBiblio(blockId);
+        
+        if(refList.size()>0) {
+            try {
+                referenceManager.deleteRefences(refList, blockId);
```
jdamerow, line 63: for now I would not delete the references, just the block. they are two different entities that we want to manage independently of each other. maybe later we'll add an option to remove references with the block, but for now let's just remove the block.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DeleteReferenceController.java
```diff
@@ -0,0 +1,73 @@
...
+    private IReferenceManager referenceManager;
+
+    @RequestMapping(value = "/staff/module/{moduleId}/slide/{id}/biblio/{biblioId}/reference/{refId}", method = RequestMethod.DELETE)
+    public ResponseEntity<String> deleteReference(@PathVariable("moduleId") String moduleId, @PathVariable("id") String slideId, 
+            @PathVariable("biblioId") String biblioId, @PathVariable("refId") String refId) throws IOException {
+        
+        ISlide slide = slideManager.getSlide(slideId);
+        IModule module = moduleManager.getModule(moduleId);
+        IBiblioBlock biblio = contentBlockManager.getBiblioBlock(biblioId);
+        // If the slide Id, module Id and Biblio ID is not found, show message on screen.
```
jdamerow, line 47: unnecessary

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DeleteReferenceController.java
```diff
@@ -0,0 +1,73 @@
...
+        }
+        
+        if(biblio==null) {
+            logger.warn("Biblio Block Id does not exist, bad request.");
+            return new ResponseEntity<String>(HttpStatus.BAD_REQUEST);
+        }
+        
+        try {
+            referenceManager.deleteReferenceById(refId, biblioId);
+        } catch (Exception e) {
```
jdamerow, line 65: this should not throw exception

jdamerow (changes requested):

pradeep-017:
Closing this as we are reviewing this in https://github.com/diging/virtual-spaces-2.0/pull/219

jdamerow:
Please do not open pull requests with conflicts.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ReferenceManager.java
ajay3568yadav: isn't it still a good practice though?

vspace/src/main/java/edu/asu/diging/vspace/core/exception/ReferenceListDeletionForBiblioException.java
```diff
@@ -0,0 +1,30 @@
+package edu.asu.diging.vspace.core.exception;
+
+public class ReferenceListDeletionForBiblioException extends Exception {
+
+    /**
+     * 
+     */
+    private static final long serialVersionUID = 8621469001444367106L;
```
jdamerow, line 8: let's start at 1.

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/BiblioBlock.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+    
+    @Override
+    @Transient
+    public String toString() {
+        return "Bibliography title=" + biblioTitle + ", description=" + description ;
+    }
+    
+    @Transient
+    public String htmlRenderedBiblio() {
```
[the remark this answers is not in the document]
jdamerow: ?

jdamerow (changes requested):
please resolve code factor issues before I continue reviewing.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ContentBlockManager.java
```diff
@@ -463,6 +474,46 @@ public IChoiceBlock createChoiceBlock(String slideId, List<String> selectedChoic
         return choiceBlockRepo.save((ChoiceBlock) choiceBlock);
     }
 
+    @Override
+    public IBiblioBlock createBiblioBlock(String slideId, String title, String description, Integer contentOrder) {
+        
+
```
jdamerow, line 480: delete empty lines

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ContentBlockManager.java
```diff
@@ -266,4 +276,39 @@ public IChoiceBlock createChoiceBlock(String slideId, List<String> selectedChoic
...
+    }
+
+    @Override
+    public void deleteBiblioBlockById(String id) throws BlockDoesNotExistException {
+        if (id == null) {
+            return;
+        }
+        try {
+            biblioBlockRepo.deleteById(id);         
+        } catch (EmptyResultDataAccessException e) {
```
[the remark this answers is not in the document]
jdamerow: If I see this correctly, `deleteById` does not throw `EmptyResultDataAccessException` (see https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/repository/CrudRepository.html)

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ReferenceManager.java
jdamerow: I wouldn't (given it's a runtime exception and this case can never happen). I would just make a comment noting this.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ReferenceManager.java
```diff
@@ -0,0 +1,87 @@
...
+            logger.error("Unable to delete reference" + referenceId + ". ", exception);
+        }
+    }
+
+    @Override
+    public void deleteReferences(List<IReference> references, String biblioId) {
+        for(IReference ref : references) {
+            deleteReferenceById(ref.getId(), biblioId);
+        }
+
```
jdamerow, line 80: remove empty line

vspace/src/main/java/edu/asu/diging/vspace/referenceExpose/CitationStyleDefault.java
jdamerow: ?

vspace/src/main/java/edu/asu/diging/vspace/referenceExpose/ICitationStyle.java
jdamerow: ?

vspace/src/main/java/edu/asu/diging/vspace/referenceExpose/CitationStyleDefault.java
```diff
@@ -0,0 +1,7 @@
+package edu.asu.diging.vspace.referenceExpose;
```
jdamerow, line 1: bad package name, what is "reference expose"? Why not just "references"?

jdamerow (changes requested):
I stop here. There are unaddressed comments.

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/Reference.java
```diff
@@ -0,0 +1,213 @@
...
+    @GeneratedValue(generator = "reference_id_generator")
+    @GenericGenerator(name = "reference_id_generator", 
+        parameters = @Parameter(name = "prefix", value = "REF"),
+        strategy = "edu.asu.diging.vspace.core.data.IdGenerator")
+    private String id;
+    
+    @JsonIgnore
+    @ManyToMany
+    @JoinTable(name = "Biblio_Reference", joinColumns = @JoinColumn(name = "REFERENCE_ID"), inverseJoinColumns = @JoinColumn(name = "BIBLIO_ID"))
+
```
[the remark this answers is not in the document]
ajay3568yadav: what do you mean by "both" there is JsonIgnore and ManyToMany, are you talking about that??

vspace/src/main/java/edu/asu/diging/vspace/referenceExpose/CitationStyleDefault.java
ajay3568yadav: it was not being used anywhere else, not even for the references directly or in directly so I deleted that.

vspace/src/main/java/edu/asu/diging/vspace/referenceExpose/ICitationStyle.java
ajay3568yadav: it seems to be implemented by someone else and wasn't useful for the references thus I removed that

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DeleteReferenceController.java
```diff
@@ -0,0 +1,68 @@
...
+
+    @RequestMapping(value = "/staff/module/{moduleId}/slide/{id}/biblio/{biblioId}/reference/{refId}", method = RequestMethod.DELETE)
+    public ResponseEntity<String> deleteReference(@PathVariable("moduleId") String moduleId, @PathVariable("id") String slideId, 
+            @PathVariable("biblioId") String biblioId, @PathVariable("refId") String refId) throws IOException {
+        
+        ISlide slide = slideManager.getSlide(slideId);
+        IModule module = moduleManager.getModule(moduleId);
+        IBiblioBlock biblio = contentBlockManager.getBiblioBlock(biblioId);
+
+        if(slide==null) {
```
jdamerow, line 48: same as before (404)

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditBiblioBlockController.java
```diff
@@ -0,0 +1,47 @@
...
+
+        ObjectMapper mapper = new ObjectMapper();
+        JsonNode rootNode = mapper.readTree(biblioBlockData);
+
+        // Extract biblioTitle and description from the JSON
+        String biblioTitle = rootNode.get("biblioTitle").asText();
+        String description = rootNode.get("description").asText();
+        String id = rootNode.get("id").asText();
+        
+        IBiblioBlock biblioBlock = contentBlockManager.getBiblioBlock(id);
```
jdamerow, line 39: this needs to throw a 404 if the block does not exist

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditReferenceController.java
```diff
@@ -0,0 +1,65 @@
...
+        String url = rootNode.get("url").asText();
+        String volume = rootNode.get("volume").asText();
+        String issue = rootNode.get("issue").asText();
+        String pages = rootNode.get("pages").asText();
+        String editor = rootNode.get("editors").asText();
+        String type = rootNode.get("type").asText();
+        String note = rootNode.get("note").asText();
+        String id = rootNode.get("id").asText();
+        
+        IReference reference = referenceManager.getReference(id);
```
jdamerow, line 48: throw 404 if not exist

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/Reference.java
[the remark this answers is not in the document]
jdamerow: this is still not addressed?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ContentBlockManager.java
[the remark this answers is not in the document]
jdamerow: nothing has changed?

jdamerow (changes requested):
there are unaddressed comments

jdamerow (changes requested):
there are unaddressed comments

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/Reference.java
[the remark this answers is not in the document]
ajay3568yadav: removed @JsonIgnore

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditReferenceController.java
ajay3568yadav: added HttpStatus.NOT_FOUND

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/Reference.java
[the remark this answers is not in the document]
jdamerow: > remove empty line; 

Please read comments carefully.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ContentBlockManager.java
```diff
@@ -463,6 +469,44 @@ public IChoiceBlock createChoiceBlock(String slideId, List<String> selectedChoic
...
+    @Override
+    public void deleteBiblioBlockById(String id) throws BlockDoesNotExistException {
+        if (id == null) {
+            return;
+        }
+        
+        try {
+            biblioBlockRepo.deleteById(id);         
+        } catch (IllegalArgumentException e) {
+            throw new IllegalArgumentException(e);
```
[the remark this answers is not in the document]
jdamerow: Made a comment about this somewhere. No point in wrapping an `IllegalArgumentException` into another `IllegalArgumentException`.

jdamerow (changes requested):
unaddressed comments

jdamerow:
There are more than what I just commented on. Please review the whole PR carefully and don't open it again until everything has been addressed.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddBiblioBlockController.java
ajay3568yadav: i do not see "reference/add" so i think this is already addressed

jdamerow: `AddReferenceController` uses `references/add`, so here it should be `references` not `bilbiography`

ajay3568yadav: this is AddBiblioBlockController, so shouldn't it be just bibliography? it add bibliography block to the slides

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditBiblioBlockController.java
ajay3568yadav: Added HTTP.NOT_FOUND

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ContentBlockManager.java
```diff
@@ -463,6 +469,44 @@ public IChoiceBlock createChoiceBlock(String slideId, List<String> selectedChoic
...
+    @Override
+    public void deleteBiblioBlockById(String id) throws BlockDoesNotExistException {
+        if (id == null) {
+            return;
+        }
+        
+        try {
+            biblioBlockRepo.deleteById(id);         
+        } catch (IllegalArgumentException e) {
+            System.err.println("Failed to delete biblio block: " + e.getMessage());
```
jdamerow, line 492: This is not how exceptions should be handled. Ever. I'm fairly certain we had a bi-weekly about this.

Exceptions need to either:
- be handled (logically that is, meaning is there a direct consequence that should happen based on the exception? Or can we just log and return? Log and ignore? note that *log* here means actual logging, system out is not logging, since you never know where those messages end up (if they end up somewhere at all))
- thrown up to the next level for them to handle. If that's the case then is it a general enough exception that can be thrown up without causing tight coupling? if so then just add a `throws` with the type of exception and that's it. If it would cause tight coupling, then wrap the exception in another exception of a type that makes sense to through here (e.g. `BlockDoesNotExistException` which this method throws according to its signature).

vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddBiblioBlockController.java
jdamerow: ah yes! that makes sense.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DeleteBiblioBlockController.java
```diff
@@ -0,0 +1,64 @@
...
+    @Autowired
+    private IContentBlockManager contentBlockManager;
+    
+    @Autowired
+    private IModuleManager moduleManager;
+    
+    @Autowired
+    private ISlideManager slideManager;
+    
+    @RequestMapping(value = "/staff/module/{moduleId}/slide/{id}/biblio/{blockId}", method = RequestMethod.DELETE)
```
jdamerow, line 37: then this should be `.../slide/{id}/bibliography/...` here too

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DeleteReferenceController.java
jdamerow: not sure what these means?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DeleteReferenceController.java
```diff
@@ -0,0 +1,64 @@
...
+    @Autowired
+    private IModuleManager moduleManager;
+    
+    @Autowired
+    private ISlideManager slideManager;
+    
+    @Autowired
+    private IReferenceManager referenceManager;
+
+    @RequestMapping(value = "/staff/module/{moduleId}/slide/{id}/biblio/{biblioId}/reference/{refId}", method = RequestMethod.DELETE)
```
jdamerow, line 40: `blibliography` as above

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditBiblioBlockController.java
```diff
@@ -0,0 +1,51 @@
...
+import edu.asu.diging.vspace.core.model.impl.BiblioBlock;
+import edu.asu.diging.vspace.core.services.IContentBlockManager;
+
+@Controller
+public class EditBiblioBlockController {
+
+    @Autowired
+    private IContentBlockManager contentBlockManager;
+
+    @RequestMapping(value = "/staff/module/{moduleId}/slide/{id}/biblio/edit", method = RequestMethod.POST)
```
jdamerow, line 27: same here

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditReferenceController.java
```diff
@@ -0,0 +1,69 @@
...
+import edu.asu.diging.vspace.core.model.impl.Reference;
+import edu.asu.diging.vspace.core.services.IReferenceManager;
+
+@Controller
+public class EditReferenceController {
+
+    @Autowired
+    private IReferenceManager referenceManager;
+
+    @RequestMapping(value = "/staff/module/{moduleId}/slide/{id}/biblio/{biblioId}/reference/edit", method = RequestMethod.POST)
```
jdamerow, line 27: same here

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditReferenceController.java
jdamerow: still not sure what you mean. do you mean "updated" not "outdated"?

vspace/src/main/webapp/WEB-INF/views/staff/modules/slides/slide.html
```diff
@@ -1,1724 +1,2933 @@
...
+         background: solid var(--primary);
+         color: #fff;
+         font-size: 15px;
+         }
+         a.nav-link.active.show {
+         border-color: transparent;
+         border-bottom: 4px solid var(--primary);
+         color: var(--primary) !important;
+         background: transparent;
+         }
```
jdamerow, line 34: where did all the indentation go? and did anything change here? please don't reformat or I can' see what changed.

vspace/src/main/webapp/WEB-INF/views/staff/modules/slides/slide.html
jdamerow, line 1: so many changes, and I have no clue what actually changed and needs my attention

vspace/src/main/webapp/resources/extra/Home.css
jdamerow, line 1: why does this whole file show up as changed?

jdamerow (changes requested):

vspace/src/main/webapp/resources/extra/Home.css
ajay3568yadav: in this commit, the changes from the develop branch was merged into VSPC-147: https://github.com/diging/virtual-spaces-2.0/commit/f8c00cbc4d918ceb46c7689543fd5b69cc954b52

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditReferenceController.java
ajay3568yadav: I have added HTTP.NOT_FOUND but this is not showing up here, if you do not see this in this file then i guess there is something wrong, that its not showing the change I made in this file, because I addressed this comment.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ContentBlockManager.java
ajay3568yadav: I didn't write this piece of code but I'll make changes to this, by adding logger, let me know if that is good enough

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DeleteReferenceController.java
ajay3568yadav: I added HTTP.NOT_FOUND, but if you don't see that while reviewing this then I guess there is something wrong because i dont see it here in but it is in the VSPC-147 branch's `vspace/src/main/java/edu/asu/diging/vspace/web/staff/DeleteReferenceController.java` file

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ContentBlockManager.java
```diff
@@ -463,6 +469,47 @@ public IChoiceBlock createChoiceBlock(String slideId, List<String> selectedChoic
...
+    @Override
+    public void deleteBiblioBlockById(String id) throws BlockDoesNotExistException {
+        if (id == null) {
+            logger.warn("Attempted to delete biblio block with null id.");
+            return;
+        }
+
+        try {
+            biblioBlockRepo.deleteById(id);
+            logger.info("Successfully deleted biblio block with id: {}", id);
```
jdamerow, line 492: remove

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ContentBlockManager.java
```diff
@@ -463,6 +469,47 @@ public IChoiceBlock createChoiceBlock(String slideId, List<String> selectedChoic
...
+        } catch (IllegalArgumentException e) {
+            logger.error("Failed to delete biblio block with id: {}. Reason: {}", id, e.getMessage());
```
jdamerow, line 494: full stracktrace needs to be logged

jdamerow: actually never mind, if it's thrown nothing should be logged here

vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddBiblioBlockController.java
```diff
@@ -0,0 +1,49 @@
...
+
+
+@Controller
+public class AddBiblioBlockController {
+
+    @Autowired
+    private IContentBlockManager contentBlockManager;
+
+    @RequestMapping(value = "/staff/module/{moduleId}/slide/{id}/bibliography", method = RequestMethod.POST)
+    public ResponseEntity<String> addTextBlock(@PathVariable("id") String slideId,
```
jdamerow, line 32: why is this method called `addTextBlock`?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DeleteReferenceController.java
jdamerow: hm, seems to show up now!

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
There are test failures.

jdamerow (changes requested):
unaddressed comments

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ContentBlockManager.java
ajay3568yadav: removed logger, thinking that is what you wanted

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditReferenceController.java
ajay3568yadav: Screenshot 2024-09-25 at 15 32 40
can you see HTTP.NOT_FOUND line

ajay3568yadav: Is this the unaddressed you commented about, last? or can you see the HTTPStatus.NOT_FOUND on your side, because it is not showing up on the review side for me but it is there in the branch's respective file.

ajay3568yadav:
> unaddressed comments

I looked at the review comments and all of them show "outdated",(even the ones that were collapsed) thus I feel that they have been addressed, Let me know if there is any comment that is not unaddressed. there should not be any i hope.

jdamerow:
Make it so, Jenkins.

jdamerow:
creating a new bibliography block throws an exception:
```
Java 8 date/time type `java.time.OffsetDateTime` not supported by default: add Module "com.fasterxml.jackson.datatype:jackson-datatype-jsr310" to enable handling (through reference chain: edu.asu.diging.vspace.core.model.impl.BiblioBlock["creationDate"])
```
same after adding a reference

and after adding a reference, this happens on the public side (and the rest of the slide is not shown):
image

ajay3568yadav:
I am not getting this exception on my side what do you think could be the reason

jdamerow:
Have you tried this? 
https://howtodoinjava.com/jackson/java-8-date-time-type-not-supported-by-default/

It could also be that it's running under Java 17 on staging.

Also, though, the presentation of bibliography blocks on the public pages is not fixed.

vspace/pom.xml
```diff
@@ -319,7 +319,12 @@
             <artifactId>jackson-databind</artifactId>
             <version>2.12.6.1</version>
         </dependency>
-
+        
+        <dependency>
+		    <groupId>com.fasterxml.jackson.datatype</groupId>
+		    <artifactId>jackson-datatype-jsr310</artifactId>
+		    <version>2.10.0.pr1</version>
+		</dependency>
```
jdamerow, line 327: indentation

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/BiblioBlock.java
```diff
@@ -0,0 +1,74 @@
...
+        this.references = references;
+    }
+
+    @Override
+    public List<Reference> getReferences() {
+        return references;
+    }
+    
+    @Override
+    public String toString() {
```
jdamerow, line 62: this is not what `toString` should be used for.

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/BiblioBlock.java
```diff
@@ -0,0 +1,74 @@
...
+        return "**Bibliography title:** " + biblioTitle + "\n**description:** " + description ;
+    }
+    
+    public String htmlRenderedBiblio() {
+        Parser parser = Parser.builder().build();
+        String formattedText = toString().replace("\n", "<br>");
```
jdamerow, line 68: I'm definitely not in love with this solution. this mixes display information with business logic. how a block is formatted should be done in the template, not in the model class .

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/Reference.java
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return  title + ", " + author + ", " + year + ", " + journal + ", " + url
+                + ", " + volume + ", " + issue + ", " + pages + ", " + editors + ", "
+                + type + ", " + note;
+    }
+
+    public String htmlRenderedReference() {
```
jdamerow, line 174: what is this for?

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/BiblioBlock.java
```diff
@@ -0,0 +1,73 @@
...
+    @Override
+    public String toString() {
+        return "bibliography title: " + biblioTitle + " description: " + description ;
+    }
+    
+    public String htmlRenderedBiblio() {
+        Parser parser = Parser.builder().build();
+        Node document = parser.parse(toString());
+        HtmlRenderer renderer = HtmlRenderer.builder().build();
+        return renderer.render(document);
```
jdamerow, line 70: what is this for?

jdamerow (changes requested):
the issue with toString and the html method is still in there. any kind of formatting should happen in the template or through external classes, not the model class.

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/Reference.java
ajay3568yadav: it parses the text returned by toString() (which in this case is "bibliography title: " + biblioTitle + " description: " + description) as Markdown, and then render that Markdown as HTML. it is used in modules for biblio block and reference block to render "Biblio block" and references, its just returning the string
Screenshot 2025-02-07 at 2 53 50 PM

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/BiblioBlock.java
ajay3568yadav: Same as before, is this how it should be done? Or should I change it? If yes, then let me know how

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/Reference.java
jdamerow: ya, that should not be done in the domain model class. that needs to be done in the template or through an extra class.

ajay3568yadav: okay

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/BiblioBlock.java
ajay3568yadav: addressed comment

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditReferenceController.java
```diff
@@ -0,0 +1,69 @@
...
+import edu.asu.diging.vspace.core.model.impl.Reference;
+import edu.asu.diging.vspace.core.services.IReferenceManager;
+
+@Controller
+public class EditReferenceController {
+
+    @Autowired
+    private IReferenceManager referenceManager;
+
+    @RequestMapping(value = "/staff/module/{moduleId}/slide/{id}/bibliography/{biblioId}/reference/edit", method = RequestMethod.POST)
```
jdamerow, line 27: should be consistent, either
`.../reference/edit` and `.../reference/add`
or
`.../references/edit` and `.../references/add`

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditReferenceController.java
```diff
@@ -0,0 +1,69 @@
...
+        String issue = rootNode.get("issue").asText();
+        String pages = rootNode.get("pages").asText();
+        String editor = rootNode.get("editors").asText();
+        String type = rootNode.get("type").asText();
+        String note = rootNode.get("note").asText();
+        String id = rootNode.get("id").asText();
+        
+        IReference reference = referenceManager.getReference(id);
+        if (reference == null) {
+            return new ResponseEntity<String>(HttpStatus.NOT_FOUND);
```
jdamerow, line 50: needs to log why this is returned

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditReferenceController.java
```diff
@@ -0,0 +1,69 @@
...
+        reference.setYear(year);
+        reference.setJournal(journal);
+        reference.setUrl(url);
+        reference.setVolume(volume);
+        reference.setIssue(issue);
+        reference.setPages(pages);
+        reference.setEditors(editor);
+        reference.setType(type);
+        reference.setNote(note);
+        referenceManager.updateReference((Reference) reference);
```
jdamerow, line 64: it should not be necessary to cast to `Reference` here

vspace/src/main/webapp/WEB-INF/views/staff/modules/slides/slide.html
```diff
@@ -48,10 +49,12 @@
             var contentCount = [[${#lists.size(slideContents)}]];
             var markDown = null;
             var defaultValue;
+	        var formDataCancel;
```
jdamerow, line 52: indentation

vspace/src/main/webapp/WEB-INF/views/staff/modules/slides/slide.html
```diff
@@ -225,8 +228,89 @@
...
+                $(".open").css('overflow-y', 'scroll');
+                $(".open").css('max-height', '200px');	
+             } 
+             
+             function referenceBlockDoubleClick(e){
+            	 	currentRefId = $(this).find('#deleteReferenceId').val();
+            	    $(".EasyMDEContainer").remove();
+	         	    $(e.target).closest('.referenceDivStaff').addClass("open");
+	         	    var description = $("div.open p:eq(0)").text();
+             	    defaultValue = description;
```
jdamerow, line 260: what's up with the indentation in the couple of lines above?

vspace/src/main/webapp/WEB-INF/views/staff/modules/slides/slide.html
```diff
@@ -225,8 +228,89 @@
...
+                    var editors = $("div.open p span:eq(8)").text();
+                    var type = $("div.open p span:eq(9)").text();
+                    var note = $("div.open p span:eq(10)").text(); 
+             	
+             	    //showing the edit bibliography form with all the respective values
+		    	    $("#editReferenceAlert").show();
+		    	    var alert = $('<input type="hidden" id="currentBiblioId">');
+		    		/* $(alert).attr( 'value', $(this).parent().parent().parent().attr("id"));*/
+		    		$(alert).attr( 'value', $(e.target).closest('.biblioDivStaff').attr("id"));
+		    		$('.modal-footer').append(alert);
```
jdamerow, line 279: indentation (all the above)

vspace/src/main/webapp/WEB-INF/views/staff/modules/slides/slide.html
```diff
@@ -225,8 +228,89 @@
...
+                    var issue = $("div.open p span:eq(6)").text();
+                    var pages = $("div.open p span:eq(7)").text();
+                    var editors = $("div.open p span:eq(8)").text();
+                    var type = $("div.open p span:eq(9)").text();
+                    var note = $("div.open p span:eq(10)").text(); 
+             	
+             	    //showing the edit bibliography form with all the respective values
+		    	    $("#editReferenceAlert").show();
+		    	    var alert = $('<input type="hidden" id="currentBiblioId">');
+		    		/* $(alert).attr( 'value', $(this).parent().parent().parent().attr("id"));*/
```
jdamerow, line 277: remove if not needed

vspace/src/main/webapp/WEB-INF/views/staff/modules/slides/slide.html
```diff
@@ -1071,50 +1291,240 @@
...
+                     data: jsonData,
+                     success: function(data) {
+                     	data = JSON.parse(data);
+                     	var referenceblock = $('<div id="'+ data.id +'" class="referenceDivStaff tab row"><p class="col-md-10" > Reference Title: <span>' + data.title + '</span>, Author: <span>' + data.author + '</span>, Year:' 
+                         		+ '<span>' + data.year + '</span>, Journal: <span>' + data.journal + '</span>, '
+                         		+ 'Url: <span>' + data.url + '</span>, Volume: <span>' + data.volume + '</span>, '
+         						+ 'Issue: <span>' + data.issue + '</span>, Pages: <span>' + data.pages + '</span>, '
+         						+ 'Editors: <span>' + data.editors + '</span>, Type: <span>' + data.type + '</span>, '
+         						+ 'Note: <span>' + data.note + '</span>'
+                         		+ '<input type="hidden" id="deleteReferenceId" value="'+ data.id +'" /><a class="btn deleteReference" href="javascript:;" style="position: absolute; right:0px;"><i style="color: var(--error);" class="fas fa-trash-alt"></i></a></p></div>');
```
jdamerow, line 1509: doesn't this mean that every time I add a new reference, the new div will have an input with the same id? IDs have to be unique.

vspace/src/main/webapp/WEB-INF/views/staff/modules/slides/slide.html
```diff
@@ -225,8 +228,89 @@
...
+                    var pages = $("div.open p span:eq(7)").text();
+                    var editors = $("div.open p span:eq(8)").text();
+                    var type = $("div.open p span:eq(9)").text();
+                    var note = $("div.open p span:eq(10)").text(); 
+             	
+             	    //showing the edit bibliography form with all the respective values
+		    	    $("#editReferenceAlert").show();
+		    	    var alert = $('<input type="hidden" id="currentBiblioId">');
+		    		/* $(alert).attr( 'value', $(this).parent().parent().parent().attr("id"));*/
+		    		$(alert).attr( 'value', $(e.target).closest('.biblioDivStaff').attr("id"));
```
jdamerow, line 278: this seems unnecessarily complicated and brittle. 
a) why is the variable called "alert"? It does not contain an alert, does it?
b) if I understand this right, all that this code is trying to do is that the id of the reference that should be deleted in the delete form, right? If so, the block that was clicked on should just have a data attribute that holds the id of the reference that should be deleted. Then get that value here and set it in the input. 
c) The input element should probably be already in the modal, and then just set the value here. No need to create a new input element every time.

jdamerow: The same is true for all the other cases where there is similar code here in the file.

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditReferenceController.java
ajay3568yadav: Addressed, but I don't see the newly added line on the review side.

vspace/src/main/webapp/WEB-INF/views/staff/modules/slides/slide.html
ajay3568yadav: the ids are different

ajay3568yadav: Screenshot 2025-03-11 at 4 40 04 PM

Screenshot 2025-03-11 at 4 40 22 PM

ids are different for difference references

ajay3568yadav: Screenshot 2025-03-11 at 4 42 33 PM
for the input as well, the values of the unique id is being stored, and is called while deleting it

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ReferenceManager.java
```diff
@@ -0,0 +1,84 @@
...
+        reference.setYear(year);
+        reference.setJournal(journal);
+        reference.setUrl(url);
+        reference.setVolume(volume);
+        reference.setIssue(issue);
+        reference.setPages(pages);
+        reference.setEditors(editor);
+        reference.setType(type);
+        reference.setNote(note);
+        reference.getBiblios().add((BiblioBlock) biblio);
```
jdamerow, line 46: It should not be necessary to cast to BiblioBlock here

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ReferenceManager.java
```diff
@@ -0,0 +1,84 @@
...
+    @Autowired
+    private ReferenceRepository referenceRepo;
+    
+    @Autowired
+    private IContentBlockManager contentBlockManager;
+
+    @Override
+    public IReference createReference(String biblioId, String title, String author,String year,String journal, String url, String volume,String issue, String pages,String editor, String type, String note) {
+    	IReference reference = new Reference();
+        IBiblioBlock biblio = contentBlockManager.getBiblioBlock(biblioId);
```
jdamerow, line 34: this line should be right before line 46 or even merged

vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddBiblioBlockController.java
```diff
@@ -0,0 +1,51 @@
...
+
+@Controller
+public class AddBiblioBlockController {
+
+    @Autowired
+    private IContentBlockManager contentBlockManager;
+
+    @RequestMapping(value = "/staff/module/{moduleId}/slide/{id}/bibliography", method = RequestMethod.POST)
+    public ResponseEntity<String> addBiblioBlock(@PathVariable("id") String slideId,
+            @PathVariable("moduleId") String moduleId, @RequestBody String biblioBlockData) throws JsonProcessingException {
```
jdamerow, line 33: Can't you have spring parse the json into an object (just create a BiblioBlockData class with the necessary fields and use it here)? This way you can drop line 35-41.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddBiblioBlockController.java
```diff
@@ -0,0 +1,51 @@
...
+        // Parse the JSON string using ObjectMapper
+        ObjectMapper mapper = new ObjectMapper();
+        JsonNode rootNode = mapper.readTree(biblioBlockData);
+
+        // Extract biblioTitle and description from the JSON
+        String biblioTitle = rootNode.get("biblioTitle").asText();
+        String description = rootNode.get("description").asText();
+        Integer contentOrder = contentBlockManager.findMaxContentOrder(slideId);
+        contentOrder = contentOrder == null ? 0 : contentOrder + 1;
+        System.out.println("/n/n/n/n"+contentOrder + "/n/n/n/n/n");
```
jdamerow, line 44: ?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddReferenceController.java
```diff
@@ -0,0 +1,53 @@
...
+@Controller
+public class AddReferenceController {
+
+    @Autowired
+    private IReferenceManager referenceManager; 
+
+    @RequestMapping(value = "/staff/module/{id}/slide/{slideId}/bibliography/{biblioId}/reference/add", method = RequestMethod.POST)
+    public ResponseEntity<String> addReference(@PathVariable("id") String moduleId, @PathVariable("slideId") String slideId, 
+            @PathVariable("biblioId") String biblioId, 
+            @RequestBody String reference, Model model) throws JsonProcessingException {
```
jdamerow, line 32: same here, let spring do the parsing.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditBiblioBlockController.java
```diff
@@ -0,0 +1,50 @@
...
+        JsonNode rootNode = mapper.readTree(biblioBlockData);
+        
+        String biblioTitle = rootNode.get("biblioTitle").asText();
+        String description = rootNode.get("description").asText();
+        String id = rootNode.get("id").asText();
+        
+        IBiblioBlock biblioBlock = contentBlockManager.getBiblioBlock(id);
+        
+        if (biblioBlock == null) {
+            return new ResponseEntity<String>(HttpStatus.NOT_FOUND);
```
jdamerow, line 40: needs to add at least a warning to the log

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditBiblioBlockController.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @RequestMapping(value = "/staff/module/{moduleId}/slide/{id}/bibliography/edit", method = RequestMethod.POST)
+    public ResponseEntity<String> editTextBlock(@PathVariable("id") String slideId,
+            @PathVariable("moduleId") String moduleId, @RequestBody String biblioBlockData) throws IOException {
+        ObjectMapper mapper = new ObjectMapper();
+        JsonNode rootNode = mapper.readTree(biblioBlockData);
+        
+        String biblioTitle = rootNode.get("biblioTitle").asText();
+        String description = rootNode.get("description").asText();
+        String id = rootNode.get("id").asText();
```
jdamerow, line 35: same as above

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditReferenceController.java
jdamerow: codefactor.io issue

vspace/src/main/webapp/WEB-INF/views/staff/modules/slides/slide.html
jdamerow: Isn't this code creating an input field for every reference with the id "deleteReferenceId" (and a unique value)? This is what the code looks like, and if that is the case, then the id is not unique (it's "deleteReferenceId" for every hidden input field).

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditBiblioBlockController.java
ajay3568yadav: addressed this line logger.warn("No 'id' field found in the JSON data."); but the comment does not display as outdated.

vspace/src/main/webapp/WEB-INF/views/staff/modules/slides/slide.html
```diff
@@ -928,24 +1041,125 @@
...
+         		
+         		$("#deleteReference").on("click", function(e) {
+         		e.preventDefault();
+         		$("#confirmDeleteTextAlert").find('input').remove();
+         		$("#confirmDeleteImageAlert").find('input').remove();
+         		$("#confirmDeleteBiblioAlert").find('input').remove();
+         		$("#confirmDeleteChoiceBlockAlert").find('input').remove();
+         		$("#confirmDeleteRefAlert").hide();
+         		var refId = $('.deleteReferenceId').attr('value');
+         		$('.deleteReferenceId').remove();
```
jdamerow, line 1131: I haven't tested this yet, but what I expect:
- I create multiple references, which means multiple `<input type="hidden" class="deleteReferenceId" ...>` will be created.
- the above lines will probably take the first input they find with class `deleteReferenceId`
- from that input, the id is used to delete a reference.
That can't be right?

jdamerow (changes requested):
still codefactor issues

vspace/src/main/webapp/WEB-INF/views/staff/modules/slides/slide.html
ajay3568yadav: deletetion is working properly, I checked with different combinations

vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddReferenceController.java
```diff
@@ -0,0 +1,48 @@
...
+            referenceData.getUrl(), 
+            referenceData.getVolume(), 
+            referenceData.getIssue(), 
+            referenceData.getPages(), 
+            referenceData.getEditors(), 
+            referenceData.getType(), 
+            referenceData.getNote()
+        );
+        
+        return ResponseEntity.ok((Reference) ref);
```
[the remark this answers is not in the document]
ajay3568yadav: its not working if I am passing the IReference. Type mismatch: cannot convert from ResponseEntity<IReference> to ResponseEntity<Reference>
why do are we avoiding TypeCasting?

ajay3568yadav:
Screenshot 2025-03-26 at 4 36 13 PM
I don't see any codefactor issues?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddReferenceController.java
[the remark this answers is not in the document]
jdamerow: I don't get such an error. Did you change the method signature appropriately when you removed the casting?

If you cast to `Reference`, you're locked in to using that specific implementation of `IReference`. If we need to change it for some reason later, you'll have to change it here as well, and even worse if you have two different implementations of IReference that you might want to use depending on the scenario, you wouldn't be able to do that if you cast here. (Casting is usually a sign that something could be improved).

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditReferenceController.java
```diff
@@ -0,0 +1,74 @@
...
+        String editor = rootNode.get("editors").asText();
+        String type = rootNode.get("type").asText();
+        String note = rootNode.get("note").asText();
+        String id = rootNode.get("id").asText();
+        
+        IReference reference = referenceManager.getReference(id);
+        if (reference == null) {
+        	logger.warn("Reference with ID '{}' not found, returning NOT_FOUND.", id);
+        	return ResponseEntity.notFound().build();
+
```
jdamerow, line 55: here are code factor issues.

jdamerow: tabs instead of blanks I believe

vspace/src/main/webapp/WEB-INF/views/staff/modules/slides/slide.html
```diff
@@ -928,24 +1041,147 @@
...
+         		// ---------- Delete Bibliography Block -------------
+         		
+         		$("#deleteBiblio").on("click", function(e) {
+         		e.preventDefault();
+         		$("#confirmDeleteTextAlert").find('input').remove();
+         		$("#confirmDeleteImageAlert").find('input').remove();
+         		$("#confirmDeleteRefAlert").find('input').remove();
+         		$("#confirmDeleteChoiceBlockAlert").find('input').remove();
+         		$("#confirmDeleteBiblioAlert").hide();
+         		var blockId = $('#deleteBiblioId').attr('value');
```
jdamerow, line 1097: we talked about this in standup

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditReferenceController.java
jdamerow: ?

PR #232, opened 2021-07-28
Vspc 165

adhvikmadhav:

- [ ] I have explained any part of my code/implementation decisions that is not be self-explanatory

...Put description here...
  
... describe here ...


vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddSlideController.java
```diff
@@ -47,13 +48,17 @@ public String addSlide(Model model, @PathVariable("moduleId") String moduleId, @
...
         SlideType type = slideForm.getType().isEmpty() ? null : SlideType.valueOf(slideForm.getType());
+        String slideId;
         if(type.equals(SlideType.BRANCHING_POINT)) {
             IBranchingPoint branchingPoint = slideManager.createBranchingPoint(module, slideForm, type);
             IChoiceBlock choiceBlock = contentBlockManager.createChoiceBlock(branchingPoint.getId(), null, 0, true);
-        } else {
-            slideManager.createSlide(module, slideForm, type);
+            slideId = branchingPoint.getId();
+        } 
+        else {
```
jdamerow, line 57: else should be on same line as }

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
- the help should only be shown automatically the first time the user creates as slide, after that there should be a button to start it
- I can only see the help for the slide creation but nothing for the module or sequences

jdamerow:
Resolve conflicts please

vspace/.project
```diff
@@ -50,4 +50,15 @@
...
+		<filter>
+			<id>1629823327433</id>
+			<name></name>
+			<type>30</type>
+			<matcher>
+				<id>org.eclipse.core.resources.regexFilterMatcher</id>
+				<arguments>node_modules|.git|__CREATED_BY_JAVA_LANGUAGE_SERVER__</arguments>
+			</matcher>
+		</filter>
+	</filteredResources>
```
jdamerow, line 63: this should not be needed here and not commited

vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddModuleController.java
```diff
@@ -32,6 +33,8 @@ public String showAddModule(Model model) {
 	public String addModule(Model model, @ModelAttribute ModuleForm moduleForm, Principal principal) {
 	    IModule module = moduleFactory.createModule(moduleForm);
 	    moduleManager.storeModule(module);
-	    return "redirect:/staff/module/list";
+        String moduleId;
+        moduleId = module.getId();
+        return "redirect:/staff/module/" + moduleId;
```
jdamerow, line 38: merge with line 37

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
- "Add or remove existing slides in your sequence. You can change the order of slides too." should be "Create sequences to set the order of slides. You can add or remove existing slides or you can change their order in a sequence."
- "Keep track of your location in the module or navigate back up the order." should be "Keep track of your location in the module or navigate back."
- "Looks like your slide is empty. Add a text block/image." remove "Looks like your slide is empty." since this is always shown even if the slide is not empty and make it "Add content to your slide. The content will be shown in the public exhibition in the order you define here."

jdamerow:
the tour always seems to show when a new sequence (and I guess slide or module) has been created. it should only be shown once.

Also:
- "Create slides you want to show in your Module." should be "Create slides you want to show in your Module. You can create two types of slides: slides and branching points. A slide is like a power point slide, it can contain images, texts, and other content. A branching point is a slide that can link to other sequences in the module giving the user options what to look at next."

jdamerow:
Make it so, Jenkins.

PR #248, opened 2021-08-27
Bug/vspc 185

glend17:

When a module has a start sequence that is set but the module itself does not have any slides, an exception 'FileNotFoundException' is thrown. This breaks the front-end view. Instead a module template is adopted to the User to let them know that they have assigned a sequence to a module that has no slides.
  
When reproducing the issue, there was an issue with the navbar as it was missing `?` after startSpace. This was causing the code to break and I have added this change in. I have added it as a seperate commit on this branch.

### DIFF
--- a/vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSequencesController.java
+++ b/vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSequencesController.java
@@ -65,7 +65,9 @@ public class ExhibitionSequencesController {
 
         List<ISlide> slides = sequenceManager.getSequence(sequenceId).getSlides();
         if (slides.size() == 0) {
-            throw new SlidesInSequenceNotFoundException();
+        	model.addAttribute("showAlert", true);
+            model.addAttribute("message", "Sorry, slides haven't been added to this module yet.");
+            return "/exhibition/module";
         }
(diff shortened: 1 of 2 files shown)


vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSequencesController.java
```diff
@@ -65,7 +65,9 @@ public String sequence(Model model, @PathVariable("sequenceId") String sequenceI
 
         List<ISlide> slides = sequenceManager.getSequence(sequenceId).getSlides();
         if (slides.size() == 0) {
-            throw new SlidesInSequenceNotFoundException();
+            model.addAttribute("showAlert", true);
+            model.addAttribute("message", "Sorry, slides haven't been added to this module yet.");
```
jdamerow, line 69: message should say "Sorry, module has not been configured yet."

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSequencesController.java
glend17: Addressed this and have changed the message to appropriate template.

jdamerow:
Make it so, Jenkins.

PR #255, opened 2021-09-13
[VSPC-173] Added message box if space is unpublished for Staff

glend17:

https://diging.atlassian.net/browse/VSPC-173
  
Have verified that users who are not staff or admin will be redirected to /404


jdamerow:
Make it so, Jenkins.

vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSpaceController.java
```diff
@@ -53,11 +53,16 @@
     public String space(@PathVariable("id") String id, Model model) {
         ISpace space = spaceManager.getSpace(id);
         List<ISpaceLinkDisplay> spaceLinks;
+        Boolean isSpacePublished = true;
         /* (non-Javadoc)
          * Below null check is added to accommodate already existing spaces with null space status
          */
         if (space.getSpaceStatus() == null || space.getSpaceStatus().equals(SpaceStatus.PUBLISHED)
                 || authenticationFacade.getAuthenticatedUser() != null) {
+            if(space.getSpaceStatus().equals(SpaceStatus.UNPUBLISHED)) {
```
jdamerow, line 62: this throws a nullpointer if `getSpaceStatus` returns null.

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSpaceController.java
glend17: In line 62 I have added a null check space.getSpaceStatus() == null. This will check if the space status is null, if it is then the admin or staff can still access the space in exhibition and therefore we must have the status isSpacePublished as false to prompt the admin/staff to publish the space.
I have verified that when we go to the exhibition directly without logging in the space throws a 404 error

vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionSpaceController.java
```diff
@@ -53,11 +53,16 @@
     public String space(@PathVariable("id") String id, Model model) {
         ISpace space = spaceManager.getSpace(id);
         List<ISpaceLinkDisplay> spaceLinks;
+        Boolean isSpacePublished = true;
         /* (non-Javadoc)
          * Below null check is added to accommodate already existing spaces with null space status
          */
         if (space.getSpaceStatus() == null || space.getSpaceStatus().equals(SpaceStatus.PUBLISHED)
                 || authenticationFacade.getAuthenticatedUser() != null) {
+            if(space.getSpaceStatus() == null || space.getSpaceStatus().equals(SpaceStatus.UNPUBLISHED)) {
```
jdamerow, line 62: If a space has null as status, it's considered published. This should only be the case for spaces created before we added the status field and all of them are considered published.

jdamerow (changes requested):

tusanand:
Simplified the conditions.

PUBLISHED spaces: Visible to all
UNPUBLISHED spaces:
	Authenticated user: Space visible with message
	Unauthenticated user: 404 error page visible

jdamerow:
Make it so, Jenkins.

jdamerow:
The alert should just say "This space has not been published."

tusanand:
Changed the message to "This space has not been published."

PR #258, opened 2021-09-29
Story/vspc 182

glend17:

https://diging.atlassian.net/browse/VSPC-182
  
Only published modules would show up on the public side, this is dependent on VSPC-141 and hasn't been implemented yet.
Existing modules without a status are treated as published. Therefore unless they are changed to UNPUBLISHED will behave the same as PUBLISHED modules.

### DIFF
--- a/vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionModuleController.java
+++ b/vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionModuleController.java
@@ -32,7 +32,7 @@ public class ExhibitionModuleController {
         }
         IModule module = moduleManager.getModule(id);
         model.addAttribute("module", module);
-        if (module == null || module.getModuleStatus() == null || module.getModuleStatus() == ModuleStatus.UNPUBLISHED) {
+        if (module == null || module.getModuleStatus() == ModuleStatus.UNPUBLISHED) {
             return "redirect:/exhibit/404";
         } else if (module.getStartSequence() == null) {
             model.addAttribute("showAlert", true);
--- a/vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionModuleController.java
+++ b/vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionModuleController.java
@@ -12,6 +12,7 @@ import edu.asu.diging.vspace.core.services.IModuleManager;
 import edu.asu.diging.vspace.core.services.ISpaceManager;
 import edu.asu.diging.vspace.web.exception.ModuleNotFoundException;
 import edu.asu.diging.vspace.web.exception.SpaceNotFoundException;
+import edu.asu.diging.vspace.core.model.impl.ModuleStatus;
 
 @Controller
 public class ExhibitionModuleController {
@@ -31,7 +32,7 @@ public class ExhibitionModuleController {
         }
         IModule module = moduleManager.getModule(id);
         model.addAttribute("module", module);
-        if (module == null) {
+        if (module == null || module.getModuleStatus() == null || module.getModuleStatus() == ModuleStatus.UNPUBLISHED) {
             return "redirect:/exhibit/404";
         } else if (module.getStartSequence() == null) {
             model.addAttribute("showAlert", true);
--- a/vspace/src/main/java/edu/asu/diging/vspace/web/staff/ModuleController.java
+++ b/vspace/src/main/java/edu/asu/diging/vspace/web/staff/ModuleController.java
@@ -6,6 +6,7 @@ import org.springframework.ui.Model;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.RequestMapping;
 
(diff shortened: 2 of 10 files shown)


pradeep-017:
As discussed Glen, Code wise and working wise things look good and I checked it in my system. Only concern is if you could align the status option below the Start sequence.

vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionModuleController.java
```diff
@@ -31,7 +32,7 @@ public String module(@PathVariable("id") String id, @PathVariable("spaceId") Str
         }
         IModule module = moduleManager.getModule(id);
         model.addAttribute("module", module);
-        if (module == null) {
+        if (module == null || module.getModuleStatus() == null || module.getModuleStatus() == ModuleStatus.UNPUBLISHED) {
```
jdamerow, line 35: modules without module status should be assumed to be published otherwise if we deploy this all existing modules won't show up anymore.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ModuleController.java
```diff
@@ -16,10 +17,11 @@
 
     @RequestMapping("/staff/module/{id}")
     public String showModule(@PathVariable String id, Model model) {
-
+        ModuleStatus status = moduleManager.getModule(id).getModuleStatus();
         model.addAttribute("module", moduleManager.getModule(id));
         model.addAttribute("slides", moduleManager.getModuleSlides(id));
         model.addAttribute("sequences", moduleManager.getModuleSequences(id));
+        model.addAttribute("moduleStatus", status);
```
jdamerow, line 24: get the module once and then use the module in line 21 and 24.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/UpdateModuleStatusController.java
```diff
@@ -0,0 +1,33 @@
...
+import edu.asu.diging.vspace.core.model.impl.ModuleStatus;
+import edu.asu.diging.vspace.core.services.IModuleManager;
+
+@Controller
+public class UpdateModuleStatusController {  
+    @Autowired
+    private IModuleManager moduleManager;
+    
+    @RequestMapping(value="/staff/module/{moduleId}/status", method=RequestMethod.POST)
+    public String updateStatus(HttpServletRequest request,RedirectAttributes attributes, @PathVariable("moduleId") String spaceId, @RequestParam("statusParam") ModuleStatus status) {
```
jdamerow, line 23: is the request object used anywhere?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/UpdateModuleStatusController.java
```diff
@@ -0,0 +1,33 @@
...
+        IModule module = moduleManager.getModule(spaceId);
+        module.setModuleStatus(status);
+        moduleManager.storeModule(module);
+        attributes.addAttribute("alertType", "success");
+        attributes.addAttribute("message", "Status successfully updated!");
+        attributes.addAttribute("showAlert", "true");
```
jdamerow, line 29: Does this work? doesn't it have to be `addFlashAttribute`?

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ModuleController.java
glend17: Done

vspace/src/main/java/edu/asu/diging/vspace/web/publicview/ExhibitionModuleController.java
glend17: Removed getModuleStatus from null check, treats it as the same as module being published

vspace/src/main/java/edu/asu/diging/vspace/web/staff/UpdateModuleStatusController.java
glend17: No, have removed it

vspace/src/main/java/edu/asu/diging/vspace/web/staff/UpdateModuleStatusController.java
glend17: Screen Shot 2022-01-20 at 1 00 34 PM

The status when updated appears like this

jdamerow:
Make it so, Jenkins.

jdamerow:
- The dropdowns should be aligned:
image
- The unpublished status should be indicated by an icon (like for spaces):
image
- on spaces staff page, the icon indicating that a link links an unpublished module should have a title that says so.

glend17:
> * The dropdowns should be aligned:
>   image
> * The unpublished status should be indicated by an icon (like for spaces):
>   image
> * on spaces staff page, the icon indicating that a link links an unpublished module should have a title that says so.

Fixed this now

jdamerow:
Make it so, Jenkins.

PR #259, opened 2021-10-01
Story/vspc 141

pradeep-017:

Search on Public site should bring in all possible space, module, slide and textblock results.
  
... describe here ...


vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionSearchDataAspect.java
```diff
@@ -0,0 +1,158 @@
...
+import edu.asu.diging.vspace.core.model.IdPrefix;
+import edu.asu.diging.vspace.core.model.impl.Exhibition;
+import edu.asu.diging.vspace.core.model.impl.SpaceStatus;
+import edu.asu.diging.vspace.core.services.IExhibitionManager;
+import edu.asu.diging.vspace.core.services.IModuleManager;
+import edu.asu.diging.vspace.core.services.ISpaceManager;
+
+@Component
+@Aspect
+public class ExhibitionSearchDataAspect {
```
jdamerow, line 31: this looks like a copy of `ExhibitionDataAspect`. What is this for? now two aspects are called for all methods in the general package doing almost the same?

vspace/src/main/java/edu/asu/diging/vspace/core/data/SequenceRepository.java
```diff
@@ -18,5 +19,7 @@
     
     @Query("SELECT d FROM Sequence d WHERE d.module.id = ?1 and d.id = ?2")
     public Sequence findSequenceForModuleAndSequence(String moduleId, String sequenceId);
-
+    
+    @Query(value="SELECT d.startSequence_id FROM Module d where d.startSequence_id IS NOT NULL", nativeQuery = true)
```
jdamerow, line 23: is there a reason this needs to be a native query?

vspace/src/main/java/edu/asu/diging/vspace/core/data/SlideRepository.java
```diff
@@ -25,5 +26,8 @@
 
     Page<ISlide> findDistinctByNameContainingOrDescriptionContaining(Pageable requestedPage, String name,
             String description);
+    
+    @Query(value="SELECT d.Slide_Id FROM Sequence_Slides d, Module m WHERE d.Sequence_Id = m.startSequence_id", nativeQuery = true)
```
jdamerow, line 30: see above

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/ModuleWithSpace.java
```diff
@@ -0,0 +1,15 @@
+package edu.asu.diging.vspace.core.model.impl;
```
jdamerow, line 1: this package contains our entities that are stored in the db, but this one is not if I'm not mistaken? So it shouldn't be in this package.

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/SlideWithSpace.java
```diff
@@ -0,0 +1,25 @@
+package edu.asu.diging.vspace.core.model.impl;
```
jdamerow, line 1: see above

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/StaffSearchSlide.java
```diff
@@ -0,0 +1,27 @@
+package edu.asu.diging.vspace.core.model.impl;
```
jdamerow, line 1: see above

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/StaffSearchSlideTextBlock.java
```diff
@@ -0,0 +1,37 @@
+package edu.asu.diging.vspace.core.model.impl;
```
jdamerow, line 1: see above

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/StaffSearchSpace.java
```diff
@@ -0,0 +1,16 @@
+package edu.asu.diging.vspace.core.model.impl;
+
+import java.util.List;
+
+public class StaffSearchSpace {
```
jdamerow, line 5: are the staff serach classes used anywhere?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/PublicSearchManager.java
```diff
@@ -0,0 +1,144 @@
...
+        int totalSlideTextPage = slidetextPage.getTotalPages();
+        if (page > totalSlideTextPage) {
+            totalSlideTextPage = totalSlideTextPage == 0 ? 1 : totalSlideTextPage;
+            requestedPageForSlideText = PageRequest.of(totalSlideTextPage - 1, pageSize);
+            slidetextPage = textContentBlockRepo.findWithNameOrDescription(requestedPageForSlideText, searchTerm);
+        }
+        return slidetextPage;
+    }
+
+}
```
jdamerow, line 144: this class has a lot of overlapping code with the staffsearchmanager, doesn't it? those two should be consolidated to make it DRY.

jdamerow (changes requested):
- are the serach result objects any different than the ones for the staff search? If not, just reuse the existing ones.

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/StaffSearchSpace.java
swetalina-prusty25: No, This class have not been used anywhere, There is a similar class present naming StaffSearchSpaceResult, which has same body as StaffSearchSpace , so that is being used every where.

vspace/src/main/java/edu/asu/diging/vspace/core/data/SequenceRepository.java
swetalina-prusty25: Changed the native query to jpql query.

PR #261, opened 2021-10-25
Story/vspc 190

glend17:

https://diging.atlassian.net/browse/VSPC-190
  
No


vspace/src/main/java/edu/asu/diging/vspace/core/model/SpacesCustomOrder.java
```diff
@@ -0,0 +1,67 @@
...
+    private String id;
+
+    @OneToOne(targetEntity = Space.class)
+    private ISpace space;
+
+    private Integer customOrder;
+    
+    public SpacesCustomOrder() {
+        
+    }
```
jdamerow, line 35: move to line 33

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ExhibitionManager.java
```diff
@@ -66,4 +67,13 @@ public IExhibition getStartExhibition() {
         }
         return null;
     }
+    
+    @Override
+    public IExhibition updateSpaceOrderMode(ExhibitionSpaceOrderMode mode) {
+        // for now we just take the first one created, there shouldn't be more than one
+        List<Exhibition> exhibitions = exhibitRepo.findAllByOrderByIdAsc();
+        Exhibition exhibition = exhibitions.get(0);
```
jdamerow, line 75: just call `getStartExhibition`? this will take care of any checks etc.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceManager.java
```diff
@@ -297,4 +303,16 @@ public void deleteSpaceById(String id) {
...
     }
+
+    /**
+     * Method to sort published spaces based on the given order, i.e Alphabetical,
+     * Creation Date, or, Custom
+     * @param publishedSpaces
+     */
+    @Override
+    public List<ISpace> sortPublishedSpacesByGivenOrder(List<ISpace> publishedSpaces) {
+        IExhibition exhibitition = exhibitionManager.findAll().get(0);
```
jdamerow, line 314: why not use `getStartExhibition`?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpacesCustomOrderManager.java
```diff
@@ -0,0 +1,145 @@
...
+
+@Service
+@Transactional(rollbackFor = { Exception.class })
+public class SpacesCustomOrderManager implements ISpacesCustomOrderManager {
+
+    @Autowired
+    SpacesCustomOrderRepository spacesCustomOrderRepository;
+    
+    @Autowired
+    ISpaceManager spaceManager;
```
jdamerow, line 35: class variables should be private

vspace/src/main/java/edu/asu/diging/vspace/core/model/SpacesCustomOrder.java
```diff
@@ -0,0 +1,67 @@
...
+
+import org.hibernate.annotations.GenericGenerator;
+import org.hibernate.annotations.Parameter;
+
+import edu.asu.diging.vspace.core.model.impl.Space;
+import edu.asu.diging.vspace.core.model.impl.VSpaceElement;
+
+@Entity
+@Inheritance(strategy = InheritanceType.JOINED)
+public class SpacesCustomOrder extends VSpaceElement implements ISpacesCustomOrder {
```
jdamerow, line 19: Is there a reason you create one object for each space instead of one object that has an order list of spaces? This seems unnecessarily complex and is not very flexible. What if we want to allow different orders depending on certain conditions.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpacesCustomOrderManager.java
```diff
@@ -0,0 +1,145 @@
...
+     */
+    @Override
+    public void updateCustomOrder(List<SpacesCustomOrder> spacesCustomOrderCurrentList) {
+
+        if (spacesCustomOrderCurrentList == null) {
+            return;
+        }
+        List<SpacesCustomOrder> spacesCustomOrderNewList = new ArrayList<>();
+        for (SpacesCustomOrder eachSpace : spacesCustomOrderCurrentList) {
+            logger.info("custom order is {}", eachSpace.getCustomOrder());
```
jdamerow, line 57: not useful, as we talked about yesterday ;)

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpacesCustomOrderManager.java
```diff
@@ -0,0 +1,145 @@
...
+        spacesCustomOrderRepository.saveAll(spacesCustomOrderNewList);
+    }
+    
+    /**
+     * This method persists all the published spaces into the SpacesCustomOrder table 
+     * This method is called every time we go to custom order tab as that will 
+     * ensure the table is populated with latest published spaces before changing the order
+     */    
+    @Override
+    public void persistPublishedSpacesToSpacesCustomOrder() {
```
jdamerow, line 76: this seems unnecessary; why not create objects as they are needed

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpacesCustomOrderManager.java
```diff
@@ -0,0 +1,145 @@
...
+        }
+        
+    }
+    
+    /**
+     * This method updates the custom order of existing spaces when a space is changed to unpublished
+     * @param space
+     */
+    @Override
+    public void updateStatusChangeToUnpublished(ISpace space) {
```
jdamerow, line 129: I don't think this is necessary; as long as when presented it is checked that a space is published, they can be part of the ordering even when they are unpublished.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ExhibitionSpaceOrderUtility.java
```diff
@@ -0,0 +1,71 @@
...
+import edu.asu.diging.vspace.core.model.impl.SpaceStatus;
+import edu.asu.diging.vspace.core.services.IExhibitionManager;
+import edu.asu.diging.vspace.core.services.ISpacesCustomOrderManager;
+import edu.asu.diging.vspace.core.services.impl.SpaceManager;
+
+@Component
+public class ExhibitionSpaceOrderUtility {
+    
+    @Autowired
+    ISpacesCustomOrderManager spacesCustomOrderManager;
```
jdamerow, line 24: should be private

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ExhibitionSpaceOrderUtility.java
```diff
@@ -0,0 +1,71 @@
...
+        Collections.sort(publishedSpaces, new Comparator<ISpace>() {
+            @Override
+            public int compare(ISpace space1, ISpace space2) {
+                return (space1.getCreationDate()).compareTo(space2.getCreationDate());
+            }
+        });
+        return publishedSpaces;
+    }
+    
+    public List<ISpace> customSpaceOrder(){
```
jdamerow, line 46: method names should be verbs

jdamerow (changes requested):
I'll stop with the review here as I think it'll change a lot if you change the entity to store the custom order.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ExhibitionManager.java
glend17: Done

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceManager.java
glend17: Done

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ExhibitionSpaceOrderUtility.java
glend17: Done

PR #267, opened 2021-12-01
Story/vspc 175

AvirupBiswas:

Right now we have a browser window wide option. We also want a fullscreen option that puts the browser into fullscreen.
  
... describe here ...


vspace/src/main/webapp/WEB-INF/views/exhibition/space.html
```diff
@@ -353,12 +393,25 @@
...
             externalLinks.forEach(function(link,index){
+            	console.log(link);
                 if(link!=null){
                     var posX = parseInt($("#space").css('margin-left')) + $("#space").position().left; 
     		        var posY = $("#space").position().top;
     		        var externalLink = $('<a></a>');
     		        externalLink.attr('href', link.externalLink.externalLink);
-    		        externalLink.attr('target', "_blank");
+    		        if(link.howToOpen == null)
+	    		        {
```
pradeep-017, line 403: indentation needs to be fixed.

vspace/src/main/webapp/WEB-INF/views/exhibition/space.html
```diff
@@ -353,12 +393,25 @@
                 }
             })
             externalLinks.forEach(function(link,index){
+            	console.log(link);
```
jdamerow, line 396: remove

jdamerow (changes requested):
Indentation still needs fixing in all the added js parts

jdamerow:
Make it so, Jenkins.

jdamerow:
- the icon to switch to full sceens needs a different icon than the image resize button and it needs a title
- once full screen mode is on it should stay in full screen mode until the user goes back to no full screen (right now if I click on a link it goes back to non full screen)
- the modules dropdown button (top right corner) disappears in full screen
- after going to full screen and back, the resize button doesn't work anymore
- if I first resize the image and then click on full screen the images does not become fully full screen
- in some cases the button to leave fullscreen mode stays visible after returning from fullscreen (could not figure out yet when that happens, I think it has something to do with the resize button)

vspace/src/main/webapp/WEB-INF/views/exhibition/space.html
```diff
@@ -353,12 +393,25 @@
...
+	    		        }
+    		        else
+    		        	{
+    		        	  if(link.howToOpen == "NEWWINDOW"){
+    		        		  externalLink.attr('target', "_blank");
+    		        		  }
+    		        	  else{
+    		        		  externalLink.attr('target', "_self"); 
+    		        	  }
+    		        	}
```
jdamerow, line 414: actually the above code is not part of this.story is it?

jdamerow:
Make it so, Jenkins.

jdamerow:
- The fullscreen button for modules need to be placed with he other buttons:
image
 -let's the `icon-visible` icon as icon for the fullscreen button.

jdamerow:
Make it so, Jenkins.

jdamerow:
these two buttons are a little misaligned:
image

jdamerow:
Same java 11 issue as with the other PRs. Once the required plugin is merged into develop, this build should pass.

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

jdamerow:
- fullscreen for modules still doesn't have a exit button in safari

jdamerow:
Make it so, Jenkins.

PR #286, opened 2022-04-04
Bug/vspc 207

swetalina-prusty25:


swetalina-prusty25:
finished the development of an image search field that allows users to search for images by name, file name, and description. It will filter the images.

vspace/src/main/java/edu/asu/diging/vspace/core/data/ImageRepository.java
```diff
@@ -17,6 +17,8 @@
 public interface ImageRepository extends PagingAndSortingRepository<VSImage, String> {
 
     List<VSImage> findByFilenameLikeOrNameLike(String filename, String name);
+    
+    List<VSImage> findByFilenameLikeOrNameLikeOrDescriptionLike(String filename, String name ,String Description);
```
jdamerow, line 21: variables/parameter names are lowercased

jdamerow: also, blank goes after not before the comma

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
```diff
@@ -232,6 +232,18 @@ public IVSImage getImageById(String imageId) throws ImageDoesNotExistException {
         results.forEach(r -> imageResults.add(r));
         return imageResults;
     }
+    
+    @Override
+    public List<IVSImage> findByFilenameOrNameContainsOrDescription(String searchTerm) {
```
jdamerow, line 237: just call this method `search`

jdamerow: also javadoc is missing

vspace/src/main/java/edu/asu/diging/vspace/web/staff/api/ImageSearchApiController.java
```diff
@@ -45,4 +58,32 @@
         }
         return new ResponseEntity<String>(idArray.toString(), HttpStatus.OK);
     }
+    
+    @RequestMapping("/staff/images/searchImagesncd")
```
jdamerow, line 62: let's make this `/staff/images/search/full`

vspace/src/main/java/edu/asu/diging/vspace/web/staff/api/ImageSearchApiController.java
```diff
@@ -45,4 +58,32 @@
...
+    public String searchImageDescription(@RequestParam(value = "searchText", required = false) String searchTerm , Model model, RedirectAttributes attributes) {
```
jdamerow, line 63: this method should be in its own controller

vspace/src/main/webapp/WEB-INF/views/staff/images/imagelist.html
```diff
@@ -199,6 +206,35 @@ <h2>Images</h2>
...
+                	 if($("#searchImageField").val().trim()) {
+                		 
+                          $('#searchImageButton').removeAttr('disabled');
+                	 }
+                	 else {
+                         $('#searchImageButton').attr('disabled','true');
+                	 }
+                	 if (!$('#searchImageButton').attr('disabled') && $("#searchImageField").val().trim()){
+                		 
+                			$("#searchImageLink").attr('href','[[@{/staff/images/searchImagesncd?searchText=}]]'+$("#searchImageField").val());
```
jdamerow, line 221: indentation

vspace/src/main/webapp/WEB-INF/views/staff/modules/sequences/sequence.html
```diff
@@ -1,348 +0,0 @@
-<html layout:decorate="~{layouts/main_staff}">
```
jdamerow, line 1: why is this file being deleted?

jdamerow (changes requested):

jdamerow:
The ticket description is not from the right ticket?

PR #298, opened 2022-05-09
Story/vspc 202 Upload files

glend17:

https://diging.atlassian.net/browse/VSPC-202

It should be possible to upload files
  
... describe here ...


glend17:
Closing PR to resolve conflicts

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/FileApiManager.java
```diff
@@ -0,0 +1,109 @@
...
+import edu.asu.diging.vspace.core.data.FileRepository;
+import edu.asu.diging.vspace.core.exception.FileStorageException;
+import edu.asu.diging.vspace.core.factory.IFileFactory;
+import edu.asu.diging.vspace.core.file.IStorageEngine;
+import edu.asu.diging.vspace.core.model.IVSFile;
+import edu.asu.diging.vspace.core.model.impl.VSFile;
+import edu.asu.diging.vspace.core.services.IFileApiManager;
+
+@Service
+public class FileApiManager implements IFileApiManager {
```
jdamerow, line 23: Why is this called "File**Api**Manager"? What does the API refer to? Unless there is a good reason, it should just be `FileManager`.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/api/FileApiController.java
```diff
@@ -0,0 +1,67 @@
...
+    public ResponseEntity<byte[]> downloadFile(Model model, @PathVariable String fileId) {
+        byte[] fileContent = null;
+        try {
+            fileContent = fileManager.downloadFile(fileId);
+        } catch (IOException e) {
+            return new ResponseEntity<byte[]>(fileContent, HttpStatus.INTERNAL_SERVER_ERROR);
+        }
+        
+        return new ResponseEntity<byte[]>(fileContent, HttpStatus.OK);
+    }
```
jdamerow, line 65: the three methods above are not about adding files and shoulld not be in the `AddFileApiController` (but in their own controllers).

vspace/src/main/java/edu/asu/diging/vspace/web/staff/api/FileApiController.java
```diff
@@ -0,0 +1,67 @@
+
+package edu.asu.diging.vspace.web.staff.api;
```
jdamerow, line 2: why are the file controllers in the `api` package? unless they are returning json, they should not be in here.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/api/FileApiController.java
```diff
@@ -0,0 +1,67 @@
...
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+
+import edu.asu.diging.vspace.core.model.IVSFile;
+import edu.asu.diging.vspace.core.model.impl.VSFile;
+import edu.asu.diging.vspace.core.services.impl.FileApiManager;
+import edu.asu.diging.vspace.web.staff.forms.FileForm;
+
+@Controller
+public class FileApiController {
```
jdamerow, line 25: same as with the manager, what does the `Api` part refer to?

vspace/src/main/webapp/WEB-INF/views/staff/files/add.html
```diff
@@ -0,0 +1,39 @@
...
+
+</script>
+</head>
+<body>
+<div class="main-content" layout:fragment="content">
+	<h2>Upload new File</h2>
+	
+	
+	<form method="POST" th:action="@{|/staff/files/add?${_csrf.parameterName}=${_csrf.token}|}" th:object="${files}" enctype="multipart/form-data" style="margin-top:40px">
+		<input type="hidden" th:name="${_csrf.parameterName}" th:value="${_csrf.token}">
```
jdamerow, line 18: the csrf token is already added to the url and doesn't need to be in the form.

vspace/src/main/webapp/WEB-INF/views/staff/files/file.html
```diff
@@ -0,0 +1,124 @@
...
+            $('.errorDelete').append(alert);
+        }
+    });
+}
+
+</script>
+<style>
+/* .download{
+margin-left:1200px;
+} */
```
jdamerow, line 75: remove if not needed

vspace/src/main/webapp/WEB-INF/views/staff/files/file.html
```diff
@@ -0,0 +1,124 @@
...
+ 			<div>
+	    		<h2 th:text=${file.filename}/>
+		        <p>[[${file.fileDescription}]]</p>
+		        
+		        
+	    	</div>
+ 			 
+ 		</div>
+ 		<div>
+ 		<button class="download btn primary-btn" type="submit" value="submit" onClick = "deleteFile()" style="float:right">
```
jdamerow, line 92: why not just linking to the file download url instead of using javascript?

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/FileApiManagerTest.java
```diff
@@ -0,0 +1,98 @@
...
+        String relativePath = "relativePath";
+        byte[] fileBytes = fileContentString.getBytes();
+        when(storageEngine.deleteFile(file)).thenReturn(true);
+        when(fileFactory.createFile(fileName, fileId)).thenReturn(file);
+        when(storageEngine.storeFile(fileBytes, fileName, fileContentString)).thenReturn(relativePath);
+        CreationReturnValue returnValue = serviceToTest.storeFile(fileBytes, fileName, fileDescription, fileId);
+        verify(fileFactory).createFile(fileName, "text/plain");
+    }
+    
+
```
jdamerow, line 96: what about test cases for not successful scenarios?

jdamerow (changes requested):

vspace/src/main/webapp/WEB-INF/views/staff/files/file.html
pkharge: Didn't get you, javascript is processing the response in success scenario for both download and delete.

jdamerow: ah, now I get it. I just saw the css class called "download" and assumed it was the download link. Should. probably be renamed ;)

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/core/aspects/ExhibitionDataAspect.java
```diff
@@ -119,7 +119,7 @@ private Object redirectRequest(ProceedingJoinPoint jp, String spaceId, String mo
         }
         // If user is not logged in and exhibition is not active, show maintenance page.
         if(authFacade.getAuthenticatedUser()==null && !exhibitionMode.equals(ExhibitionModes.ACTIVE)) {
-            return "maintenance";
+            return "exhibition/maintenance";
```
pkharge, line 122: Noticed that the path to the maintenance file was incorrect. Should I let this fix remain as part of this PR?

jdamerow:
Resolve conflicts please

vspace/src/main/java/edu/asu/diging/vspace/core/file/impl/StorageEngine.java
```diff
@@ -8,85 +8,126 @@
...
-		File fileObject = new File(path + File.separator + directory + File.separator + filename);
+    /**
+    * Method to rename image   
+    * 
+    * @param image - image file
+    * @param newFileName - new name of the file
+    * @return true if file renaming was successful, otherwise return false 
+    */ 
+    @Override
+    public boolean renameImage(IVSImage image, String newFileName) {
```
jdamerow, line 77: `IVSImage` and `IVSFile` should have a common super interface (or maybe `IVSImage` should be a subtype of `IVSFile`?). Then you need only one method here.

jdamerow: and actually, this class should just get a string anyways (it shouldn't need to know about our domain classes). All it should know about are how to get stuff from the file system and store stuff, no matter where the file to be stored comes from (image or file).

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/VSFile.java
```diff
@@ -0,0 +1,90 @@
...
+
+    @Id
+    @GeneratedValue(generator = "file_id_generator")
+    @GenericGenerator(name = "file_id_generator", parameters = @Parameter(name = "prefix", value = "FILE"), strategy = "edu.asu.diging.vspace.core.data.IdGenerator")
+    private String id;
+
+    @Lob
+    private String filename;
+    @Lob
+    private String parentPath;
```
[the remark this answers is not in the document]
jdamerow: ?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/FileManager.java
```diff
@@ -0,0 +1,108 @@
...
+    private IStorageEngine storageEngine;
+    
+    @Autowired
+    private IFileFactory fileFactory;
+    
+    @Autowired
+    private FileRepository fileRepo;
+    
+    @Value("${file_uploads_directory}")
+    private String fileUploadDir;
```
jdamerow, line 38: that should not be needed here. either the storage engine is the one deciding where to store something,  or this class but not both.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DownloadFileController.java
```diff
@@ -0,0 +1,45 @@
...
+            file = fileManager.getFileById(fileId);
+            resource = fileManager.downloadFile(file.getFilename());                  
+        } catch (IOException e) {
+            return new ResponseEntity<Resource>(resource, HttpStatus.INTERNAL_SERVER_ERROR);
+        }
+
+        return ResponseEntity.ok()
+                .header(HttpHeaders.CONTENT_DISPOSITION, "attachment; filename=" + file.getFilename())
+                .contentLength(resource.contentLength())
+                .contentType(MediaType.APPLICATION_OCTET_STREAM)
```
jdamerow, line 42: this will always make the browser download the file. the content-header should be the correct of the file.

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DownloadFileController.java
pkharge: @jdamerow, Do you mean the content type should be the actual file type like pdf / text etc?

pkharge:
@jdamerow , parentPath basically holds the name of the folder/directory it is stored in.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DownloadFileController.java
jdamerow: it should be the content type of the file (e.g. text/plain, application/pdf, etc).

pkharge:
@jdamerow I have a question here, for all the images I see a unique folder being created for each image, with the image id as folder name. All the files are going under a single folder configured in properties. 
Do we not want to follow the same behaviour for files as it is for images?

jdamerow:
> @jdamerow I have a question here, for all the images I see a unique folder being created for each image, with the image id as folder name. All the files are going under a single folder configured in properties. Do we not want to follow the same behaviour for files as it is for images?

Probably not a bad idea. I haven't looked too closely at how files are currently stored,  but you have to avoid overriding a file when a new file with the same filename is uploaded. Having a folder with the file id to store the file in solves that issue.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddFileController.java
```diff
@@ -0,0 +1,57 @@
...
+        byte[] fileBytes = null;
+        String originalFileName = null;
+        CreationReturnValue returnVal = null;
+        if (file != null) {
+            try {
+                fileBytes = file.getBytes();
+                originalFileName = file.getOriginalFilename();
+                returnVal = fileManager.storeFile(fileBytes, originalFileName, fileForm.getFileName(),fileForm.getDescription());
+            } catch (IOException e) {
+                logger.error("Error occured while creating file", e);
```
jdamerow, line 42: there needs to be an error message shown to the user.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DeleteFileController.java
```diff
@@ -0,0 +1,28 @@
...
+@Controller
+public class DeleteFileController {
+    
+    @Autowired
+    private FileManager fileManager;
+    
+    @RequestMapping(value = "/staff/files/delete/{fileId}", method = RequestMethod.POST)
+    public ResponseEntity<String> deleteFile(Model model, @PathVariable String fileId) {
+        if(!fileManager.deleteFile(fileId)) {
+            return new ResponseEntity<String>("Could not delete file", HttpStatus.BAD_REQUEST);
```
jdamerow, line 23: unless the request was actually a bad request (if the sender can do something differently), it shouldn't return a bad request. In this case a 500 is probably in order as it is unclear why this didn't work.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DownloadFileController.java
```diff
@@ -0,0 +1,45 @@
...
+    
+    
+    @RequestMapping(value = "/staff/files/download/{fileId}", method = RequestMethod.GET )
+    public ResponseEntity<Resource> downloadFile(Model model, @PathVariable String fileId) throws IOException {
+        Resource resource = null;  
+        IVSFile file= null;
+        try {        
+            file = fileManager.getFileById(fileId);
+            resource = fileManager.downloadFile(file.getFilename());                  
+        } catch (IOException e) {
```
jdamerow, line 35: exception needs to be logged.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditFileController.java
```diff
@@ -0,0 +1,30 @@
...
+    @Autowired
+    private FileManager fileManager;
+
+    @RequestMapping(value = "/staff/files/edit/{fileId}", method = RequestMethod.POST)
+    public String editFile(Model model, @PathVariable String fileId, @ModelAttribute FileForm fileForm) {
+        String fileName = fileForm.getFileName();
+        String description = fileForm.getDescription();
+        IVSFile file = fileManager.editFile(fileId, fileName, description);
+        String id = file.getId();
+        return "redirect:/staff/files/"+id;
```
jdamerow, line 27: merge with line 26

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DeleteFileController.java
pkharge: @jdamerow , should we send Internal server error instead?

jdamerow:
There are code factor issues.

jdamerow:
Make it so, Jenkins.

pkharge:
Question: if a file is created but nothing is uploaded, do we want to provide an option to later upload it?

jdamerow:
I think for now I would just make it a requirement that there needs to be a file selected when create a new file. We can add more functionality later to upload new versions, etc.

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/VSFile.java
```diff
@@ -0,0 +1,58 @@
...
+import javax.persistence.Lob;
+import javax.persistence.MappedSuperclass;
+
+import org.hibernate.annotations.GenericGenerator;
+import org.hibernate.annotations.Parameter;
+
+import edu.asu.diging.vspace.core.model.IVSFile;
+
+@Entity
+
```
jdamerow, line 16: remove empty line

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/FileManager.java
```diff
@@ -0,0 +1,110 @@
...
+    @Autowired
+    private IStorageEngine storageEngine;
+    
+    @Autowired
+    private IFileFactory fileFactory;
+    
+    @Autowired
+    private FileRepository fileRepo; 
+    
+    private final Logger logger = LoggerFactory.getLogger(getClass());
```
jdamerow, line 39: should  be declared first

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DeleteFileController.java
```diff
@@ -0,0 +1,28 @@
...
+
+import edu.asu.diging.vspace.core.services.impl.FileManager;
+
+@Controller
+public class DeleteFileController {
+    
+    @Autowired
+    private FileManager fileManager;
+    
+    @RequestMapping(value = "/staff/files/delete/{fileId}", method = RequestMethod.POST)
```
jdamerow, line 20: all  the  urls should  be `.../files/{fileId}/<action>` e.g. `.../files/{fileId}/delete`

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ListFileController.java
```diff
@@ -0,0 +1,49 @@
...
+
+@Controller
+public class ListFileController {
+    
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+    
+    @Autowired
+    private FileManager fileManager;
+    
+    @RequestMapping(value = "/staff/files/{id}", method = RequestMethod.GET)
```
jdamerow, line 32: this is for showing a file,  not listing it, right? Should be in its own controller.

vspace/src/main/java/edu/asu/diging/vspace/core/file/impl/StorageEngine.java
```diff
@@ -8,85 +8,115 @@
...
+
+    @Override
+    public Resource downloadFile(String fileName, String directory) throws IOException {               
+        File fileObject = getFile(directory, fileName);
+        Path path = Paths.get(fileObject.getAbsolutePath());
+        return new ByteArrayResource(Files.readAllBytes(path));   
+    }
+
+    @Override
+    public byte[] getFileContent(File fileObject) throws IOException  {
```
jdamerow, line 77: is this used anywhere outside of this class? if not, it should be private.

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
- file type should  be listed in file list
- there should  be a delete icon in each row
- view button should be an  icon (and should be edit)
- files list needs to be paginated

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ListFileController.java
```diff
@@ -0,0 +1,37 @@
...
+@Controller
+public class ListFileController {
+        
+    @Autowired
+    private FileManager fileManager;
+    
+    @RequestMapping(value = "/staff/files/list", method = RequestMethod.GET)
+    public String getFilesList(Model model, @RequestParam(value = "filesPagenum", required = false, defaultValue = "1") String filesPagenum) {
+               
+        Page<VSFile> filesPage = fileManager.getAllFiles(Integer.parseInt(filesPagenum));
```
jdamerow, line 26: if `filesPagenum` isn't a number this will throw an exception

vspace/src/main/java/edu/asu/diging/vspace/web/staff/FileController.java
```diff
@@ -0,0 +1,26 @@
...
+import edu.asu.diging.vspace.core.model.IVSFile;
+import edu.asu.diging.vspace.core.services.impl.FileManager;
+
+@Controller
+public class FileController {
+
+    @Autowired
+    private FileManager fileManager;
+
+    @RequestMapping(value = "/staff/files/{id}", method = RequestMethod.GET)
```
jdamerow, line 20: doesn't this page show the edit form? I would make the url `.../edit` and then move this method into the edit controller.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/FileManagerTest.java
```diff
@@ -0,0 +1,143 @@
...
+    public void init() {
+        MockitoAnnotations.initMocks(this);
+    }
+    
+    @Test
+    public void test_getFileById_success() {
+        VSFile file = new VSFile();
+        when(fileRepo.findById(fileId)).thenReturn(Optional.of(file));
+        serviceToTest.getFileById(fileId);
+        verify(fileRepo).findById(fileId);
```
jdamerow, line 61: we know the method behaves correctly if the returned file has the given id. it really doesn't matter if the repo is called or not. if we exchange the backend or a backend method this test should still pass even if the repo is not called as long as the returned result is correct.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/FileManagerTest.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+    
+    @Test
+    public void test_getAllFiles_success() {
+        serviceToTest.getAllFiles();
```
jdamerow, line 66: does this still compile? doesn't it expect a page number?

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/FileManagerTest.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+    
+    @Test
+    public void test_editFile_success() {
+        VSFile file = new VSFile();
+        file.setFilename(fileName);
+        when(fileRepo.findById(fileId)).thenReturn(Optional.of(file));
+        when(storageEngine.renameFile(file.getFilename(), fileName, null)).thenReturn(true);
+        IVSFile returnedFile = serviceToTest.editFile(fileId, fileName, fileDescription);
+        assertEquals(file.getFilename(), returnedFile.getFilename());
```
jdamerow, line 77: what's being edited here? you need to check that the edited values are correct.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/FileManagerTest.java
```diff
@@ -0,0 +1,143 @@
...
+    public void test_storeFile_success() throws FileStorageException {
+        String fileId = "fileId";
+        VSFile file = new VSFile();
+        String relativePath = "relativePath";
+        byte[] fileBytes = fileContentString.getBytes();
+        when(storageEngine.deleteFile(file.getFilename(), null)).thenReturn(true);
+        when(fileFactory.createFile(fileName, fileId)).thenReturn(file);
+        when(storageEngine.storeFile(fileBytes, fileName, fileContentString)).thenReturn(relativePath);
+        CreationReturnValue returnValue = serviceToTest.storeFile(fileBytes, fileName, fileDescription, fileId);
+        verify(fileFactory).createFile(fileName, "text/plain");
```
jdamerow, line 100: again, the returned value needs to be checked.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/FileManagerTest.java
```diff
@@ -0,0 +1,143 @@
...
+        String fileId = "fileId";
+        VSFile file = new VSFile();
+        byte[] fileBytes = fileContentString.getBytes();
+        when(storageEngine.deleteFile(file.getFilename(), null)).thenReturn(true);
+        when(fileFactory.createFile(fileName, "text/plain")).thenReturn(file);
+        when(storageEngine.storeFile(fileBytes, fileName, null)).thenThrow(new FileStorageException());
+        CreationReturnValue returnValue = serviceToTest.storeFile(fileBytes, fileName, fileDescription, fileId);
+
+        Assert.assertTrue(returnValue.getErrorMsgs().get(0).contains("File could not be stored: "));
+        verify(fileFactory).createFile(fileName, "text/plain");
```
jdamerow, line 115: not really important

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/FileManagerTest.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+    
+    @Test
+    public void test_editFile_failure() {
+        VSFile file = new VSFile();
+        file.setFilename(fileName);
+        when(fileRepo.findById(fileId)).thenReturn(Optional.of(file));
+        when(storageEngine.renameFile(file.getFilename(), fileName, null)).thenReturn(true);
+        IVSFile returnedFile = serviceToTest.editFile(fileId, fileName, fileDescription);
+        assertEquals(file.getFilename(), returnedFile.getFilename());
```
jdamerow, line 133: the passed filename is the same as the original one, isn't it? so this is true even if the method doesn't fail?

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/FileManagerTest.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+    @Test
+    public void test_downloadFile_failure() throws IOException {
+        String fileId = "fileId";
+        VSFile file = new VSFile();
+        when(fileRepo.findById(fileId)).thenReturn(Optional.of(file));
+        when(storageEngine.downloadFile(Mockito.any(String.class) ,Mockito.any( String.class))).thenThrow(new IOException());
+        assertThrows(IOException.class, () -> serviceToTest.downloadFile(file.getFilename(), fileId));
+    }
```
jdamerow, line 142: test methods should be sorted by tested methods (all editFile tests, then all downloadFile tests, etc)

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ListFileController.java
pkharge: Since we are setting a default value to "1", will there be a case when it won't be number?

jdamerow: since it's a request parameter you can't control what's being send to the server. Could be anything. default value is only used when no parameter is sent.

pkharge: okay yes. got it.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/FileManager.java
```diff
@@ -0,0 +1,122 @@
...
+        }
+        Pageable requestedPageForFiles = PageRequest.of(filesPagenum - 1, pageSize);
+
+        return fileRepo.findAll(requestedPageForFiles);
+    }
+    
+    @Override
+    public IVSFile getFileById(String id) {
+        Optional<VSFile> optional = fileRepo.findById(id);
+        return optional.get();
```
jdamerow, line 89: this will throw an exception I believe if optional is empty

vspace/src/main/java/edu/asu/diging/vspace/web/staff/EditFileController.java
```diff
@@ -0,0 +1,34 @@
...
+    @Autowired
+    private FileManager fileManager;
+
+    @RequestMapping(value = "/staff/files/{fileId}/update", method = RequestMethod.POST)
+    public String updateFile(Model model, @PathVariable String fileId, @ModelAttribute FileForm fileForm) {
+        IVSFile file = fileManager.editFile(fileId, fileForm.getFileName(), fileForm.getDescription());
+        return "redirect:/staff/files/"+file.getId();
+    }
+    
+    @RequestMapping(value = "/staff/files/{id}/edit", method = RequestMethod.GET)
```
jdamerow, line 27: the url should be the same for get and post

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ListFileController.java
```diff
@@ -0,0 +1,46 @@
...
+    
+    @RequestMapping(value = "/staff/files/list", method = RequestMethod.GET)
+    public String getFilesList(Model model, @RequestParam(value = "filesPagenum", required = false, defaultValue = "1") String filesPagenum) {
+        Integer pageNum = 1;
+        try {           
+            pageNum =  Integer.parseInt(filesPagenum);
+        } catch(NumberFormatException e) {
+            logger.error("Invalid page number", e);
+        }        
+        Page<VSFile> filesPage = fileManager.getAllFiles(pageNum);
```
jdamerow, line 35: the manager should return the interface not the implementation

vspace/src/main/java/edu/asu/diging/vspace/web/staff/forms/FileForm.java
```diff
@@ -0,0 +1,23 @@
+package edu.asu.diging.vspace.web.staff.forms;
+
+public class FileForm {
+
+    String fileName;
+    String description;
```
jdamerow, line 6: should be private

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/FileManagerTest.java
```diff
@@ -0,0 +1,170 @@
...
+    
+    @Test
+    public void test_getFileById_success() {
+        VSFile file = new VSFile();
+        file.setId(fileId);
+        when(fileRepo.findById(fileId)).thenReturn(Optional.of(file));
+        IVSFile fileResponse = serviceToTest.getFileById(fileId);
+        assertEquals(fileResponse.getId(), "FILE_ID");
+        
+    }
```
jdamerow, line 76: test case for passing an non-existing id is missing

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/FileManagerTest.java
```diff
@@ -0,0 +1,170 @@
...
+    
+    @Test
+    public void test_editFile_success() {
+        VSFile file = new VSFile();
+        file.setFilename(fileName);
+        file.setId(fileId);
+        when(fileRepo.findById(fileId)).thenReturn(Optional.of(file));
+        when(storageEngine.renameFile(file.getFilename(), "NEW_FILENAME", fileId)).thenReturn(true);
+        IVSFile returnedFile = serviceToTest.editFile(fileId, "NEW_FILENAME", fileDescription);                
+        assertEquals("NEW_FILENAME", returnedFile.getFilename());
```
jdamerow, line 102: file description needs to be checked

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/FileManagerTest.java
```diff
@@ -0,0 +1,170 @@
...
+   
+    @Test
+    public void test_downloadFile_failure() throws IOException {
+        String fileId = "fileId";
+        VSFile file = new VSFile();
+        when(fileRepo.findById(fileId)).thenReturn(Optional.of(file));
+        when(storageEngine.downloadFile(Mockito.any(String.class) ,Mockito.any( String.class))).thenThrow(new IOException());
+        assertThrows(IOException.class, () -> serviceToTest.downloadFile(file.getFilename(), fileId));
+    }
+}
```
jdamerow, line 170: what about successful `downloadFile`?

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
- after saving a file, it shows a 404 (the redirect is wrong)
- when moving the mouse over the delete icon, the mouse cursor needs to change like when moving the mouse over a link
- before actually deleting a file, there needs to be a warning modal telling the user that deleting will be permanent and ask if they want to proceed
- there need to be hover texts over the icons in the file list, telling the user what an icon does
- there needs to be a download icon in the file list, that directly downloads a file

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DownloadFileController.java
```diff
@@ -0,0 +1,52 @@
...
+    @RequestMapping(value = "/staff/files/{fileId}/download", method = RequestMethod.GET )
+    public ResponseEntity<Resource> downloadFile(Model model, @PathVariable String fileId) throws IOException {
+        Resource resource = null;  
+        IVSFile file= null;
+        try {        
+            file = fileManager.getFileById(fileId);
+            if(file!=null) {
+                resource = fileManager.downloadFile(file.getFilename(), fileId);   
+            } else {
+                logger.error("Could not find file " + fileId);
```
jdamerow, line 39: if the file doesn't exist, the controller will still return an ok, wouldn't it? It should return a 404.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/FileManagerTest.java
```diff
@@ -0,0 +1,179 @@
...
+        VSFile file = new VSFile();
+        file.setId(fileId);
+        files.add(file);
+        Pageable requestedPageForFiles = PageRequest.of(0, 10);
+        when(fileRepo.findAll(requestedPageForFiles)).thenReturn( new PageImpl<VSFile>(files));
+        Page<IVSFile> filesResponse = serviceToTest.getAllFiles(1);
+        assertEquals(filesResponse.getNumberOfElements(), 1);
+        assertEquals(filesResponse.getContent().get(0).getId(), fileId);
+
+    }
```
jdamerow, line 100: there should be a test for not finding any results.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/FileManagerTest.java
jdamerow: ?

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

jdamerow:
- I go to edit a file
- I changed the title
- I click save
- the file is saved but there is no indication it is (no success message, it's not returning to the list)
-> it needs to be clear the file has been saved

- then when I click on "Back", I get an error message or nothing happens (I think the error message appears when I hit "save" twice)
- also, after saving the first time, the csrf token is shown in the address bar, which it probably shouldn't

pkharge:
@jdamerow , 
After adding a redirect for a successful edit file, I am unable to see the csrf token in address bar, and any error on clicking Save button. Back button also seems fine. please let me know if the issue still persists.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/FileManagerTest.java
```diff
@@ -0,0 +1,202 @@
...
+        files.add(file);
+        Pageable requestedPageForFiles = PageRequest.of(0, 10);
+        when(fileRepo.findAll(requestedPageForFiles)).thenReturn( new PageImpl<VSFile>(files));
+        Page<IVSFile> filesResponse = serviceToTest.getAllFiles(1);
+        assertEquals(filesResponse.getNumberOfElements(), 1);
+        assertEquals(filesResponse.getContent().get(0).getId(), fileId);
+    }
+    
+    @Test
+    public void test_getAllFiles_failure() {
```
jdamerow, line 107: that's not really a failure but no results

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/FileManagerTest.java
jdamerow: I think this creates issues. The storage engine returns `file`. Presumably, `editFile` returns that same file object. So `file` == `returnedFile`, which means this line would test if `file.getFilename() is the same as file.isFilename()`. I would test it against the actual string (or test if `NEW_FILENAME` is not the same as `returnedFile.getFilename()`. Does that make sense?

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
success message after saving a file is still missinig.

jdamerow:
Resolve conflicts please

jdamerow (changes requested):

vspace/src/main/webapp/WEB-INF/views/staff/files/filelist.html
```diff
@@ -0,0 +1,206 @@
...
+	}).on("page", function(event, num){
+		console.log(num);
+		window.location.assign("[(@{'/staff/files/list'})]"+'?'+"filesPagenum="+num);
+	});
+	
+
+	const urlParams = new URLSearchParams(window.location.search);
+
+	if(urlParams.has('uploadFileSuccess') && urlParams.get('uploadFileSuccess') ) {
+		var alert = $('<div class="alert alert-success alert-dismissible fade show" role="alert"><p>File uploaded successfully.</p><button type="button" class="close" data-dismiss="alert" aria-label="Close"><span aria-hidden="true">&times;</span></button></div>');
```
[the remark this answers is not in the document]
pkharge: Hi @jdamerow , are you referring to the if condition here or the alert itself?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddFileController.java
```diff
@@ -0,0 +1,60 @@
...
+    
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+    
+    @Autowired
+    private FileManager fileManager;
+    
+    @RequestMapping(value = "/staff/files/add", method = RequestMethod.POST)
+    public ResponseEntity<String> createFile( Model model, @ModelAttribute FileForm fileForm, @RequestParam("file") MultipartFile file,
+            Principal principal) {
+        byte[] fileBytes = null;
```
jdamerow, line 35: define where needed

vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddFileController.java
```diff
@@ -0,0 +1,60 @@
...
+        String originalFileName = null;
```
jdamerow, line 36: define where needed

vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddFileController.java
```diff
@@ -0,0 +1,60 @@
...
+        CreationReturnValue returnVal = null;
+        if (file != null) {
+            try {
+                fileBytes = file.getBytes();
+                originalFileName = file.getOriginalFilename();
+                returnVal = fileManager.storeFile(fileBytes, originalFileName, fileForm.getFileName(),fileForm.getDescription());
```
jdamerow, line 42: is `returnVal` used anywhere?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DownloadFileController.java
```diff
@@ -0,0 +1,53 @@
...
+    @Autowired
+    private FileManager fileManager;
+    
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+    
+    
+    @RequestMapping(value = "/staff/files/{fileId}/download", method = RequestMethod.GET )
+    public ResponseEntity<Resource> downloadFile(Model model, @PathVariable String fileId) throws IOException {
+        Resource resource = null;  
+        IVSFile file= null;
```
jdamerow, line 33: define where needed

vspace/src/main/webapp/WEB-INF/views/staff/files/add.html
```diff
@@ -0,0 +1,78 @@
...
+				cache : false,
+				enctype : 'multipart/form-data',
+				contentType : false,
+				processData : false,
+				data : formData,
+				success : function(data) {
+					window.location.href = "[(@{'/staff/files/list/'})]"+'?'+"uploadFileSuccess=true&message="+data;
+				},
+				error : function(data) {
+					var alert = $('<div class="alert alert-danger alert-dismissible fade show" role="alert"><p>File upload failed. Please try again after sometime.</p><button type="button" class="close" data-dismiss="alert" aria-label="Close"><span aria-hidden="true">&times;</span></button></div>');
```
jdamerow, line 27: Replace "Please try again after sometime." with "Please try again later."

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
Deleting a file from the file details page, there should also be a confirmation modal like on the file list page.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DownloadFileController.java
pkharge: This object is used in the return statement. Hence, defined outside try block

vspace/src/main/java/edu/asu/diging/vspace/core/file/impl/StorageEngine.java
```diff
@@ -21,38 +26,81 @@
...
+        }
+        input.close();
+        output.flush();
+        output.close();
+
+        byteOutput.flush();
+        byte[] bytes = byteOutput.toByteArray();
+        byteOutput.close();
+        return bytes;
+    }
```
jdamerow, line 99: didn't notice until now, but this looks like almost the exact code as `getMediaContent`.

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
this still doesn't seem to be addressed:
> Deleting a file from the file details page, there should also be a confirmation modal like on the file list page.

vspace/src/main/java/edu/asu/diging/vspace/core/file/impl/StorageEngine.java
```diff
@@ -21,38 +26,56 @@
...
         return directory;
-	}
-	
-	@Override
-	public byte[] getMediaContent(String directory, String filename) throws IOException {
-		File fileObject = new File(path + File.separator + directory + File.separator + filename);
+    }
+
+    @Override
+    public byte[] getImageContent(String directory, String filename) throws IOException {
```
jdamerow, line 58: if this method just calls another method with the exact same parameters, you don't need two.

vspace/src/main/java/edu/asu/diging/vspace/core/file/impl/StorageEngine.java
```diff
@@ -21,38 +26,56 @@
...
+    private String path;
+
+
+    /* (non-Javadoc)
+    * @see edu.asu.diging.vspace.core.file.impl.IStorageEngine#storeFile(byte[], java.lang.String, java.lang.String)
+    */
+    @Override
+    public String storeFile(byte[] fileContent, String filename, String directory) throws FileStorageException {
+        
+
```
jdamerow, line 39: remove the two blank lines

vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddFileController.java
```diff
@@ -0,0 +1,57 @@
...
+                String originalFileName = file.getOriginalFilename();
+                fileManager.storeFile(fileBytes, originalFileName, fileForm.getFileName(),fileForm.getDescription());
+              
+            } catch (IOException e) {
+                logger.error("Error occured while creating file", e);  
+                return new ResponseEntity<String>("Could not create file", HttpStatus.INTERNAL_SERVER_ERROR);
+            }
+            
+        }      
+        return new ResponseEntity<String>("File uploaded successfully", HttpStatus.OK);
```
jdamerow, line 47: if there is no file to be uploaded, it should probably return a bad request or something like that.

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddFileController.java
jdamerow: there should be a period at the end of the sentence.

jdamerow:
Make it so, Jenkins.

PR #302, opened 2022-05-23
Story/vspc 213

pkharge:

In the exhibition configuration, users should be able to specify multiple languages for the exhibition. This ticket is just for the specification of languages, not providing text for multiple languages. So, all this needs to be doing is letting the user select languages from a list of languages. There needs to be a default language as well. So the user would select one of the languages as default language.

The follow up tickets will deal with adding slide and space texts in multiple languages. So the data model for this ticket should be a new Language or ExhibitionLanguage class which at the moment won’t have much additional info besides a label and a 2 character ISO code (e.g. en, de, it,…).
  
... describe here ...


vspace/src/main/java/edu/asu/diging/vspace/config/ExhibitionLanguageConfig.java
```diff
@@ -0,0 +1,58 @@
...
+
+@Component
+@PropertySource(value= "classpath:exhibitionLanguages.properties" , factory=JsonPropertySourceFactory.class)
+@Configuration
+public class ExhibitionLanguageConfig {
+
+    @Autowired
+    private Environment environment;
+    
+    List<Map>  exhibitionLanguageList = new ArrayList<Map>();
```
jdamerow, line 32: shoudl be private

vspace/src/main/java/edu/asu/diging/vspace/config/ExhibitionLanguageConfig.java
```diff
@@ -0,0 +1,58 @@
...
+    
+    /**
+     * Fetches the configured language list from environment property source and stores in exhibitionLanguageList
+     * 
+     */
+    @PostConstruct
+    public void init() {
+        for(Iterator it = ((AbstractEnvironment) environment).getPropertySources().iterator(); it.hasNext(); ) {
```
jdamerow, line 40: since only one is listed in the `PropertySource`, can there be more in the list than just this one?

vspace/src/main/java/edu/asu/diging/vspace/config/ExhibitionLanguageConfig.java
```diff
@@ -0,0 +1,58 @@
...
+            org.springframework.core.env.PropertySource propertySource = (org.springframework.core.env.PropertySource) it.next();
+            if (propertySource instanceof MapPropertySource) {
+                MapPropertySource mapSource = ((MapPropertySource) propertySource);
+                if("json-property".equals(mapSource.getName())) {
```
jdamerow, line 44: strings should never be hard-coded. If a string is needed here, it should be in a constant.

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/ExhibitionLanguage.java
```diff
@@ -0,0 +1,93 @@
...
+
+import org.hibernate.annotations.GenericGenerator;
+import org.hibernate.annotations.Parameter;
+
+@Entity
+public class ExhibitionLanguage {
+
+    public ExhibitionLanguage() {
+        super();
+    }
```
jdamerow, line 20: is this constructor necessary?

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/ExhibitionLanguage.java
```diff
@@ -0,0 +1,93 @@
...
+
+    @Id
+    @GeneratedValue(generator = "exhibit_language_id_generator")
+    @GenericGenerator(name = "exhibit_language_id_generator", parameters = @Parameter(name = "prefix", value = "EXH"), strategy = "edu.asu.diging.vspace.core.data.IdGenerator")
```
jdamerow, line 24: `EXH` is the prefix for exhibitiion objects. here it should be a prefix for language objects.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ExhibitionManager.java
```diff
@@ -20,6 +25,12 @@ public class ExhibitionManager implements IExhibitionManager {
 
     @Autowired
     private ExhibitionRepository exhibitRepo;
+    
+    @Autowired
+    private ExhibitionLanguageConfig exhibitionLanguageConfig;
+    
+    private final Logger logger = LoggerFactory.getLogger(getClass());
```
jdamerow, line 32: llogger decllarations should be first in the class

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ExhibitionManager.java
```diff
@@ -66,4 +77,50 @@ public IExhibition getStartExhibition() {
...
+        if(CollectionUtils.isNotEmpty(codes)) {
+            if(defaultLanguage!=null) {
+                codes.add(defaultLanguage);
+            }
+            codes.forEach(code -> {
+                Optional<ExhibitionLanguage> languageMap = exhibitionLanguageConfig.getExhibitionLanguageList()
+                        .stream().filter(map-> code.equalsIgnoreCase((String) map.get("code")))
+                        .map(language -> { 
+                            ExhibitionLanguage exhibitionLanguage =   new ExhibitionLanguage((String) language.get("label"),
+                                    (String) language.get("code"), exhibition);
```
jdamerow, line 100: same as before, hard-coded strings should be in constants

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ExhibitionManager.java
```diff
@@ -66,4 +77,50 @@ public IExhibition getStartExhibition() {
...
+     * @param languages
+     */
+    @Override
+    public void updateExhibitionLanguages(Exhibition exhibition, List<String> codes, String defaultLanguage) {
+        List<ExhibitionLanguage> languageMapList = new ArrayList();
+        if(CollectionUtils.isNotEmpty(codes)) {
+            if(defaultLanguage!=null) {
+                codes.add(defaultLanguage);
+            }
+            codes.forEach(code -> {
```
jdamerow, line 95: I think I would turn this around and iterate over exhibitionLanguageConfig.getExhibitionLanguageList(). You would only need one stream and filter for everything in the codes list. So only one iteration instead of multiple if I'm not mistaken.

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/config/ExhibitionLanguageConfig.java
pkharge: Didn't get you. There are multiple property sources. we are iterating through it to search for the particular property we need

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/ExhibitionLanguage.java
pkharge: Yeah. Actually there is a parameterized constructor added.

vspace/src/main/java/edu/asu/diging/vspace/config/ExhibitionLanguageConfig.java
jdamerow: Are there more than the one listed in the `@PropertySource` annoation?

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/ExhibitionLanguage.java
jdamerow: Ah I see, I didn't see it because constructors should be grouped together after the field declarations ;)

vspace/src/main/java/edu/asu/diging/vspace/config/ExhibitionLanguageConfig.java
pkharge: Yes, there are around 10 in the list.

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/ExhibitionLanguage.java
```diff
@@ -0,0 +1,93 @@
...
+               && Objects.equals(label, other.label);
+    }
+
+    public boolean isDefault() {
+        return isDefault;
+    }
+
+    public void setDefault(boolean isDefault) {
+        this.isDefault = isDefault;
+    }
```
jdamerow, line 91: getter and setters should be in order of variable declaration. hashcode/equals should be after getters/setters.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ExhibitionManager.java
```diff
@@ -66,4 +79,42 @@ public IExhibition getStartExhibition() {
...
+    @Override
+    public void updateExhibitionLanguages(Exhibition exhibition, List<String> codes, String defaultLanguage) {
+        if(CollectionUtils.isNotEmpty(codes)) {
+            if(defaultLanguage!=null) {
+                codes.add(defaultLanguage);
+            }
+            if(CollectionUtils.isNotEmpty(exhibitionLanguageConfig.getExhibitionLanguageList())) {
+                List<ExhibitionLanguage> exhibitionLanguages =   exhibitionLanguageConfig.getExhibitionLanguageList()
+                        .stream().filter(languageMap -> codes.contains(languageMap.get(ConfigConstants.CODE)))
+                        .map(map -> {
```
jdamerow, line 99: `map` (as variable name) is a language, isn't it? It should have a. better name.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ExhibitionManager.java
```diff
@@ -66,4 +79,42 @@ public IExhibition getStartExhibition() {
...
+     * Updates the Exhibition with given list of languages. It fetches the language from exhibitionLanguageConfig using code.
+     *  
+     * @param exhibition
+     * @param defaultLanguage 
+     * @param languages
+     */
+    @Override
+    public void updateExhibitionLanguages(Exhibition exhibition, List<String> codes, String defaultLanguage) {
+        if(CollectionUtils.isNotEmpty(codes)) {
+            if(defaultLanguage!=null) {
```
jdamerow, line 93: `&& !codes.contains(defaultLanguage)` or you might end up with the same code twice?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ExhibitionManager.java
```diff
@@ -66,4 +79,42 @@ public IExhibition getStartExhibition() {
...
+                                    (String) map.get(ConfigConstants.CODE), exhibition);
+
+                            if(exhibitionLanguage.getCode().equalsIgnoreCase(defaultLanguage)) {
+                                exhibitionLanguage.setDefault(true);
+                            }
+                            return exhibitionLanguage;
+                        }).collect(Collectors.toList());
+
+                if(CollectionUtils.isNotEmpty(exhibitionLanguages)) {
+                    logger.info("Updating Exhibition with Languages" + codes);
```
jdamerow, line 110: remove; it doesn't add any value. If at all, this should be trace.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ExhibitionManager.java
```diff
@@ -66,4 +79,42 @@ public IExhibition getStartExhibition() {
...
+                    exhibition.getLanguages().addAll(exhibitionLanguages);
+                }
+            } else {
+                logger.error("Language list configuration not available");
```
jdamerow, line 114: in this case we probably want to throw an exception since that would indicate a faulty installation.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ExhibitionConfigurationController.java
```diff
@@ -37,18 +44,29 @@ public class ExhibitionConfigurationController {
...
 
     @RequestMapping("/staff/exhibit/config")
     public String showExhibitions(Model model) {
         // for now we assume there is just one exhibition
         IExhibition exhibition = exhibitManager.getStartExhibition();
         if(exhibition!=null) {
             model.addAttribute("exhibition", exhibition);
+            if(exhibition.getLanguages() != null ) {
+                model.addAttribute("mappedLanguages", exhibition.getLanguages()
+                        .stream().map(language -> language.getLabel()).collect(Collectors.toList()));
```
jdamerow, line 61: what are "mappedLanguages"? and why not putting the language list in the model (not just the labels), you probably need them somewhere (or if not, it's just a `.label` more to show the language names).

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ExhibitionLanguageTest.java
```diff
@@ -0,0 +1,154 @@
...
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+
+import edu.asu.diging.vspace.config.ExhibitionLanguageConfig;
+import edu.asu.diging.vspace.core.data.ExhibitionRepository;
+import edu.asu.diging.vspace.core.model.impl.Exhibition;
+import edu.asu.diging.vspace.core.model.impl.ExhibitionLanguage;
+
+public class ExhibitionLanguageTest {
```
jdamerow, line 23: Tests class should be named `ClassToBeTestedNameTest`

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ExhibitionConfigurationController.java
pkharge: mappedLanguages is just used to highlight the already mapped languages to an exhibition, in the selection dropdown. This is to avoid adding any more logic in front end. 
The entire language object is being sent  to populate the dropdown in "languageList".

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ExhibitionManager.java
```diff
@@ -66,4 +81,79 @@ public IExhibition getStartExhibition() {
...
+    public void updateExhibitionLanguages(Exhibition exhibition, List<String> codes, String defaultLanguage) throws LanguageListConfigurationNotFound {
+        if(CollectionUtils.isNotEmpty(codes)) {
+            
+            // Adds defaultLanguage to codes list if not already exists.
+            if(defaultLanguage!=null && !codes.contains(defaultLanguage)) {
+                codes.add(defaultLanguage);
+            }
+
+
+            if(CollectionUtils.isNotEmpty(exhibitionLanguageConfig.getExhibitionLanguageList())) {
```
jdamerow, line 103: check this right at the beginning. no need to do anything if there is no exhibition language list.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ExhibitionManager.java
```diff
@@ -66,4 +81,79 @@ public IExhibition getStartExhibition() {
...
+     * @param exhibition
+     * @param languageMap
+     * @return
+     */
+    private ExhibitionLanguage addExhibitionLanguage(Exhibition exhibition, Map languageMap) {
+        ExhibitionLanguage exhibitionLanguage =   new ExhibitionLanguage((String) languageMap.get(ConfigConstants.LABEL),
+                (String) languageMap.get(ConfigConstants.CODE), exhibition);
+
+        int index =  exhibition.getLanguages().indexOf(exhibitionLanguage);
+        if( index < 0 ) {
```
jdamerow, line 132: why not use `exhibition.getLanguages().contains()`?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ExhibitionManager.java
```diff
@@ -66,4 +81,79 @@ public IExhibition getStartExhibition() {
...
+            exhibition.getLanguages().add(exhibitionLanguage);
+        } 
+        else {
```
jdamerow, line 135: please follow our style guide

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ExhibitionManager.java
```diff
@@ -66,4 +81,79 @@ public IExhibition getStartExhibition() {
...
+     * @param exhibitionLanguage
+     * @param defaultLanguage
+     */
+    private void updateDefault(ExhibitionLanguage exhibitionLanguage, String defaultLanguage) {
+       
+        if(exhibitionLanguage.getCode().equalsIgnoreCase(defaultLanguage)) {
+            exhibitionLanguage.setDefault(true);
+        } else {
+            exhibitionLanguage.setDefault(false);
+        }
```
jdamerow, line 155: I don't think you need an extra method for this. Just make it 
`exhibitionLanguage.setDefault(exhibitionLanguage.getCode().equalsIgnoreCase(defaultLanguage))`

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ExhibitionConfigurationController.java
jdamerow: ah, makes sense. It needs a better name though. It's not clear what "mappedLanguages" would be in this context.

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ExhibitionManager.java
pkharge: I would need the object to update isDefault to false in case a different default language is selected by user. Hence using index.

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/Exhibition.java
```diff
@@ -116,4 +129,31 @@ public void setAboutPageConfigured(boolean aboutPageConfigured) {
...
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj)
+            return true;
+        if (obj == null)
+            return false;
+        if (getClass() != obj.getClass())
+            return false;
+        Exhibition other = (Exhibition) obj;
+        return Objects.equals(id, other.id);
```
jdamerow, line 155: merge with line 154

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ExhibitionManager.java
```diff
@@ -66,4 +80,58 @@ public IExhibition getStartExhibition() {
...
+     *  
+     * @param exhibition
+     * @param defaultLanguage 
+     * @param languages
+     * @throws LanguageListConfigurationNotFound 
+     */
+    @Override
+    public void updateExhibitionLanguages(Exhibition exhibition, List<String> codes, String defaultLanguage) throws LanguageListConfigurationNotFound {
+        if(CollectionUtils.isEmpty(exhibitionLanguageConfig.getExhibitionLanguageList())) {
+            throw new LanguageListConfigurationNotFound("Exhibition Language Configuration not found");
```
jdamerow, line 95: I think I would make `LanguageListConfigurationNotFound` a RuntimeException. If this happens it means that the installation is incomplete or something else is going wrong and there is really no good way out. So I'd make it a runtime exception and not declare or handle the exception. Then just add a handler to the `ExhibitionExceptionHandler` that handles all runtime exceptions by showing a 500 page.

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
On the confiiguration page, let's make the default language dropdown show '----' or something if there is none selected yet.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ExhibitionManager.java
pkharge: I have added it to the exception handler. The exception is thrown and not being handled. Is there any thing else to be done for this?

vspace/src/main/java/edu/asu/diging/vspace/web/exception/ExhibitionExceptionHandler.java
```diff
@@ -90,5 +92,19 @@ protected ModelAndView handleSpaceNotFoundException(HttpServletRequest request,
...
+    
+    @ExceptionHandler({ LanguageListConfigurationNotFound.class })
+    protected ModelAndView handleLanguageListConfigurationNotFoundException(HttpServletRequest request,
+            LanguageListConfigurationNotFound ex) {
+        ModelAndView modelAndView = new ModelAndView();
+        modelAndView.addObject("error_code", language_list_configuration_not_found);
+        modelAndView.addObject("showAlert", true);
+        modelAndView.addObject("message", ex.getMessage());
+        logger.info("LanguageListConfigurationNotFound Occured:: URL=" + request.getRequestURL());
+        logger.info("Code:: "+language_list_configuration_not_found+" Message:: " + ex.getMessage());
```
jdamerow, line 104: you want to make sure to log the stacktrace

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ExhibitionManager.java
```diff
@@ -75,4 +89,58 @@ public IExhibition getStartExhibition() {
...
+    /**
+     * Updates the Exhibition with given list of languages. It fetches the language from exhibitionLanguageConfig using code.
+     *  
+     * @param exhibition
+     * @param defaultLanguage 
+     * @param languages
+     * @throws LanguageListConfigurationNotFound 
+     */
+    @Override
+    public void updateExhibitionLanguages(Exhibition exhibition, List<String> codes, String defaultLanguage) throws LanguageListConfigurationNotFound {
```
jdamerow, line 102: the exception should be a runtime exception and wouldn't need to be declared.

vspace/src/main/java/edu/asu/diging/vspace/core/exception/LanguageListConfigurationNotFound.java
```diff
@@ -0,0 +1,27 @@
+package edu.asu.diging.vspace.core.exception;
+
+public class LanguageListConfigurationNotFound  extends Exception {
```
jdamerow, line 3: this should be a runtime exception and exception class names should end in `Exception`

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
Besides that, looks good.

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ExhibitionConfigurationController.java
```diff
@@ -59,13 +82,16 @@ public String showExhibitions(Model model) {
      * @param spaceParam
      * @param attributes
      * @return
+     * @throws LanguageListConfigurationNotFoundException
```
jdamerow, line 85: method is not throwing this anymore

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
After that should be good.

jdamerow:
Make it so, Jenkins.

jdamerow:
There seems to be a bug. I have 4 languages currently selected, I've deselected 2 and saved. But the four are still selected when the page reloads.

PR #306, opened 2022-06-17
Story/vspc 141 rebase

pkharge:

The search should search all spaces (title, description), all modules (title, description), and all slides (title, text blocks). the user should be able to go directly to a found element.
  
This is a PR to correct the previous git merge errors in the branch.


vspace/src/main/webapp/WEB-INF/views/staff/modules/module.html
```diff
@@ -285,22 +285,27 @@ <h2>
...
+							th:selected="${#strings.equals(moduleStatus,'UNPUBLISHED')}">Unpublished</option>
+					</select>
+					<p style="display: inline; padding-left: 10px;">
+						<input class="btn primary-btn btn-sm" type="submit" value="Submit" />
+					</p>
+				</form>
+			</div>
+		</div>
+		<div id="sequences" class="tab sequences" style="margin-top: 0;">
     		<div style="display: flex; justify-content: space-between">
```
pkharge, line 309: only a formatting change

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/PublicSearchManager.java
```diff
@@ -0,0 +1,65 @@
...
+        Page<ISpace> spacePage = spaceManager.findBySpaceStatusAndNameOrDescription(requestedPageForSpace, SpaceStatus.PUBLISHED, searchTerm);
+        int totalSpacePage = spacePage.getTotalPages();
+        /* if page>total pages,last page is returned */
+        if (page > totalSpacePage) {
+            totalSpacePage = totalSpacePage == 0 ? 1 : totalSpacePage;
+            requestedPageForSpace = PageRequest.of(totalSpacePage - 1, pageSize);
+            spacePage = spaceManager.findBySpaceStatusAndNameOrDescription(requestedPageForSpace, SpaceStatus.PUBLISHED, searchTerm);
+        }
+        return spacePage;
+    }
```
jdamerow, line 64: This class is pretty much the same as the staff search manager. They should be combined. It's probably a good use case for the template method pattern.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SpaceManager.java
```diff
@@ -297,4 +297,9 @@ public Iterable<Space> addIncomingLinkInfoToSpaces(Iterable<Space> spaces) {
     public Page<ISpace> findByNameOrDescription(Pageable requestedPage, String searchText) {
         return spaceRepo.findDistinctByNameContainingOrDescriptionContaining(requestedPage, searchText,searchText);
     }
+    
+    @Override
+    public Page<ISpace> findBySpaceStatusAndNameOrDescription(Pageable requestedPage, SpaceStatus spaceStatus, String searchTerm) {
+        return spaceRepo.findDistinctBySpaceStatusAndNameContainingOrDescriptionContaining(requestedPage, spaceStatus, searchTerm, searchTerm);
+    }
```
jdamerow, line 304: this and the previous method should be combined.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/model/ModuleWithSpace.java
```diff
@@ -0,0 +1,16 @@
+package edu.asu.diging.vspace.core.services.impl.model;
+import edu.asu.diging.vspace.core.model.impl.Module;
+
+public class ModuleWithSpace extends Module {
```
jdamerow, line 4: is this any different from the one used on the staff search?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/model/SlideWithSpace.java
```diff
@@ -0,0 +1,25 @@
+package edu.asu.diging.vspace.core.services.impl.model;
+import edu.asu.diging.vspace.core.model.impl.Slide;
+public class SlideWithSpace extends Slide {
```
jdamerow, line 3: is this any different from the one used on the staff search?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/model/StaffSearchSlide.java
```diff
@@ -0,0 +1,28 @@
+package edu.asu.diging.vspace.core.services.impl.model;
+
+import java.util.List;
+import java.util.Map;
+import edu.asu.diging.vspace.core.model.impl.Slide;
+
+public class StaffSearchSlide {
```
jdamerow, line 7: is this any different from the one used on the staff search?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/model/StaffSearchSlideTextBlock.java
```diff
@@ -0,0 +1,39 @@
+package edu.asu.diging.vspace.core.services.impl.model;
+
+import java.util.List;
+import java.util.Map;
+
+import edu.asu.diging.vspace.core.model.impl.Slide;
+
+public class StaffSearchSlideTextBlock {
```
jdamerow, line 8: is this any different from the one used on the staff search?

vspace/src/main/java/edu/asu/diging/vspace/web/general/search/PublicSearchController.java
```diff
@@ -0,0 +1,195 @@
...
+    private void paginationForSlide(String slidePagenum, Model model, String searchTerm) {
+        Page<ISlide> slidePage = staffSearchManager.searchInSlides(searchTerm, Integer.parseInt(slidePagenum));
+        model.addAttribute("slideCurrentPageNumber", Integer.parseInt(slidePagenum));
+        model.addAttribute("slideTotalPages", slidePage.getTotalPages());
+        List<ISlide> slideList = new ArrayList<>();
+        Set<String> slideSet = slideManager.getAllSlidesFromStartSequences();
+        
+       //Adding space info for each slide
+        for(ISlide slide : slidePage.getContent()) {
+            if(slideSet.contains(slide.getId())) {
```
jdamerow, line 135: what is this condition for?

vspace/src/main/java/edu/asu/diging/vspace/web/general/search/PublicSearchController.java
```diff
@@ -0,0 +1,195 @@
...
+                        BeanUtils.copyProperties(slideWithSpace, slide);
+                    } catch (IllegalAccessException | InvocationTargetException e) {
+                        logger.error("Could not create moduleWithSpace.", e);
+                    }
+                    slideWithSpace.setSpaceId(moduleLink.getSpace().getId());
+                    slideWithSpace.setStartSequenceId(slide.getModule().getStartSequence().getId());
+                    slideList.add(slideWithSpace);
+                }
+            }
+        }
```
jdamerow, line 149: this looks like business logic and should not be in a controller

jdamerow (changes requested):
without having looked at it closely, I assume the search controllers are very similar to the staff search controllers. I would have a common super class to avoid duplications.

vspace/src/main/java/edu/asu/diging/vspace/web/general/search/PublicSearchController.java
pkharge: the search result for slides will list only those slides which are part of a start sequence. Is this not required?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/model/SlideWithSpace.java
pkharge: It seems it is only used in public search scenario where space info is added to slide.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/model/ModuleWithSpace.java
pkharge: It seems it is only used in public search scenario where space info is added to module.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SearchManager.java
```diff
@@ -0,0 +1,267 @@
...
+import edu.asu.diging.vspace.core.services.ISearchManager;
+import edu.asu.diging.vspace.core.services.ISequenceManager;
+import edu.asu.diging.vspace.core.services.ISlideManager;
+import edu.asu.diging.vspace.core.services.impl.model.StaffSearchModuleResults;
+import edu.asu.diging.vspace.core.services.impl.model.StaffSearchSlideResults;
+import edu.asu.diging.vspace.core.services.impl.model.StaffSearchSlideTextBlockResults;
+import edu.asu.diging.vspace.core.services.impl.model.StaffSearchSpaceResults;
+
+
+@Service
```
jdamerow, line 29: this shouldn't be necessary here. especially since there couldn't be an instance of an abstract class.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SearchManager.java
```diff
@@ -0,0 +1,267 @@
...
+            slidetextPage = textContentBlockRepo.findWithNameOrDescription(requestedPageForSlideText, searchTerm);
+        }
+        return slidetextPage;
+    }
+    
+    protected abstract  Page<ISpace> spaceSearch(Pageable requestedPageForSpace ,  String searchTerm);
+
+    
+    @Override
+    public StaffSearchModuleResults getStaffSearchModuleResults(List<IModule> moduleList) {
```
jdamerow, line 184: staff search related methods should be in the `StaffSearchManager`

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SearchManager.java
```diff
@@ -0,0 +1,267 @@
...
+        }
+        searchModuleResults.setModuleImageIdMap(moduleFirstSlideImage);
+        searchModuleResults.setModuleAlertMessages(isModuleConfiguredMap);
+
+        return searchModuleResults;
+    }
+
+    
+    @Override
+    public StaffSearchSlideResults getStaffSearchSlideResults(List<ISlide> slideList) {
```
jdamerow, line 221: see above

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SearchManager.java
jdamerow: for all the `get...Results` methods, what are they for? It looks like one has to first search and then pass the search results to this method? Why? Unless there is a reason this should be two different methods called from another class, one call should be enough.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SearchManager.java
```diff
@@ -0,0 +1,267 @@
...
+     *         the requested page.
+     */
+    @Override
+    public Page<ISlide> searchInSlides(String searchTerm, int page) {
+        /* if page<1, 1st page is returned */
+        if (page < 1) {
+            page = 1;
+        }
+        Pageable requestedPageForSlide = PageRequest.of(page - 1, pageSize);
+        Page<ISlide> slidePage = slideManager.findByNameOrDescription(requestedPageForSlide, searchTerm);
```
jdamerow, line 125: similar to my above comment. If a slide is not in a sequence reachable from the start sequence and the module is not linked, it should not show up in the search results.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/model/ModuleWithSpace.java
jdamerow, line 4: needs javadoc

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SearchManager.java
```diff
@@ -0,0 +1,267 @@
...
+     *         in the requested page.
+     */
+    @Override
+    public Page<IModule> searchInModules(String searchTerm, int page) {
+        /* if page<1, 1st page is returned */
+        if (page < 1) {
+            page = 1;
+        }
+        Pageable requestedPageForModule = PageRequest.of(page - 1, pageSize);
+        Page<IModule> modulePage = moduleManager.findByNameOrDescription(requestedPageForModule, searchTerm);
```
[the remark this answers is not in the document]
pkharge: Yes. But if a module is not linked to any space, it doesn't show up in the public search results. Because this method returns an empty list. edu.asu.diging.vspace.core.services.IPublicSearchManager.updateModuleListWithSpaceInfo(Page<IModule>)

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SearchManager.java
pkharge: the pagination.. methods which give the search results work differently for public search and staff search. hence both these are not merged into a single call.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SearchManager.java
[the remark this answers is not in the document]
jdamerow: how is that possible? `findByNameOrDescription` in the module manager has this code:
`return moduleRepo.findDistinctByNameContainingOrDescriptionContaining(requestedPage,searchText,searchText);`
that doesn't check if a module is reachable. Or am I missing something?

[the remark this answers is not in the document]
pkharge: Once findByNameOrDescription returns the modules, only for public search, it goes into updateModuleListWithSpaceInfo() which tries to look for linked spaces, since it doesn't find any spaces, it returns an empty list.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SearchManager.java
jdamerow: but why is this method in the searchmanager if it's a staff search specific method?

pkharge: I have renamed all the staff search methods and classes which are common to both public and staff

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SearchManager.java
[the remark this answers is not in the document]
jdamerow: but then the pagination is off isn't it?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/PublicSearchManager.java
```diff
@@ -0,0 +1,225 @@
...
+    private IModuleManager moduleManager;
+    
+    @Autowired
+    private ISlideManager slideManager;
+    
+    @Autowired
+    private TextContentBlockRepository textContentBlockRepo;
+    
+    @Override
+    protected Page<ISpace> spaceSearch(Pageable requestedPageForSpace, String searchTerm) {
```
jdamerow, line 58: method names should  be verbs/actions (e.g. searchSpaces).

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/PublicSearchManager.java
```diff
@@ -0,0 +1,225 @@
...
+                }
+                slideWithSpace.setSpaceId(moduleLink.getSpace().getId());
+                slideWithSpace.setStartSequenceId(slide.getModule().getStartSequence().getId());
+                slideList.add(slideWithSpace);
+            }
+        }
+        return slideList;
+    }
+
+    public List<IModule> updateModuleListWithSpaceInfo(Page<IModule> modulePage) {
```
jdamerow, line 85: names should be consistent, either `update...Page...` or `update...List...`

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/PublicSearchManager.java
jdamerow: this applies to several methods in this class

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SearchManager.java
```diff
@@ -0,0 +1,270 @@
...
+     * Method to return the requested spaces whose name or description contains the
+     * search string
+     * 
+     * @param searchTerm string which has been searched
+     * @param page.      The page number starts from 1.
+     * @return set of spaces whose name or description contains the search string in
+     *         the requested page.
+     */
+    @Override
+    public Page<ISpace> paginationInSpaces(String searchTerm, int page) {
```
jdamerow, line 56: similar as above; name  should be an action that describes what this method is doing.

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SearchManager.java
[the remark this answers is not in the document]
pkharge: @jdamerow question, in public search , we are only listing modules which are linked to a space. this should be Published space only right? same question for slides and slide texts.

jdamerow:
Make it so, Jenkins.

jdamerow:
build failed due to compilation errors

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SearchManager.java
```diff
@@ -0,0 +1,270 @@
...
+     * Method to return the requested spaces whose name or description contains the
+     * search string
+     * 
+     * @param searchTerm string which has been searched
+     * @param page.      The page number starts from 1.
+     * @return set of spaces whose name or description contains the search string in
+     *         the requested page.
+     */
+    @Override
+    public Page<ISpace> searchSpacesAndPaginate(String searchTerm, int page) {
```
jdamerow, line 56: I would add another abstract method `process` or `processResults` or something. The subclasses should do whatever processing is needed with the search results. It doesn't make sense that the controllers need to call those extra methods, they should have to make only one call to get the search results.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SearchManager.java
```diff
@@ -0,0 +1,270 @@
...
+                    searchTerm);
+        }
+        return slidetextPage;
+    }
+    
+
+
+    
+    @Override
+    public SearchModuleResults getSearchModuleResults(List<IModule> moduleList) {
```
jdamerow, line 187: what is this method for? Why does it require an extra call? Whoever uses `SearchManager` should only have to make one call to get results.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/StaffSearchManager.java
```diff
@@ -24,161 +26,104 @@
...
+    protected Page<ISlide> searchSlides(Pageable requestedPageForSlide, String searchTerm) {
+        return slideManager.findByNameOrDescription(requestedPageForSlide, searchTerm);
+    }
+
+    @Override
+    protected Page<ISlide> searchSlideTexts(Pageable requestedPageForSlideText, String searchTerm) {
+        return textContentBlockRepo.findWithNameOrDescription(requestedPageForSlideText,
+                searchTerm);
+    }
+
```
jdamerow, line 65: first the public methods should be listed then protected/private.

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

jdamerow (changes requested):

jdamerow:
The search bar should be a little smaller so there is some space around it in the menu.
image

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/PublicSearchManager.java
```diff
@@ -0,0 +1,223 @@
...
+                moduleList.add(modWithSpace);
+            }
+        }
+
+        return moduleList;
+    }
+
+
+    @Override
+    public List<ISlide> updateSlideTextPageWithSpaceInfo(Page<ISlide> slideTextPage) {
```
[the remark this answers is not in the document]
pkharge: @jdamerow , these methods are being called from PublicSearchController. Hence should be public right?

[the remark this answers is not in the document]
jdamerow: why isn't it called from `searchSlideTextsAndPaginate`?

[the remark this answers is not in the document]
pkharge: @jdamerow , I have a doubt, we update the model with the results right, in the REST api. should that be done on the controller end only? or that can be moved to the manager classes.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/PublicSearchManager.java
```diff
@@ -0,0 +1,160 @@
...
+                moduleList.add(modWithSpace);
+            }
+        }
+
+        return new PageImpl<>(moduleList, modulePage.getPageable(), modulePage.getTotalElements());
+    }
+
+
+    @Override
+    public Page<ISlide> updateSlideTextPageWithSpaceInfo(Page<ISlide> slideTextPage) {
```
jdamerow, line 115: does this need to be public?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/PublicSearchManager.java
```diff
@@ -0,0 +1,160 @@
...
+                slideWithSpace.setStartSequenceId(slide.getModule().getStartSequence().getId());
+                slideList.add(slideWithSpace);
+            }
+        }
+        
+        return new PageImpl<>(slideList, slidePage.getPageable(), slidePage.getTotalElements());
+    }
+
+    @Override
+    public Page<IModule> updateModulePageWithSpaceInfo(Page<IModule> modulePage) {
```
jdamerow, line 92: does this need to be public?

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
test cases fail, e.g. 
```
test_searchInModules_negativePage(edu.asu.diging.vspace.core.services.impl.PublicSearchManagerTest)  Time elapsed: 0.001 sec  <<< FAILURE!
java.lang.AssertionError: expected:<2> but was:<0>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:647)
	at org.junit.Assert.assertEquals(Assert.java:633)
	at edu.asu.diging.vspace.core.services.impl.PublicSearchManagerTest.test_searchInModules_negativePage(PublicSearchManagerTest.java:193)
```

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SearchManager.java
```diff
@@ -0,0 +1,364 @@
...
+     * This method is used to search the searched string specified in the input
+     * parameter(searchTerm) and return the published spaces corresponding to
+     * the page number specified in the input parameter(spacePagenum) whose name or
+     * description contains the search string.
+     * 
+     * @param spacePagenum current page number sent as request parameter in the URL.
+     * @param searchTerm   This is the search string which is being searched.
+     */
+    @Override
+    public SearchSpaceResults searchForSpace(String spacePagenum, String searchTerm) {
```
jdamerow, line 142: why does this and the following methods not do anything (except calling another method)? That seems unnecessary.

vspace/src/main/java/edu/asu/diging/vspace/web/general/search/PublicSearchSpaceController.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @Autowired
+    private IPublicSearchManager publicSearchManager;
+
+    @RequestMapping(value = "/exhibit/search/space")
+    public ResponseEntity<SearchSpaceResults> searchInVspace(
+            @RequestParam(value = "spacePagenum", required = false, defaultValue = "1") String spacePagenum,
+            Model model, @RequestParam(name = "searchText") String searchTerm) throws JsonProcessingException {
+
+        SearchSpaceResults publicSearch = publicSearchManager.searchForSpace(spacePagenum, searchTerm);
```
jdamerow, line 31: what's the different between `searchForSpace` and `searchSpaceAndPaginate`?

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/web/general/search/PublicSearchSpaceController.java
pkharge: Essentially, searchForSpace is called from PublicSearchSpaceController and searchSpaceAndPaginate is called from PublicSearchController.  Both these controllers have different use cases and require a different form of output. Hence these methods are kept different. Same applies to all other searches for modules/ slides/ slidetexts.

vspace/src/main/java/edu/asu/diging/vspace/web/general/search/PublicSearchController.java
```diff
@@ -0,0 +1,118 @@
+package edu.asu.diging.vspace.web.general.search;
```
jdamerow, line 1: there is a `publicpages` package in which all these controllers should probably live, or is there a reason for this new package?

vspace/src/main/java/edu/asu/diging/vspace/web/general/search/PublicSearchSpaceController.java
jdamerow: I still don't get it, it seems like `SearchSpaceResults` has the same info as the `Page` that is returned from `searchForSpaceAndPaginate`, maybe page number is missing. Why not add missing information to the results object and then only have one method? It's very confusing right now.

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/core/services/ISearchManager.java
```diff
@@ -0,0 +1,63 @@
...
+
+ 
+
+    SearchSpaceResults searchForSpace(String spacePagenum, String searchTerm);
+
+    SearchModuleResults searchForModule(String modulePagenum, String searchTerm);
+
+    SearchSlideResults searchForSlide(String slidePagenum, String searchTerm);
+
+    SearchSlideTextBlockResults searchForSlideText(String slideTextPagenum, String searchTerm);
```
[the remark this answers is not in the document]
swetalina-prusty25: added javadoc

vspace/src/main/java/edu/asu/diging/vspace/web/general/search/PublicSearchController.java
swetalina-prusty25: moved controllers to publicpages package

swetalina-prusty25:
SearchSpaceResults class has only only one element that is list of spaces. though we can not set total pages and current pages to the result object. To do that we need to add these elements in the class. So shall i create these element and convert them into one method?

jdamerow:
yes

jdamerow:
Make it so, Jenkins.

jdamerow:
Build is failing:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:compile (default-compile) on project vspace: Compilation failure
[ERROR] /var/lib/jenkins/workspace/VSpace_deploy_to_review_on_pr/vspace/src/main/java/edu/asu/diging/vspace/web/general/search/PublicSearchSpaceController.java:[31,62] cannot find symbol
[ERROR] symbol:   method searchForSpace(java.lang.String,java.lang.String)
[ERROR] location: variable publicSearchManager of type edu.asu.diging.vspace.core.services.IPublicSearchManager
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
```

pooja-thalur:
closing as i think it still has some unaddressed review comments

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SearchManager.java
pooja-thalur: Should I still look into this change? The code was changed and the comment looks old

vspace/src/main/java/edu/asu/diging/vspace/core/data/SlideRepository.java
```diff
@@ -25,5 +28,16 @@ public interface SlideRepository extends PagingAndSortingRepository<Slide, Strin
...
+    
+    @Query(value="SELECT d.Slide_Id FROM Sequence_Slides d, Module m WHERE d.Sequence_Id = m.startSequence_id", nativeQuery = true)
+    Set<String> findAllSlidesFromStartSequences();
+    
+    @Query("SELECT DISTINCT s from Slide s where  s.module.id IN "
+            + "(SELECT DISTINCT module.id from Module module JOIN ModuleLink moduleLink ON module.id = moduleLink.module.id JOIN Space space "
+            + "on space.id = moduleLink.space.id WHERE space.spaceStatus = ?3 AND module.moduleStatus = ?4) "
+            + "AND s.id IN (Select d.id from Sequence ss join ss.slides d , Module m where m.startSequence.id = ss.id)"
+            + "AND s.name like %?1% OR s.description like %?2%")
+    Page<ISlide> findDistinctByNameContainingOrDescriptionContainingLinkedToSpace(Pageable requestedPage, String name,
```
jdamerow, line 40: since this method is using a query instead of generating the query by the method name, let's give it a more handy name.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/SearchManager.java
jdamerow: this is probably fine.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/model/SearchSpaceResults.java
```diff
@@ -0,0 +1,37 @@
+package edu.asu.diging.vspace.core.services.impl.model;
+
+import java.util.List;
+
+import edu.asu.diging.vspace.core.model.ISpace;
+
+public class SearchSpaceResults {
```
jdamerow, line 7: Doesn't the page object have exactly these fields? Can't you use that one?

vspace/src/main/resources/app.properties
```diff
@@ -13,5 +13,5 @@ app_url=${app.url}
 uploads_path=${uploaded.files.path}
 file_uploads_directory=${files.directory.path}
 
-hibernate_show_sql=${hibernate.show_sql}
+hibernate_show_sql=true
```
jdamerow, line 16: should not be changed

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

vspace/src/main/java/edu/asu/diging/vspace/core/data/ModuleRepository.java
```diff
@@ -24,4 +29,10 @@ Page<IModule> findDistinctByNameContainingOrDescriptionContaining(Pageable reque
     List<Module> findAllByOrderByCreationDateDesc();
     
     List<Module> findAllByModuleStatus(ModuleStatus moduleStatus);
+    
+    List<ISequence> findAllByStartSequenceNotNull();
+       
+    @Query("SELECT DISTINCT module from Module module JOIN ModuleLink moduleLink ON module.id = moduleLink.module.id JOIN Space space on space.id = moduleLink.space.id where module.moduleStatus =?1 AND space.spaceStatus = ?2 AND ( module.name like %?3% OR module.description like %?4% )") 
+    Page<IModule> findDistinctByModuleStatusNameContainingOrDescriptionContainingLinkedToSpace(Pageable requestedPage,
```
jdamerow, line 36: If there is a query annotation, the method name really didn't need to be that long. I'm sure there is a shorter name that conveys what the method is doing that's not that long.

vspace/src/main/java/edu/asu/diging/vspace/core/data/SlideRepository.java
```diff
@@ -25,5 +28,16 @@ public interface SlideRepository extends PagingAndSortingRepository<Slide, Strin
...
+    @Query(value="SELECT d.Slide_Id FROM Sequence_Slides d, Module m WHERE d.Sequence_Id = m.startSequence_id", nativeQuery = true)
+    Set<String> findAllSlidesFromStartSequences();
+    
+    @Query("SELECT DISTINCT s from Slide s where  s.module.id IN "
+            + "(SELECT DISTINCT module.id from Module module JOIN ModuleLink moduleLink ON module.id = moduleLink.module.id JOIN Space space "
+            + "on space.id = moduleLink.space.id WHERE space.spaceStatus = ?3 AND module.moduleStatus = ?4) "
+            + "AND s.id IN (Select d.id from Sequence ss join ss.slides d , Module m where m.startSequence.id = ss.id)"
+            + "AND s.name like %?1% OR s.description like %?2%")
+    Page<ISlide> findDistinctSlidesInSpaceContainingNameOrDescription(Pageable requestedPage, String name,
+            String description, SpaceStatus spaceStatus, ModuleStatus moduleStatus);
```
jdamerow, line 41: same for those two methods (e.g. `searchInNameOrDescription`)

vspace/src/main/java/edu/asu/diging/vspace/core/data/TextContentBlockRepository.java
```diff
@@ -17,4 +19,9 @@ public interface TextContentBlockRepository extends PagingAndSortingRepository<T
     @Query("SELECT DISTINCT c.slide FROM ContentBlock c, TextBlock t WHERE c.id = t.id AND t.text LIKE %?1%")
     public Page<ISlide> findWithNameOrDescription(Pageable requestedPage, String searchText);
     
+    @Query("SELECT DISTINCT c.slide FROM ContentBlock c, TextBlock t WHERE c.id = t.id AND t.text LIKE %?1% and c.slide.module.id IN "
+            + "(SELECT DISTINCT module.id from Module module JOIN ModuleLink moduleLink ON module.id = moduleLink.module.id JOIN Space space on space.id = moduleLink.space.id WHERE space.spaceStatus = ?2 AND module.moduleStatus = ?3)"
+            + "AND c.slide.id IN (Select d.id from Sequence ss join ss.slides d , Module m where m.startSequence.id = ss.id)")
+    public Page<ISlide> findWithNameOrDescriptionLinkedToSpace(Pageable requestedPage, String searchText, SpaceStatus spaceStatus, ModuleStatus moduleStatus);
```
jdamerow, line 25: needs a better name too, what does this mean?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/PublicSearchManager.java
```diff
@@ -0,0 +1,144 @@
...
+            if(moduleLink!=null) {
+                SlideWithSpace slideWithSpace = new SlideWithSpace();
+                try {
+                    BeanUtils.copyProperties(slideWithSpace, slide);
+                } catch (IllegalAccessException | InvocationTargetException e) {
+                    logger.error("Could not create slideWithSpace.", e);
+                }
+                slideWithSpace.setSpaceId(moduleLink.getSpace().getId());
+                slideWithSpace.setStartSequenceId(slide.getModule().getStartSequence().getId());
+                slideList.add(slideWithSpace);
```
jdamerow, line 77: should the above 3 lines happen if the properties couldn't be copied?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/PublicSearchManager.java
```diff
@@ -0,0 +1,144 @@
...
+            ModuleLink moduleLink = moduleLinkManager.findFirstByModule(module);
+            if(moduleLink!=null) {
+                ModuleWithSpace modWithSpace = new ModuleWithSpace();
+                try {
+                    BeanUtils.copyProperties(modWithSpace, module);
+                } catch (IllegalAccessException | InvocationTargetException e) {
+                    logger.error("Could not create moduleWithSpace.", e);
+                }
+                modWithSpace.setSpaceId(moduleLink.getSpace().getId());
+                moduleList.add(modWithSpace);
```
jdamerow, line 98: same as before

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/PublicSearchManager.java
```diff
@@ -0,0 +1,144 @@
...
+            if(moduleLink!=null) {
+                SlideWithSpace slideWithSpace = new SlideWithSpace();
+                try {
+                    BeanUtils.copyProperties(slideWithSpace, slide);
+                } catch (IllegalAccessException | InvocationTargetException e) {
+                    logger.error("Could not create slideTextWithSpace.", e);
+                }
+                slideWithSpace.setSpaceId(moduleLink.getSpace().getId());
+                slideWithSpace.setStartSequenceId(slide.getModule().getStartSequence().getId());
+                slideTextList.add(slideWithSpace);
```
jdamerow, line 121: same as above

jdamerow (changes requested):

jdamerow:
Resolve conflicts please

PR #309, opened 2022-08-03
Story/vspc 229

pkharge:

- [xI do not have any sysout statements in my code or commented out code that isn’t needed anymore

Maybe an icon in the right upper corner somewhere.  It should have a hover text saying something like “You’re currently viewing a preview of this virtual space.” If he exhibiton is offline or in maintenance it should also say “The virtual space is currently [offline|in maintenance] and cannot via its public url.”
  
... describe here ...

### DIFF
--- a/vspace/src/main/webapp/WEB-INF/views/exhibition/space.html
+++ b/vspace/src/main/webapp/WEB-INF/views/exhibition/space.html
@@ -143,8 +143,7 @@
     z-index: 1;
     width: 50px;
     height: 44px;
-    left: 20%;
-    top: 8%;
+
     overflow: visible;
     cursor: pointer;
 }
@@ -557,16 +556,7 @@
                         class="tooltiptextforPreviewModeAlert">You’re currently viewing a preview of this virtual space.</span>
                 </button>
         </div> 
-        <div th:if="${#strings.equals(exhibitionConfig.mode,'MAINTENANCE')} OR ${#strings.equals(exhibitionConfig.mode,'OFFLINE')}" 
-			   class="exhModeAlert">
-				<button class="btn circle-btn"
-					style="width: 35px; height: 35px; margin-left: 10px; margin-top: 5px; padding: 3px;">
-					<span class="icon-warning secondary"
-						style="float: right; left: -4px;"></span> 
-					<span class="tooltiptextforExhModeAlert">The virtual space is currently [[${exhibitionConfig.mode}]] and cannot be viewed via
-						its public url.</span>
-				</button>
-        </div>
+     
 		<div class="imgResizeDiv">
                 <button class="btn circle-btn" id="toggleBtn" style="width: 35px; height: 35px; margin-left: 10px; margin-top: 5px; padding: 3px;"
                     onclick="toggleImgSize(1)">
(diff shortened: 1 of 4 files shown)


jdamerow:
Make it so, Jenkins.

jdamerow:
this message is wrong, when the exhibition is active there is a public URL:
image
When the exhibition is public, there shouldn't be any message (not needed). When it's inactive or in maintenance then there should be a message because only staff users could see the exhibition. 
Also, the icon should be somewhere else, maybe in the menu bar. Right now, it's with the other buttons that let users do something with the exhibition, but the icon is just info, nothing that the user could do.

pkharge:
@jdamerow I have shifted the icon to the left side. Can't add it to navbar since those are also action buttons for user. Let me know if these icon needs to be somewhere else

vspace/src/main/webapp/WEB-INF/views/exhibition/space.html
```diff
@@ -489,6 +550,23 @@
...
+        </div> 
+        <div th:if="${#strings.equals(exhibitionConfig.mode,'MAINTENANCE')} OR ${#strings.equals(exhibitionConfig.mode,'OFFLINE')}" 
+						class="exhModeAlert">
+				<button class="btn circle-btn"
+					style="width: 35px; height: 35px; margin-left: 10px; margin-top: 5px; padding: 3px;">
+					<span class="icon-warning secondary"
+						style="float: right; left: -4px;"></span> <span
+						class="tooltiptextforExhModeAlert">The virtual space is currently [[${exhibitionConfig.mode}]] and cannot be viewed via
+						its public url.</span>
+				</button>
```
jdamerow, line 568: indentation

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
> Can't add it to navbar since those are also action buttons for user. Let me know if these icon needs to be somewhere else
I don't understand, why can't it be in the navigation bar? I think I would put it right next to the exhibition title. And it shouldn't be buttons, just icons with hover. Otherwise it looks like clicking it would do anything.

jdamerow:
Make it so, Jenkins.

jdamerow:
It should look something like that:
image
If you hover over the badge that says "offline" (the badge should say the status so offline or maintenance), it should say: "The virtual space is currently [offline|in maintenance] and cannot be viewed via its public url." When hovering of the "preview" badge, it should say "You’re currently viewing a preview of this virtual space."

jdamerow:
Now there doesn't seem to be any icon/message.

pkharge:
Hi @jdamerow , I am able to see the icons. Could you help me understand in what scenario you are unable to see?
Screen Shot 2023-03-30 at 3 11 43 PM

jdamerow:
Never mind, I think I looked in the wrong place 🤦‍♀️ One thing though, the label text doesn't seem to be centered and the font size should be a little smaller:
image

jdamerow:
Make it so, Jenkins.

jdamerow:
Not 100% if this is related to this issue, but the module pages are broken in preview mode:
https://diging-dev.asu.edu/vspace-dev/preview/EXH_PREVIEW_fb149799/SPA000000002/module/MOD000000010/sequence/SEQ000000020/slide/SLI000000049?branchingPoint=&previousSequenceId=&aboutPageConfigured=true&isExhPreview=true

jdamerow:
Make it so, Jenkins.

PR #341, opened 2024-04-19
Story/vspc 208

pooja-thalur:

https://diging.atlassian.net/browse/VSPC-208

...Put description here...
  On the exhibition configuration page where you can set the exhibition title, etc. There should be three additional settings: scene link icon, module link icon, and external link icon. For each setting the user should be able to upload an image to be used as the default image for all links of that type. 

This ticket requires changes on the staff side (to configure the link images) and the public side (use the default image if configured) for "arrow" type links.

... describe here ...


vspace/src/main/java/edu/asu/diging/vspace/core/model/IExhibition.java
```diff
@@ -28,6 +28,18 @@ public interface IExhibition extends IVSpaceElement {
...
+
+    void setSpacelinkDefaultImage(IVSImage spacelinkDefaultImage);
+
+    IVSImage getModulelinkDefaultImage();
+
+    void setModulelinkDefaultImage(IVSImage modulelinkDefaultImage);
+
+    IVSImage getExternallinkDefaultImage();
+
+    void setExternallinkDefaultImage(IVSImage externallinkDefaultImage);
```
jdamerow, line 41: in all of the above, `link` should be `Link`

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/Exhibition.java
```diff
@@ -47,6 +48,18 @@ public class Exhibition extends VSpaceElement implements IExhibition {
...
+    @NotFound(action = NotFoundAction.IGNORE)
+    private IVSImage spacelinkDefaultImage;
+    
+    @OneToOne(targetEntity = VSImage.class)
+    @NotFound(action = NotFoundAction.IGNORE)
+    private IVSImage modulelinkDefaultImage;
+    
+    @OneToOne(targetEntity = VSImage.class)
+    @NotFound(action = NotFoundAction.IGNORE)
+    private IVSImage externallinkDefaultImage;
```
jdamerow, line 61: `Link`

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
```diff
@@ -250,4 +263,52 @@ public void removeCategory(IVSImage image, ImageCategory category) {
...
+                storedImage.setHeight(imageData.getHeight());
+                storedImage.setWidth(imageData.getWidth());
+            }
+            imageRepo.save((VSImage) storedImage);
+        }
+        return storedImage;
+    }
+
+    @Override
+    public byte[] getImageContent(IVSImage image) {
```
jdamerow, line 302: why not use storageengine's `getMediaContent` instead of this one?

jdamerow: or inside of this one is even better.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DefaultImageController.java
```diff
@@ -0,0 +1,91 @@
...
+import edu.asu.diging.vspace.core.services.IExhibitionManager;
+import edu.asu.diging.vspace.core.services.IImageService;
+
+@Controller
+public class DefaultImageController {
+
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+    public static final String API_DEFAULT_SPACE_IMAGE_PATH = "/api/image/default/link/space/";
+    public static final String API_DEFAULT_MODULEIMAGE_PATH = "/api/image/default/link/module/";
+    public static final String API_DEFAULT_EXTERNALIMAGE_PATH = "/api/image/default/link/externalLink/";
```
jdamerow, line 31: should be `.../link/external`

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DefaultImageController.java
```diff
@@ -0,0 +1,91 @@
...
+
+    @RequestMapping(value = API_DEFAULT_EXTERNALIMAGE_PATH, method = RequestMethod.GET)
+    public ResponseEntity<byte[]> getDefaultExternalImage() {
+        IExhibition exhibition = exhibitManager.getStartExhibition();
+        IVSImage externalLinkImage = exhibition.getExternallinkDefaultImage();
+        return getResponseWithDefaultHeaders(externalLinkImage);
+    }
+
+    @RequestMapping(value = API_DEFAULT_SPACE_IMAGE_DETAILS, method = RequestMethod.GET)
+    public ResponseEntity<String> getDefaultImageDetails() {
```
jdamerow, line 62: this should not be called "details" if it doesn't provide image details (it just says true/false, doesn't it?). same for the url. "image details" implies details about the image.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DefaultImageController.java
```diff
@@ -0,0 +1,91 @@
...
+
+        JsonObject jsonObj = new JsonObject();
+        jsonObj.addProperty("defaultSpaceImageFlag", defaultSpaceImageFlag);
+        jsonObj.addProperty("defaultModuleImageFlag", defaultModuleImageFlag);
+        jsonObj.addProperty("defaultExternalLinkImageFlag", defaultExternalLinkImageFlag);
+        return new ResponseEntity<>(jsonObj.toString(), HttpStatus.OK);
+    }
+
+    private ResponseEntity<byte[]> getResponseWithDefaultHeaders(IVSImage image) {
+    	byte[] imageContent = null;
```
jdamerow, line 80: define in line 84

vspace/src/main/webapp/WEB-INF/views/exhibition/space.html
```diff
@@ -391,17 +411,17 @@
...
     		        }
     		        else if(link.type=="IMAGE" && link.image != null){
                         var spaceImageURL = [[@{/api/image/}]]+link.image.id;
                         linkDisplay = $('<img class="InfoSpace_'+index+'" src="'+spaceImageURL+'" />');
     		        }
-    		        else{
-                        linkDisplay = $('<div class="InfoSpace_'+index+' Info_cz_Class"><svg class="Ellipse_8_c"><ellipse fill="rgba(222,222,222,1)" class="Ellipse_8_c_Class" rx="14.5" ry="14.5" cx="14.5" cy="14.5"></ellipse></svg><svg class="Ellipse_10_c"><ellipse fill="rgba(240,240,240,1)" class="Ellipse_10_c_Class" rx="12.5" ry="12.5" cx="12.5" cy="12.5"></ellipse></svg><svg class="Ellipse_9_c"><ellipse fill="rgba(255,255,255,1)" class="Ellipse_9_c_Class" rx="10.5" ry="10.5" cx="10.5" cy="10.5"></ellipse></svg><span class="icon-move Icon_awesome_info_c"></span><span class="tooltiptext">'+link.link.name+'</span></div>');
+    		        else if(defaultSpaceImageFlag === true) {
+    		        	var defaultSpaceURL = [[@{/api/image/default/link/space/}]];
```
jdamerow, line 423: why is this called `defaultSpaceURL` it's not a deafult space is it?

jdamerow (changes requested):
I'll review the rest when this has been reopened.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DefaultImageController.java
```diff
@@ -0,0 +1,89 @@
...
+        IVSImage defaultExternalLinkImage = exhibition.getExternalLinkDefaultImage();
+
+        boolean defaultSpaceImageFlag = defaultSpaceImage != null? true : false;
+        boolean defaultModuleImageFlag = defaultModuleImage != null? true : false;
+        boolean defaultExternalLinkImageFlag = defaultExternalLinkImage != null? true : false;
+
+        JsonObject jsonObj = new JsonObject();
+        jsonObj.addProperty("defaultSpaceImageFlag", defaultSpaceImageFlag);
+        jsonObj.addProperty("defaultModuleImageFlag", defaultModuleImageFlag);
+        jsonObj.addProperty("defaultExternalLinkImageFlag", defaultExternalLinkImageFlag);
```
jdamerow, line 74: merge above three lines with lines 67-69

jdamerow: or even better with lines 63-69

jdamerow: and add a comment why this is necessary

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DefaultImageController.java
```diff
@@ -0,0 +1,89 @@
...
+        return new ResponseEntity<>(jsonObj.toString(), HttpStatus.OK);
+    }
+
+    private ResponseEntity<byte[]> getResponseWithDefaultHeaders(IVSImage image) {
+        if(image == null) {
+            return null;
```
jdamerow, line 80: I don't think this should return null, this should return a response entity that the client understands (some kind of error code)

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ExhibitionConfigurationController.java
```diff
@@ -84,31 +88,50 @@ public RedirectView createOrUpdateExhibition(HttpServletRequest request,
...
+        
         Exhibition exhibition;
         if (exhibitID == null || exhibitID.isEmpty()) {
             exhibition = (Exhibition) exhibitFactory.createExhibition();
         } else {
             exhibition = (Exhibition) exhibitManager.getExhibitionById(exhibitID);
         }
+        
+        IVSImage spaceDefaultImage = spacelinkImage != null ? 
+                imageService.storeImage(spacelinkImage.getBytes(), spacelinkImage.getOriginalFilename()) : null;
```
jdamerow, line 110: If no spacelinkImage is passed in, then `spaceDefaultImage` will be set to null here and the default image will be deleted. But if the user does not select a new image on the configuration page, then the existing one should be retained, but I think the image wouldn't be passed to the controller again, would it? I think there are two operations needed, one to change the image (or keep the same) and one to delete the image (which probably needs to be an extra button/controller call?).

jdamerow: In fact, I think I would not include the default images with the general configuration form but have them separate actions, so that you don't have to submit the whole exhibition configuration form to change or delete the default images, those could be ajax calls.

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
jdamerow, line 1: javadoc missing

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DefaultImageController.java
```diff
@@ -0,0 +1,86 @@
...
+        return getResponseWithDefaultHeaders(externalLinkImage);
+    }
+
+    /**
+     * Retrieves the status of default images for exhibition links
+     * @return A JSON response containing flags indicating the availability of default images.
+     */
+    @RequestMapping(value = API_DEFAULT_SPACE_IMAGE_STATUS, method = RequestMethod.GET)
+    public ResponseEntity<String> getDefaultImageStatus() {
+        IExhibition exhibition = exhibitManager.getStartExhibition();exhibition.getSpaceLinkDefaultImage();
```
jdamerow, line 66: ?

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ExhibitionConfigurationController.java
```diff
@@ -99,16 +109,123 @@ public RedirectView createOrUpdateExhibition(HttpServletRequest request,
...
+     * To create or update the default space link image.
+     * 
+     * @param exhibitID - the ID of the exhibition to update
+     * @param spaceLinkImage - the MultipartFile representing the space link image to be uploaded
+     * @param attributes - the RedirectAttributes object to add flash attributes
+     * @return a RedirectView to the exhibition configuration page
+     * @throws IOException if an input or output error occurs
+     */
+    @RequestMapping(value = "/staff/exhibit/config/spaceLinkImage", method = RequestMethod.POST)
+    public RedirectView createOrUpdateSpaceLinkImage(HttpServletRequest request,
```
jdamerow, line 136: these methods should go in their own controller

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ExhibitionConfigurationController.java
```diff
@@ -99,16 +109,123 @@ public RedirectView createOrUpdateExhibition(HttpServletRequest request,
...
     }
+    
+    /**
+     * Method to return values indicating a successful operation
+     * 
+     * @param exhibition
+     * @param attributes
+     * @return 
+     */
+    private Map<String, String> addSuccessAttributes(Exhibition exhibition) {
```
jdamerow, line 223: why is that taking `Exhibition` and not `IExhibition`?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ExhibitionConfigurationController.java
```diff
@@ -99,16 +109,123 @@ public RedirectView createOrUpdateExhibition(HttpServletRequest request,
...
+    
+    /**
+     * To create or update the default external link image 
+     * 
+     * @param exhibitID - the ID of the exhibition to update
+     * @param externalLinkImage - the MultipartFile representing the external link image
+     * @param attributes - the RedirectAttributes object 
+     * @return a RedirectView to the exhibition configuration page
+     */
+    @RequestMapping(value = "/staff/exhibit/config/externalLinkImage", method = RequestMethod.POST)
```
jdamerow, line 195: It doesn't seem like there are 3 different methods needed, are there? Only one line is different, isn't it?

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ExhibitionConfigurationController.java
pooja-thalur: I have removed the 3 different methods used and changed the name of this method, so kept it as it is in the same controller because it is a common method for default link images

jdamerow:
Resolve conflicts please

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
```diff
@@ -247,6 +260,64 @@ public void addCategory(IVSImage image, ImageCategory category) {
...
         imageRepo.save((VSImage) image);
+    }   
+    
+    /**
+     * Method to store an image in the file location
+     * 
+     *@param image - The image data as a byte array
+     *@param filename - The name of the file to be stored
+     *@return {@link IVSImage} instance 
+     *@throws FileStorageException If there is an error storing the file
```
jdamerow, line 271: is that true? where is it thrown?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ExhibitionConfigurationController.java
```diff
@@ -99,16 +111,82 @@ public RedirectView createOrUpdateExhibition(HttpServletRequest request,
...
+            attributes.addAttribute("alertType", "danger");
+            attributes.addAttribute("showAlert", "true");
+            attributes.addAttribute("message", "Please select an image");
+            return new RedirectView(request.getContextPath() + "/staff/exhibit/config");
+        }
+        
+        exhibition = (Exhibition) exhibitManager.storeExhibition(exhibition);
+        attributes.addAllAttributes(addSuccessAttributes(exhibition));
+
+        return new RedirectView(request.getContextPath() + "/staff/exhibit/config");
```
jdamerow, line 175: there should be three different forms to submit the default images; each should call this function with a parameter defining which default image to set; and then map the parameter to a function that sets the default on the image at exhibition.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ExhibitionConfigurationController.java
```diff
@@ -99,16 +111,82 @@ public RedirectView createOrUpdateExhibition(HttpServletRequest request,
...
+     * @param spaceLinkImage - the MultipartFile representing the space link image to be uploaded
+     * @param moduleLinkImage - the MultipartFile representing the module link image to be uploaded
+     * @param externalLinkImage - - the MultipartFile representing the external link image to be uploaded
+     * @param attributes - - the RedirectAttributes object to add flash attributes
+     * @return - a RedirectView to the exhibition configuration page
+     * @throws IOException if an input or output error occurs
+     */
+    @RequestMapping(value = "/staff/exhibit/config/link/defaultImage", method = RequestMethod.POST)
+    public RedirectView createOrUpdateLinkImage(HttpServletRequest request,
+            @RequestParam(required = false, name = "exhibitionParam") String exhibitID,
```
jdamerow, line 141: this should not be needed; right now, you can just get the start exhibition

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DefaultImageController.java
```diff
@@ -0,0 +1,87 @@
...
+import edu.asu.diging.vspace.core.services.IExhibitionManager;
+import edu.asu.diging.vspace.core.services.IImageService;
+
+@Controller
+public class DefaultImageController {
+
+    public static final String API_DEFAULT_SPACE_IMAGE_PATH = "/api/image/default/link/space/";
+    public static final String API_DEFAULT_MODULE_IMAGE_PATH = "/api/image/default/link/module/";
+    public static final String API_DEFAULT_EXTERNAL_IMAGE_PATH = "/api/image/default/link/external/";
+    public static final String API_DEFAULT_SPACE_IMAGE_STATUS = "/api/exhibition/default/link/image/status/";
```
jdamerow, line 31: API endpoints should be in the api package.

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
There are test failures.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
```diff
@@ -247,6 +260,63 @@ public void addCategory(IVSImage image, ImageCategory category) {
     public void removeCategory(IVSImage image, ImageCategory category) {
         image.getCategories().remove(category);
         imageRepo.save((VSImage) image);
+    }   
+    
+    /**
+     * Method to store an image in the file location
```
jdamerow, line 266: what file location? this is confusing. the method only gets an image and a filename, right? nothing about location?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
```diff
@@ -247,6 +260,63 @@ public void addCategory(IVSImage image, ImageCategory category) {
...
+        CreationReturnValue returnValue = new CreationReturnValue();
+        returnValue.setErrorMsgs(new ArrayList<>());
+
+        if (storedImage != null) {
+            String relativePath = null;
+            try {
+                relativePath = storage.storeFile(image, filename, storedImage.getId());
+            } catch (FileStorageException e) {
+                logger.error(DEFAULT_IMAGE_EXCEPTION,e);
+                returnValue.getErrorMsgs().add(DEFAULT_IMAGE_EXCEPTION + e.getMessage());
```
jdamerow, line 292: if this is the case and the image can't be stored, the vsimage object should not be stored either.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ExhibitionConfigurationController.java
```diff
@@ -99,16 +112,72 @@ public RedirectView createOrUpdateExhibition(HttpServletRequest request,
...
-        if(exhibitMode.equals(ExhibitionModes.OFFLINE) && !customMessage.equals(ExhibitionModes.OFFLINE.getValue())) {
+
+        exhibitManager.updateExhibitionLanguages(exhibition, languages, defaultLanguage);
+
+        if (exhibitMode.equals(ExhibitionModes.OFFLINE) && !customMessage.equals(ExhibitionModes.OFFLINE.getValue())) {
 
             exhibition.setCustomMessage(customMessage);
         }
+
+        exhibition = (Exhibition) exhibitManager.storeExhibition(exhibition);
```
jdamerow, line 123: why does this need to be casted to `Exhibition` here?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ExhibitionConfigurationController.java
```diff
@@ -99,16 +112,72 @@ public RedirectView createOrUpdateExhibition(HttpServletRequest request,
...
+     * To create or update the default link images.
+     * 
+     * @param image - the MultipartFile representing the link image to be uploaded
+     * @param linkType - the type of the link for which the default image must be set
+     * @param attributes - the RedirectAttributes object to add flash attributes
+     * @return - a RedirectView to the exhibition configuration page
+     * @throws IOException if an input or output error occurs
+     */
+    @RequestMapping(value = "/staff/exhibit/config/link/defaultImage", method = RequestMethod.POST)
+    public RedirectView createOrUpdateLinkImage(HttpServletRequest request,
```
jdamerow, line 139: this should be in its own controller

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ExhibitionConfigurationController.java
```diff
@@ -99,16 +112,72 @@ public RedirectView createOrUpdateExhibition(HttpServletRequest request,
...
+        }
+        else if(linkType.equals("external")) {
+            IVSImage externalLinkDefaultImage = imageService.storeImage(image.getBytes(), image.getOriginalFilename());         
+            exhibition.setExternalLinkDefaultImage(externalLinkDefaultImage);
+        }
+        else {
+            attributes.addAttribute("alertType", "danger");
+            attributes.addAttribute("showAlert", "true");
+            attributes.addAttribute("message", "Couldn't save the default image");
+            return new RedirectView(request.getContextPath() + "/staff/exhibit/config");
```
jdamerow, line 160: context path should not be needed here. if the method returns `String`, you can just return `redirect:/path...`.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ExhibitionConfigurationController.java
```diff
@@ -99,16 +112,72 @@ public RedirectView createOrUpdateExhibition(HttpServletRequest request,
...
+            exhibition.setSpaceLinkDefaultImage(spaceDefaultImage);
+        }
+        else if(linkType.equals("module")) {
+            IVSImage moduleDefaultImage = imageService.storeImage(image.getBytes(), image.getOriginalFilename());            
+            exhibition.setModuleLinkDefaultImage(moduleDefaultImage);
+        }
+        else if(linkType.equals("external")) {
+            IVSImage externalLinkDefaultImage = imageService.storeImage(image.getBytes(), image.getOriginalFilename());         
+            exhibition.setExternalLinkDefaultImage(externalLinkDefaultImage);
+        }
```
jdamerow, line 155: this is pretty ugly and would get worse if we add more link types. as said before, there should be a map that maps "space", "module",... to `IExhibition::setSpace...` and then there is no if/else needed here, just getting the right method from the map and apply it.

vspace/src/test/java/edu/asu/diging/vspace/core/services/impl/ImageServiceTest.java
```diff
@@ -265,4 +273,56 @@ private Boolean checkSortByFileNameDesc(List<IVSImage> images) {
...
+        Mockito.when(imageService.getImageData(imageBytes)).thenReturn(data);
+        Mockito.when(storage.storeFile(imageBytes, FILENAME, ID)).thenReturn(STORE_PATH);
+        IVSImage returnVal = serviceToTest.storeImage(imageBytes, FILENAME);
+        Assert.assertEquals(FILENAME, returnVal.getFilename());
+        Assert.assertEquals(IMG_CONTENT_TYPE, returnVal.getFileType());
+
+    }
+
+    @Test
+    public void test_storeDefaultImage_Error() throws FileStorageException, IOException {
```
jdamerow, line 303: `error`

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ExhibitionConfigurationController.java
pooja-thalur: because exhibitManager.storeExhibition(exhibition) is returning IExhibition type

jdamerow: yes, but why would it need to be `Exhibition` here?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
jdamerow: ?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DefaultLinkImageController.java
```diff
@@ -0,0 +1,74 @@
...
+
+import edu.asu.diging.vspace.core.model.IVSImage;
+import edu.asu.diging.vspace.core.model.impl.Exhibition;
+import edu.asu.diging.vspace.core.services.IExhibitionManager;
+import edu.asu.diging.vspace.core.services.IImageService;
+
+@Controller
+public class DefaultLinkImageController {
+    
+    private static final Map<String, BiConsumer<Exhibition, IVSImage>> imageSetterMap = new HashMap<>();
```
jdamerow, line 26: this should not be static

vspace/src/main/java/edu/asu/diging/vspace/web/staff/DefaultLinkImageController.java
```diff
@@ -0,0 +1,74 @@
...
+        imageSetterMap.put("space", Exhibition::setSpaceLinkDefaultImage);
+        imageSetterMap.put("module", Exhibition::setModuleLinkDefaultImage);
+        imageSetterMap.put("external", Exhibition::setExternalLinkDefaultImage);
+    }
+    
+    @Autowired
+    IExhibitionManager exhibitManager;
+    
+    @Autowired
+    IImageService imageService;
```
jdamerow, line 38: those need to be private

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ImageService.java
pooja-thalur: in line 291 deleting the vsimage object if an exception occurs, as we are using its id to store the image

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ExhibitionConfigurationController.java
jdamerow: this still, why does exhibition need to be cast?

jdamerow: this still

vspace/src/main/webapp/WEB-INF/views/exhibition/space.html
```diff
@@ -207,24 +208,25 @@
...
-}
 
-</script>
 </head>
 
 <body>
     <div id="mainDivContent" layout:fragment="content">
         <script th:inline="javascript">
-		//# sourceURL=click.js
+        //# sourceURL=click.js
```
jdamerow, line 217: indent this line correctly

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ExhibitionConfigurationController.java
```diff
@@ -86,29 +98,35 @@ public RedirectView createOrUpdateExhibition(HttpServletRequest request,
...
         ISpace startSpace = spaceManager.getSpace(spaceID);
-
-        Exhibition exhibition;
+        IExhibition exhibition;
+        
         if (exhibitID == null || exhibitID.isEmpty()) {
             exhibition = (Exhibition) exhibitFactory.createExhibition();
         } else {
-            exhibition = (Exhibition) exhibitManager.getExhibitionById(exhibitID);
+            exhibition = (Exhibition) exhibitionManager.getExhibitionById(exhibitID);
```
jdamerow, line 110: this does not need to be cast

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ExhibitionManager.java
```diff
@@ -140,7 +142,7 @@ public void updateExhibitionLanguages(Exhibition exhibition, List<String> codes,
      */
     private IExhibitionLanguage addExhibitionLanguage(Exhibition exhibition, Map languageMap) {
```
jdamerow, line 143: should get an IExhibition

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ExhibitionManager.java
```diff
@@ -121,13 +121,15 @@ public void updateExhibitionLanguages(Exhibition exhibition, List<String> codes,
         exhibitionLanguageConfig.getExhibitionLanguageList().stream()
             .filter(languageConfig -> codes.contains(languageConfig.get(ConfigConstants.CODE)))
             .forEach(languageMap -> {
-                IExhibitionLanguage exhibitionLanguage =  addExhibitionLanguage(exhibition , languageMap);  
+                IExhibitionLanguage exhibitionLanguage =  addExhibitionLanguage((Exhibition) exhibition , languageMap);
```
jdamerow, line 124: should not be cast

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/core/data/ExhibitionRepository.java
```diff
@@ -6,11 +6,13 @@
 import org.springframework.data.repository.PagingAndSortingRepository;
 import org.springframework.stereotype.Repository;
 
+import edu.asu.diging.vspace.core.model.IExhibition;
```
jdamerow, line 9: is this change necessary?

jdamerow (changes requested):

jdamerow:
There are a few issues with this, but I'll create a new ticket for those. @pooja-thalur if you can fix the one file real quick, I'll merge.

PR #342, opened 2024-05-09
Story/vspc 99 - overview image of where a user is in a module

pooja-thalur:

https://diging.atlassian.net/browse/VSPC-99

some kind of little network that has the node representing the current slide filled or something; if that doesn't work then maybe at least the current sequence with the current slide filled
  
addressed pr comments


jdamerow:
Make it so, Jenkins.

jdamerow:
The first slide should be the beginning of the diagram (no line that goes from nothing to the node):
image

And the nodes in the diagram need to be clickable. They should get the user to the slide they represent.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ModuleOverviewManager.java
```diff
@@ -0,0 +1,89 @@
...
+        return moduleOverview;
+    }
+    
+    /**
+     * This method is used to fetch all Sequences which belong to a module and 
+     * convert this into a SequenceOverview node. The SequenceOverviewNode is added to 
+     * ModuleOverview
+     * @return ModuleOverview which contains the module and the list of sequences and its slides
+     */   
+    public SequenceOverview createSequenceOverviewNode(ISequence sequence) {
```
jdamerow, line 57: this doesn't seem to be used anywhere but in this class, so it shouldn't be public (unless it's indented to be used anywhere else? in that case it should be part of the interface).

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ModuleOverviewManager.java
```diff
@@ -0,0 +1,89 @@
...
+        
+        SequenceOverview sequenceOverview = new SequenceOverview();
+        sequenceOverview.setName(sequence.getName());
+        sequenceOverview.setId(sequence.getId());
+        List<SlideOverview> slideOverviews = createSlideOverviewNode(sequence.getSlides());
+        sequenceOverview.setSlideOverviews(slideOverviews);
+        return sequenceOverview;  
+    }
+    
+    public List<SlideOverview> createSlideOverviewNode(List<ISlide> slides){
```
jdamerow, line 70: see above

vspace/src/main/webapp/WEB-INF/views/exhibition/module.html
```diff
@@ -24,9 +53,14 @@
...
-      </style>
-      <script th:inline="javascript">
+</style>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/d3/3.5.17/d3.min.js"></script>
+<script th:inline="javascript">
+        var json =[[${overview}]];
+        var moduleId = [[${module.id}]];
+        var spaceId = [[${spaceId}]];
+        var currentSlide = [[${currentSlideCon}]];
+        var sequenceId = [[${currentSequenceId}]];
```
jdamerow, line 63: why are these defined up here and not where they are bing used?

vspace/src/main/webapp/WEB-INF/views/exhibition/module.html
```diff
@@ -138,25 +191,261 @@ <h3 style="text-align:center;">[[${currentSlideCon.name}]]</h3>
...
                             class="list-group-item list-group-item-action">[[${choice.sequence.name}]]</a>
+				   </div>
+
+				   <div class="list-group" th:if="${contents.showsAll eq false}" th:each="choice: ${contents.choices}">
+				        <a th:href="${isExhPreview == null OR isExhPreview == false}?@{'/exhibit/'+${spaceId}+'/module/'+${module.id}+'/sequence/'+${choice.sequence.id}+'?branchingPoint='+${slideId}+'&previousSequenceId='+${currentSequenceId}}:@{'/preview/'+${previewId}+'/'+${spaceId}+'/module/'+${module.id}+'/sequence/'+${choice.sequence.id}+'?branchingPoint='+${slideId}+'&previousSequenceId='+${currentSequenceId}}"
+				            class="list-group-item list-group-item-action">[[${choice.sequence.name}]]</a>
+				    </div>
+				</div>
+			</div>
+		</div>
```
jdamerow, line 202: indentation of the above block

vspace/src/main/webapp/WEB-INF/views/exhibition/module.html
```diff
@@ -138,25 +191,261 @@ <h3 style="text-align:center;">[[${currentSlideCon.name}]]</h3>
...
+                        const branchingPointsSet = new Set();
+                        const sequencesMap = new Map();
+                        function createFromSequence(sequence, treeData, isStartSequence){
+                            let startNodeSequence = {};
+                            let startNodeSequenceSlides = [];
+                            startNodeSequence.children = [];
+                            startNodeSequence.name = sequence.name;
+                            startNodeSequence.id = sequence.id;
+                            startNodeSequence.group = "sequence";
+                        	//we will use parent identifier as id instead of name
```
jdamerow, line 223: why is the following block indented?

vspace/src/main/webapp/WEB-INF/views/exhibition/module.html
```diff
@@ -138,25 +191,261 @@ <h3 style="text-align:center;">[[${currentSlideCon.name}]]</h3>
...
+                        	startNodeSequence.parent = json.id;
+                        	let slideIndex = 0;
+                        	for(slide of sequence.slideOverviews){
+                        		let slideNode = {}
+                        		slideNode.children=[];
+                        		if(slide.branchingPoint == true){
+                        		    slideNode.branchSequences = [];
+                        		    slideNode.branchingPoint = true;
+                        		    slideNode.branchingPoints = [];
+                        			for(seq of slide.choiceSequenceNames){
```
jdamerow, line 233: and here

vspace/src/main/webapp/WEB-INF/views/exhibition/module.html
```diff
@@ -138,25 +191,261 @@ <h3 style="text-align:center;">[[${currentSlideCon.name}]]</h3>
...
+                        			    if(seq === sequence.name){
+                        			        let branchSeq = {};
```
jdamerow, line 235: define outside of the if to avoid duplication

vspace/src/main/webapp/WEB-INF/views/exhibition/module.html
```diff
@@ -138,25 +191,261 @@ <h3 style="text-align:center;">[[${currentSlideCon.name}]]</h3>
...
+                        			        branchSeq.name = seq;
```
jdamerow, line 236: and this too

vspace/src/main/webapp/WEB-INF/views/exhibition/module.html
```diff
@@ -138,25 +191,261 @@ <h3 style="text-align:center;">[[${currentSlideCon.name}]]</h3>
...
+                        var diagonal = d3.svg.diagonal().projection(function(d) { return [d.y, d.x]; });
+                        var svg = d3.select("#graph").append("svg").attr("width", width + margin.right + margin.left)
+                                    .attr("height", height + margin.top + margin.bottom)
+                                    .append("g")
+                                    .attr("transform", "translate(" + 208 + "," + margin.top + ")");
+                        root = treeData;
+                        root.x0 = height / 2;
+                        root.y0 = 0;
+                        
+                        update(root);
```
jdamerow, line 306: all the code above needs to be commented

jdamerow: extracting smaller methods also wouldn't hurt

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
- When the window is too small to show the whole modal, I have no way of seeing the bottom of the modal as it doesn't scroll with the page.
- branching points should not show up as a point, only the slides: "Russia Sequence" should not be here:
image

jdamerow:
Make it so, Jenkins.

jdamerow:
module map doesn't load for: https://diging-dev.asu.edu/vspace-dev/exhibit/SPA1/module/MOD000000001/sequence/SEQ000000007/slide/SLI000000015?branchingPoint=&previousSequenceId=&aboutPageConfigured=true

Javascript error:
`TypeError: Cannot read properties of undefined (reading 'normalize')
    at SVGCircleElement.<anonymous> (SLI000000015?branchingPoint=&previousSequenceId=&aboutPageConfigured=true:524:43)`

Here is the json:
```

                        const branchingPointsSet = new Set();
                        const sequencesMap = new Map();
                        var json ={"name":"Real module","id":"MOD000000001","startSequence":{"id":"SEQ000000007","name":"Cities","slideOverviews":[{"id":"SLI000000015","name":"Bern","choiceSequenceNames":null,"branchingPoint":false},{"id":"SLI000000016","name":"Verona","choiceSequenceNames":null,"branchingPoint":false},{"id":"SLI000000019","name":"Paris","choiceSequenceNames":null,"branchingPoint":false},{"id":"SLI000000018","name":"Where to go?","choiceSequenceNames":["Cities","Germany"],"branchingPoint":true}]},"otherSequences":[{"id":"SEQ000000008","name":"Germany","slideOverviews":[{"id":"SLI000000017","name":"Stuttgart","choiceSequenceNames":null,"branchingPoint":false}]},{"id":"SEQ000000013","name":"test","slideOverviews":[{"id":"SLI000000015","name":"Bern","choiceSequenceNames":null,"branchingPoint":false}]}]};
                        var moduleId = "MOD000000001";
                        var spaceId = "SPA1";
                        var currentSlide = {"name":"Bern","description":"","createdBy":"admin","creationDate":"2020-05-06T12:09:02Z","modifiedBy":"admin","modificationDate":"2020-05-06T12:09:16Z","id":"SLI000000015","module":{"name":"Real module","description":"test","createdBy":"admin","creationDate":"2020-05-06T12:07:05Z","modifiedBy":"admin","modificationDate":"2020-05-06T12:16:26Z","id":"MOD000000001","moduleStatus":null,"sequences":[]}};
                        var sequenceId = "SEQ000000007";

```

jdamerow:
Make it so, Jenkins.

jdamerow:
This is great! The only last thing is that the modal that opens with the network is way too long almost three times as long as necessary.

jdamerow:
Make it so, Jenkins.

PR #336, opened 2023-09-13
Story/vspc 232 214 215 multiple languages

pkharge:
# Guidelines for Pull Requests

If you haven't yet read our code review guidelines, please do so, You can find them [here](https://diging.atlassian.net/wiki/spaces/DIGING/pages/2256076801/Code+Review+Guidelines).

Please confirm the following by adding an x for each item (turn `[ ]` into `[x]`).

- [x] I have removed all code style changes that are not necessary (e.g. changing blanks across the whole file that don’t need to be changed, adding empty lines in parts other than your own code)
- [x] I am not making any changes to files that don’t have any effect (e.g. imports added that don’t need to be added)
- [x] I do not have any sysout statements in my code or commented out code that isn’t needed anymore
- [x] I am not reformatting any files in the wrong format or without cause. 
- [x] I am not changing file encoding or line endings to something else than UTF-8, LF
- [x] My pull request does not show an insane amount of files being changed although my ticket only requires a few files being changed
- [x] I have added Javadoc/documentation where appropriate
- [x] I have added test cases where appropriate
- [x] I have explained any part of my code/implementation decisions that is not be self-explanatory

## Please provide a brief description of your ticket 
(you can copy the ticket if it hasn't changed) 

This pull request combines 3 stories

VSPC-232 - It should be possible to specify about text in multiple languages
VSPC-214 - If there are multiple languages specified for an exhibition, the user should be able to enter text for a slide in each language.
VSPC-215 - If there are multiple languages defined for an exhibition, staff users should be able to specify space title and descriptions for each language.
  
## Anything else the reviewer needs to know?

Upgraded  log4j to resolve this codefactor issue - Use of vulnerable log4j:log4j

PR #348, opened 2024-06-20
Bug/vspc 271 - Exhibition Configuration: if not exhibition language should be required

pooja-thalur:

[Exhibition Configuration: if not exhibition language should be required](https://diging.atlassian.net/browse/VSPC-271)

...Put description here...
The user should select at least one exhibition language from the list (right now there is a tomcat error). There should also be a field help “To select multiple languages, hold down the shift key when selecting.”

... describe here ...


jdamerow:
There should also be server-side validation.

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ExhibitionConfigurationController.java
```diff
@@ -99,6 +99,13 @@ public RedirectView createOrUpdateExhibition(HttpServletRequest request,
         exhibition.setStartSpace(startSpace);
         exhibition.setTitle(title);
         exhibition.setMode(exhibitMode);
+        
+        if(languages.isEmpty()) {
```
jdamerow, line 103: I believe you can simply make the RequestParam `exhibitLanguage` required instead. Have you tried that?

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ExhibitionConfigurationController.java
pooja-thalur: yes, I checked that - the RequestParam exhibitLanguage is required by default and it was giving the server error. 

Can I make it required = false and add a null check here to give out an error message in the ui or remove this if block and keep it as before?

jdamerow: Given that there are quite a few request parameter by now, I would create a class the paramters and validate that. Similar to how it's done here:
https://www.baeldung.com/spring-mvc-form-tutorial

vspace/src/main/java/edu/asu/diging/vspace/web/staff/forms/ExhibitionConfigurationForm.java
```diff
@@ -0,0 +1,72 @@
...
+
+import edu.asu.diging.vspace.core.model.ExhibitionModes;
+
+public class ExhibitionConfigurationForm {
+    
+    private String customMessage;
+    
+    private String defaultExhibitLanguage;    
+    
+    private String exhibitionParam;
```
jdamerow, line 16: what's that? what's a exhibition param?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/forms/ExhibitionConfigurationForm.java
```diff
@@ -0,0 +1,72 @@
...
+    
+    @NotEmpty(message = "Please select exhibition languages")
+    private List<String> exhibitLanguage;
+    
+    @NotNull(message = "Please select exhibit mode")
+    private ExhibitionModes exhibitMode;
+    
+    @NotEmpty
+    private String spaceParam;
```
jdamerow, line 25: what's space param?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/forms/ExhibitionConfigurationForm.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+    public void setExhibitMode(ExhibitionModes exhibitMode) {
+        this.exhibitMode = exhibitMode;
+    }  
+    public void setSpaceParam(String spaceParam) {
+        this.spaceParam = spaceParam;
+    }
+    public void setTitle(String title) {
+        this.title = title;
+    }
```
jdamerow, line 71: getters and setters should be together

jdamerow (changes requested):

vspace/src/main/java/edu/asu/diging/vspace/web/staff/forms/ExhibitionConfigurationForm.java
pooja-thalur: it was passed from the form, removed it from here now

vspace/src/main/java/edu/asu/diging/vspace/web/staff/forms/ExhibitionConfigurationForm.java
```diff
@@ -0,0 +1,64 @@
...
+    
+    private String customMessage;
+    
+    private String defaultExhibitLanguage;
+    
+    @NotEmpty(message = "Please select exhibition languages")
+    private List<String> exhibitLanguage;
+    
+    @NotNull(message = "Please select exhibit mode")
+    private ExhibitionModes exhibitMode;
```
jdamerow, line 20: there is not really a reason to save 3 characters, just write out "exhibition" (and in the message it definitely need to be written out)

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
after that should be good

jdamerow:
There is something off (I think when trying to remove a language that had already text associated with it?):
```
java.sql.SQLIntegrityConstraintViolationException: Cannot delete or update a parent row: a foreign key constraint fails (`vspace_dev`.`LocalizedText`, CONSTRAINT `FK9pyq98dqhy0q6nmcxt1yheju9` FOREIGN KEY (`LOC_EXH_LANG`) REFERENCES `ExhibitionLanguage` (`id`))
```

pooja-thalur:
> There is something off (I think when trying to remove a language that had already text associated with it?):
> 
> ```
> java.sql.SQLIntegrityConstraintViolationException: Cannot delete or update a parent row: a foreign key constraint fails (`vspace_dev`.`LocalizedText`, CONSTRAINT `FK9pyq98dqhy0q6nmcxt1yheju9` FOREIGN KEY (`LOC_EXH_LANG`) REFERENCES `ExhibitionLanguage` (`id`))
> ```

The foreign key constraint was added in this file (vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/LocalizedText.java) while testing this pr (https://github.com/diging/virtual-spaces-2.0/pull/336) in the db. I think it should work fine if the constraint is dropped, or should there be any extra handling required?

vspace/src/main/java/edu/asu/diging/vspace/web/staff/ExhibitionConfigurationController.java
```diff
@@ -77,19 +81,26 @@ public String showExhibitions(Model model) {
...
+            attributes.addAttribute("alertType", "danger");
+            attributes.addAttribute("message", result.getFieldError().getDefaultMessage());
+            return new RedirectView(request.getContextPath() + "/staff/exhibit/config");
+        }
+        String title = exhibitionConfigForm.getTitle();
+        ExhibitionModes exhibitMode = exhibitionConfigForm.getExhibitionMode();
+        List<String> languages = exhibitionConfigForm.getExhibitLanguage();
+        String defaultLanguage = exhibitionConfigForm.getDefaultExhibitLanguage();
+        String customMessage = exhibitionConfigForm.getCustomMessage();
+        ISpace startSpace = spaceManager.getSpace(spaceID);
```
jdamerow, line 103: If these are not use more than once, just merge these links with where they are used, e.g.
`exhibition.setStartSpace(paceManager.getSpace(spaceID));`

jdamerow (changes requested):

jdamerow:
Also, there are conflicts.

jdamerow:
Make it so, Jenkins.

PR #354, opened 2025-05-23
Story/vspc 100

jophals:

there should be a functionality to automatically generate overview maps [Ticket](https://diging.atlassian.net/browse/VSPC-100?atlOrigin=eyJpIjoiYWFhNGNjODdhODMyNDhlNjk1MDE2YTFlM2QyNTI0MzMiLCJwIjoiaiJ9)
  
probably for now just a network of all spaces and modules that the user can adjust and then save the adjustments to be used in an overview map

### DIFF
--- a/vspace/src/main/webapp/WEB-INF/views/staff/spaces/graph.html
+++ b/vspace/src/main/webapp/WEB-INF/views/staff/spaces/graph.html
@@ -116,8 +116,11 @@
 					      .style("cursor", "default");
 					})
 	            	.on('click', function(event, d) {
+	            		//Reset highlighting
 	            		d3.selectAll("circle")
 							.style("fill", function(d) { return d.depth === 1 ? "#fff" : "#000"; });
+	            		
+	            		//Update card for node clicked
 	            		if (d.data.img) {
 		            		document.getElementById("space-overview-card-img").src = d.data.img;
 	            		} else {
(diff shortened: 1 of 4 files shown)


jdamerow:
Test failures

jdamerow:
Make it so, Jenkins.

vspace/src/main/webapp/WEB-INF/views/staff/spaces/graph.html
```diff
@@ -1,227 +1,222 @@
...
+        	}
         </style>
     </head>
     <body>
         <div layout:fragment="content">
         <div class="container">
-            <div id="graph" style="width: 100%;height: 100%;"></div>
+            <div id="graph" style="width: 100%;height: 100%; background-color: #909090"></div>
+            <div class="card" style="width: 18rem;">
+			  <img src="/vspace/resources/maintenance/img/logo-11.png" class="card-img-top" id="space-overview-card-img" alt="...">
```
jdamerow: this image does not work (not all instances are under `/vspace`).

vspace/src/main/webapp/WEB-INF/views/staff/spaces/graph.html
```diff
@@ -1,227 +1,222 @@
...
+			  <div class="card-body" id="space-overview-card-body">
+			    <h5 class="card-title" id="space-overview-card-title">Space Overview Map</h5>
+			    <p class="card-text" id="space-overview-card-text">This is a map of all spaces and modules and their links with one another. Feel free to click or drag nodes to interact with them.</p>
+			    <a href="#" class="btn btn-primary" id="space-overview-btn">Click a space or module node...</a>
```
jdamerow: not clear why this is a link/button. Should anything happen when you click onit?

jdamerow:
Also the layout is not well-suited for this. I can move the nodes but they won't stay where I moved them, so what's the point? And the color scheme should match the rest of the website.

jdamerow:
Make it so, Jenkins.

jdamerow:
image

jdamerow:
Make it so, Jenkins.

jdamerow:
There seem way too many nodes; modules are duplicated and sometimes linked more than once from the same space:
image

jdamerow:
Make it so, Jenkins.

jdamerow:
I'm going back and forth on this, but I think usuallly, modules will only be linked from one or two spaces, so module nodes should not be duplicated. Instead all spaces should link to the same module node if they link to the same module. Also, in the pop up:
image
"Part of Space" needs to be "Linked from:" and then list all spaces the module is linked from.

Girik1105:
> I'm going back and forth on this, but I think usuallly, modules will only be linked from one or two spaces, so module nodes should not be duplicated. Instead all spaces should link to the same module node if they link to the same module. Also, in the pop up: image "Part of Space" needs to be "Linked from:" and then list all spaces the module is linked from.

Should we also show all modules or only modules that are linked to a space? Currently it is set up such that all spaces are visible and only modules that are linked to spaces are shown.

jdamerow:
good point. all modules should be shown at this point.

jdamerow:
Make it so, Jenkins.

PR #356, opened 2025-06-20
story/VSPC-276

Girik1105:

Display of references on slides

references should be in apa format
  
[VSPC-276](https://diging.atlassian.net/browse/VSPC-276?atlOrigin=eyJpIjoiNjIxZGZlZDI5YjBjNDU1MGFhMzFiMzZmYjQ4ZTI1NjgiLCJwIjoiaiJ9)

... describe here ...

[VSPC-276]: https://diging.atlassian.net/browse/VSPC-276?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vspace/src/main/java/edu/asu/diging/vspace/core/util/CitationFormatter.java
```diff
@@ -0,0 +1,218 @@
+package edu.asu.diging.vspace.core.util;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class CitationFormatter {
```
jdamerow: I'm confused about what this class is for. We don't have in text citations, just a list of references, so we don't need any in-text citation formatting.

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/TextBlock.java
```diff
@@ -46,9 +47,58 @@ public void setText(String text) {
     @Override
     @Transient
     public String htmlRenderedText() {
```
jdamerow, line 48: the citations are in a biblio block, not a text block. So I don't think this is right.

jdamerow (changes requested):

jdamerow:
I think you misunderstood the ticket. This is for the biblio blocks, not the text blocks.

jdamerow (changes requested):
this should use the class in the 150 branch.

jdamerow (changes requested):
The references classes are not used it seems. The ReferenceDisplayFormatter should be used to build a reference in a certain format (not hand-build in the AP Citation formatter). The idea is that different formatters call the reference display formatter in a different order, with different parameters (e.g. parentheses or no parentheses for the year). 

The IReferenceMetadataRegistry would be called to get a provider for a specific metadatatype (e.g. APA). The provider would then handle building the metadata string using the display formatter. Either in the template or the controller, the provider registry would be called to get a provider. The bibliobock itself should not worry about the display part (otherwise, it would get bigger and bigger with each new citation style we add).

vspace/src/main/java/edu/asu/diging/vspace/core/references/ReferenceMetadataRegistryHolder.java
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Static holder for the IReferenceMetadataRegistry.
+ *
+ * This class allows JPA entities (like BiblioBlock) to access the registry
+ * without requiring dependency injection, since entities cannot use @Autowired.
+ *
+ * The registry is set via Spring autowiring when this component is initialized.
+ */
+@Component
+public class ReferenceMetadataRegistryHolder {
```
jdamerow: what is this class for?

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/BiblioBlock.java
```diff
@@ -57,5 +62,153 @@ public void setReferences(List<Reference> references) {
...
+     * Renders the bibliography block with APA-formatted references.
+     *
+     * This method delegates to the IReferenceMetadataRegistry to get the APA provider,
+     * which uses ReferenceDisplayFormatter to build the citations.
+     *
+     * @return HTML formatted string with APA-style references
+     */
+    @Override
+    @Transient
+    public String renderAPAReferences() {
```
jdamerow: this should not be here. the block should not do the rendering, that's what the APAReferenceMetadataProvider is for.

jdamerow (changes requested):
It might make sense to do the rendering in Javascript rather than in the backend. Let's maybe give this one a try: https://github.com/larsgw/citation.js. It should make the backend less complex and might give us more options in the frontend.

PR #360, opened 2025-08-28
story/VSPC-277

Girik1105:

Adding new references to a slide issues

1. New Reference button should be an icon instead of a button
2. Title should be in bold and description should be below it in appropriate format
3. Fix misaligned delete buttons

[VSPC-277](https://diging.atlassian.net/browse/VSPC-277?atlOrigin=eyJpIjoiYzJmNWRkYTYyODc5NGIwZjhiZWEzN2VmZWZlY2E3YmQiLCJwIjoiaiJ9)
  
... describe here ...

[VSPC-277]: https://diging.atlassian.net/browse/VSPC-277?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jdamerow:
Make it so, Jenkins.

jdamerow:
- the button to add a new reference should be just an icon (no button)
- not sure what's going on, but when I double-click on the reference block, it disappears and the pop up appears. But when I click submit, there are just errors.
image

- after I created a new blibliography block and added a reference, when I try to edit the bibliography block it shows wrong data:
image

but should be:
image

vspace/src/main/java/edu/asu/diging/vspace/core/util/CitationFormatter.java
```diff
@@ -0,0 +1,218 @@
+package edu.asu.diging.vspace.core.util;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class CitationFormatter {
```
jdamerow: is this class used in this pr?

jdamerow:
Resolve conflicts please

jdamerow:
Make it so, Jenkins.

jdamerow:
- when I click on the edit button for a bibliography block, the block disappears from the page
- adding a new reference, doesn't work, server returns 500 (` java.sql.SQLException: Field &#39;isPublic &#39;doesn &#39;t have a default value`)
- if there are two bibliography blocks, when I edit the block description of the second block, after I save the changes, the first block shows all reference while the second block shows none.

PR #361, opened 2025-09-03
task/VSPC-209

Girik1105:

ContentBlockManager class needs to be separated into multiple manager classes

Description

The contentblockmanager class is getting too big. Let's separate the logic out into multiple manager classes, one for each type of content block.
There is a lot of duplicate code in there. Let’s extract a superclass with a shared delete and and get method, and an abstract getRepository method. The subclasses will then implement getRepository to return the repository for the class they manage, and create and update methods.
Additionally, contentOrder should not be a parameter for the create methods. I’m fairly certain the controllers just use the findMaxContentOrder method to get the last index and then pass that into the create methods (please check). Instead, the create methods should call the findMaxContentOrder .

[VSPC-209](https://diging.atlassian.net/browse/VSPC-209?atlOrigin=eyJpIjoiM2RhNTJkMDVkOGIzNGZjNDk2MTNmMjY5M2Y5Mjc3MDIiLCJwIjoiaiJ9)
  
... describe here ...

[VSPC-209]: https://diging.atlassian.net/browse/VSPC-209?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ContentBlockManager.java
jdamerow, line 1: This should be the generic class, that get parametrized with the type of object it handles and the repo doing the CRUD functionality, e.g.
```
abstract class ContentBlockManager<T, R> {

...
public void deleteBlockById(String id, String slideId) {
   // here you get the repository via an abstract getRepository method
   // that is implemented in the subclasses
  getRepository().delete(...
}
...
```
and then the subclasses implement the getRepository method and all methods that are specific to a certain block type. 

Does that make sense?

jdamerow (changes requested):

PR #364, opened 2025-09-18
story/VSPC-189

Girik1105:

It should be possible to pull in bibliography entries from citesphere

Description
Instead of entering bibliography entries by hand, the user would select entries from Citesphere and those would then be imported. Since Citesphere doesn't have a search endpoint yet, the user would when adding a bibliography block, browse their groups in citesphere and select entries to import. So:
1. user adds bibliography block on a slide
2. instead of adding a reference by hand, they log in to Citesphere (oauth authorization flow)
3. they see all groups they are part of
4. they select a group and then one or multiple references from that group
5. The selected references will be imported into VSpace
For now, this would be a one time import. We'll worry about updating references from Citesphere later.
The code that pull data from Citesphere should be its own plugin (have Julia create the basic plugin).

[VSPC-189](https://diging.atlassian.net/browse/VSPC-189?atlOrigin=eyJpIjoiOWQ5MWRhZDhiNjUxNDA0NTlmMTNmNjk3M2UyMzg3ZWEiLCJwIjoiaiJ9)
  
... describe here ...

[VSPC-189]: https://diging.atlassian.net/browse/VSPC-189?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jdamerow:
Resolve conflicts please

vspace/src/main/java/edu/asu/diging/vspace/core/services/CitesphereAuthToken.java
```diff
@@ -0,0 +1,115 @@
...
+ * Authentication token object for Citesphere API
+ */
+public class CitesphereAuthToken {
+    
+    private String authType;
+    private Map<String, String> headers;
+    private String accessToken;
+    private String refreshToken;
+    private String username;
+    private String password;
```
jdamerow: there should be no password needed here.

.gitignore
jdamerow, line 1: Those entries should probably not be removed

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/CitesphereManager.java
```diff
@@ -0,0 +1,450 @@
...
+        if (authTokenObject.getAccessToken() == null) {
+            if (authTokenObject.getUsername() == null || authTokenObject.getPassword() == null) {
+                throw new IllegalArgumentException(
+                    "Either username and password or access_token should be present");
+            }
+        }
+        
+        if (!"oauth".equals(authTokenObject.getAuthType()) && 
+            !"basic".equals(authTokenObject.getAuthType())) {
+            throw new IllegalArgumentException("authType should be either oauth or basic");
```
jdamerow: there should be no basic authentication with citesphere.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/CitesphereManager.java
```diff
@@ -0,0 +1,450 @@
...
+                for (Map.Entry<String, String> header : authTokenObject.getHeaders().entrySet()) {
+                    requestBuilder.addHeader(header.getKey(), header.getValue());
+                }
+            }
+            
+            Request request = requestBuilder.build();
+            
+            try (Response response = client.newCall(request).execute()) {
+                // Check for unauthorized or forbidden responses
+                if (response.code() == 401 || response.code() == 403) {
```
jdamerow: the http codes are defined in constants (not sure what the class is called, it depends on the included library, maybe maybe `HttpStatus` or something along those lines?)

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/CitesphereManager.java
```diff
@@ -0,0 +1,450 @@
...
+                        // Response is an object
+                        @SuppressWarnings("unchecked")
+                        Map<String, Object> responseMap = objectMapper.readValue(responseBody, Map.class);
+                        return responseMap;
+                    }
+                }
+            }
+        } catch (CitesphereTokenException e) {
+            throw e; // Re-throw token exceptions
+        } catch (Exception e) {
```
jdamerow: what is throwing `Exception` (specific exceptions should be caught)? If it's not thrown anywhere, then this should be removed/replace with specific exceptions and the catch block above catching `CitesphereTokenException` would also not be necessary.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/CitesphereManager.java
```diff
@@ -0,0 +1,450 @@
...
+            }
+            
+            // Add file if provided
+            if (filePath != null) {
+                File file = new File(filePath);
+                if (file.exists()) {
+                    RequestBody fileBody = RequestBody.create(file, MediaType.parse("application/pdf"));
+                    builder.addFormDataPart("files", file.getName(), fileBody);
+                }
+            }
```
jdamerow: why is this needed? VSpace doesn't send files to citesphere, does it?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/CitesphereManager.java
```diff
@@ -0,0 +1,450 @@
...
+            
+            Request request = requestBuilder.build();
+            
+            try (Response response = client.newCall(request).execute()) {
+                return response;
+            }
+            
+        } catch (Exception e) {
```
jdamerow: exceptions should be specific

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/CitesphereManager.java
```diff
@@ -0,0 +1,450 @@
...
+        } catch (CitesphereTokenException e) {
+            Map<String, Object> errorMap = new HashMap<>();
+            errorMap.put("error_message", e.getMessage());
+            errorMap.put("token_expired", e.isTokenExpired());
+            return errorMap;
+        }
+    }
+    
+    @Override
+    public Map<String, Object> checkTest() {
```
jdamerow: what is this for?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/CitesphereManager.java
```diff
@@ -0,0 +1,450 @@
...
+        } catch (CitesphereTokenException e) {
+            Map<String, Object> errorMap = new HashMap<>();
+            errorMap.put("error_message", e.getMessage());
+            errorMap.put("token_expired", e.isTokenExpired());
+            return errorMap;
+        }
+    }
+    
+    @Override
+    public Map<String, Object> checkAccess(String documentId) {
```
jdamerow: I don't think this is needed, is it?

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/CitesphereManager.java
```diff
@@ -0,0 +1,450 @@
...
+        try {
+            String url = api + "/v1/groups/" + groupId;
+            return executeCommand(url);
+        } catch (CitesphereTokenException e) {
+            Map<String, Object> errorMap = new HashMap<>();
+            errorMap.put("error_message", e.getMessage());
+            errorMap.put("token_expired", e.isTokenExpired());
+            return errorMap;
+        }
+    }
```
jdamerow, line 163: many of these methods look almost identical (except the line that builds the url), so there should be one method with this call and the other methods would just build the url and call that method.

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/CitesphereManager.java
```diff
@@ -0,0 +1,450 @@
...
+                .add("refresh_token", authTokenObject.getRefreshToken())
+                .build();
+
+            Request request = new Request.Builder()
+                .url(tokenUrl)
+                .post(formBody)
+                .build();
+
+            try (Response response = client.newCall(request).execute()) {
+                if (response.code() == 401 || response.code() == 403) {
```
jdamerow: see above

vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/CitesphereManager.java
```diff
@@ -0,0 +1,450 @@
...
+                        
+                        return authTokenObject;
+                    }
+                }
+                
+                throw new CitesphereTokenException("Failed to refresh token: " + response.code());
+            }
+        } catch (CitesphereTokenException e) {
+            throw e;
+        } catch (Exception e) {
```
jdamerow: see above

jdamerow (changes requested):
gotta run, but I'll leave it with this for now.

PR #367, opened 2025-10-10
story/VSPC-216

Girik1105:

- [ ] My pull request does not show an insane amount of files being changed although my ticket only requires a few files being changed

If there are multiple languages defined, users of the public exhibition should be able to switch between them

Description

there needs to be a menu item to switch languages. Depending on the selected language, text content is shown in that language. By default the default language should be shown. If something doesn’t have text provided in the selected language (e.g. French is selected but there is no title and description for a space in French), then the default language should be shown.
  
... describe here ...


jdamerow:
Resolve conflicts please

PR #370, opened 2025-10-30
story/VSPC-150

Girik1105:

Add Bibliography section to staff page

Like users or images, it should show all references sortable by author, year, title, or created date.

[VSPC-150](https://diging.atlassian.net/browse/VSPC-150?atlOrigin=eyJpIjoiMjQ5MGIwNmM2Njg5NDFjYmI0ZDAwY2YyMTllNzRlN2IiLCJwIjoiaiJ9)
  
The previous PR for this branch [299](https://github.com/diging/virtual-spaces-2.0/pull/299) has a force push and hence cannot be re opened. 

[VSPC-150]: https://diging.atlassian.net/browse/VSPC-150?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vspace/src/main/java/edu/asu/diging/vspace/core/services/impl/ReferenceManager.java
```diff
@@ -42,6 +52,13 @@ public IReference createReference(String biblioId, String title, String author,S
         reference.setEditors(editor);
         reference.setType(type);
         reference.setNote(note);
+        if(visibility == "Private") {
```
jdamerow: strings should not be hard-coded; at a minimum this needs to be a constant somewhere

vspace/src/main/java/edu/asu/diging/vspace/core/model/impl/Reference.java
```diff
@@ -53,6 +53,8 @@ public class Reference extends VSpaceElement implements IReference {
     private String type;
 
     private String note;
+    
+    private boolean visibility;
```
jdamerow: boolean variables should be adjectives so that the methods make sense when you call them `is...` (i.e. `isVisibility` does not make much sense...)

jdamerow (changes requested):
Also, there are codefactor issues and what happened to all the classes in the "references" package (e.g. https://github.com/diging/virtual-spaces-2.0/pull/299/files#diff-8e2f9f515ef07dfc360ae67df0401001f8cd6b89b960b99068f972b3fd962eec)

vspace/src/main/java/edu/asu/diging/vspace/web/staff/AddReferenceController.java
```diff
@@ -23,7 +23,8 @@ public class AddReferenceController {
 
     @RequestMapping(value = "/staff/module/{id}/slide/{slideId}/bibliography/{biblioId}/reference/add", 
                     method = RequestMethod.POST)
-    public ResponseEntity<IReference> addReference(@PathVariable("id") String moduleId, 
+    public ResponseEntity<Reference> addReference(@PathVariable("id") String moduleId,
```
jdamerow: is this needed? why the change?

jdamerow:
Make it so, Jenkins.

jdamerow (changes requested):

jdamerow:
- when sorting the table, the column widths change, which they shouldn't
- there should be a column that has the module a reference belongs to and a link that gets you to the slide that has the reference

